PR #6, opened 2015-11-26
Adding ZooKeeper and Kafka init actions

evilsoapbox:


evilsoapbox:
Removed tags added by accident.

lesv:
We should probably update the main README as well to describe what each script does. LGTM from my point of view, but since Dennis is assigned, I'll refrain from merging.

dennishuo:
Merged with https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/commit/87b54eb1fab2505a53d83d8864aacb90c7b50b24 and https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/commit/1bacd7e3700b9733b53b83a929c15c360b742af2

Issue #30, opened 2016-02-22
Init action for Jupyter and JupyterHub


nehalecky:
[Jupyter is a widely adopted web-based user interface](http://jupyter.org/) to interactive computing sessions with a variety of kernels. Currently, dataproc init actions contain support for [~~IPython~~ Jupyter Notebook](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/tree/master/ipython-notebook), however, this does not integrate with recent features [supporting package and environment management in conda](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/tree/master/conda). 

[JupterHub supports multi-user sessions](https://jupyterhub.readthedocs.org/en/latest/) to Jupyter in a secure and efficient manner.  We should expand on existing init action functionality to also:

- [x] install Jupyter ~~and JupyterHub~~
- [ ] support install via pip or conda
- [x] launch Jupyter~~(Hub)~~ on cluster create
- [x] properly log sessions
- [ ] script / command to remotely start / stop service
- [ ] support use of multiple notebooks per user

pquentin:
@nehalecky This sounds great! The current Jupyter init script leaves a lot to be desired, and needs to be fixed for each Spark release because it specifies the py4j path.

Thanks for working on this.

For a single jupyter instance, I think it's better to launch jupyter from pyspark (using PYSPARK_PYTHON=python and PYSPARK_DRIVER_PYTHON=jupyter) rather than launching spark from jupyter. It makes things much simpler. Not sure if this still applies with JupyterHub.

nehalecky:
@pquentin thanks for the note!

> For a single jupyter instance, I think it's better to launch jupyter from pyspark (using PYSPARK_PYTHON=python and PYSPARK_DRIVER_PYTHON=jupyter) rather than launching spark from jupyter. It makes things much simpler.

This sounds like an interesting approach, and I'd be willing to implement, but I am unfamiliar the param. Could you explain more how it simplifies things (I'd like to understand better)?

> Not sure if this still applies with JupyterHub.

OK, good point. I think whatever mechanism that we associate Jupyter instance with PySpark should be consistent between Jupyter and JupyterHub, do you agree? If so, how could we find out more about this?

Thanks!

pquentin:
This is how we do it:

```
export PYSPARK_PYTHON=python
export PYSPARK_DRIVER_PYTHON=jupyter
[...]

Issue #15, opened 2015-12-06
Expose init action API to a higher level provisioner / orchestration library?


nehalecky:
I've been testing and using Dataproc for less than a week, so far, just awesome. Kudos to all the Google developers who made it a reality. I've recently been diving into custom provisioning of the cluster, and was able to quickly setup and connect to Jupyter using the IPython init action provided in this repo. Thanks, and it works great.

As the need for more refined control is always pressing (both myself and others #9, #8), it would be interesting to understand the decision behind the current architecture for Dataproc initialization actions, and how it compares to the roadmap for future functionality? 

In getting something off the ground, it's simple and effective to point to a bunch of bash shell scripts to perform provisioning. Admittedly this is what I do a majority of the time when setting up a local dev environment via provisioning a VM. In my case, however, one begins to realize that managing and passing state through a bunch of custom environment variables loses it's charm quickly, along with the need to parameterize through interactions with higher level objects become painfully apparent. :/

With this idea, was there any plan to expose Dataproc initialization actions to the provisioning capabilities of platforms like [Ansible](http://www.ansible.com/) or [Salt](http://saltstack.com/)? This would allow the *master* and *worker* nodes to simply be targeted with a playbook / formula, and allow the Dataproc end user the ability to tap into a large public resource of existing mechanisms for provisioning nodes with particular libraries and / or configurations.

Just a thought, and hope that it make sense. Thanks again for such a great resource. :)

evilsoapbox:
Hello!

Apologies for the delayed response! We've been hard at work on some new features and our 12/16 release. Thanks for positive feedback - it's always **awesome** to hear things are working well.

In the long-term, we have a few things in mind. Unfortunately, I cannot go too much into our long-term roadmap (timing or commitment) but we've definitely discussed a few things:

* When initialization actions should run (before daemons, after, both?)
* Parameters to init actions
* Alternatives to initialization actions
* Templatizing clusters

I think we'll likely expand the richness of init actions (parameters, etc.) and they will be a core part of Dataproc in the long term. We are also thinking about ways we can expand deployment and provisioning in the future, especially if other frameworks (like Flink) become popular to use alongside Spark/Hadoop (Yarn + HDFS.) 

[...]

PR #31, opened 2016-02-22
WIP: Feature/jupyter notebooks/develop 

nehalecky:
A work in progress for #30, needs code review and resolution to a few issues, please don't merge yet.


nehalecky:
@jeffkpayne, could I get your review of this init action? I was hoping to understand better an optimal pattern for passing env vars (e.g., [`DATAPROC_CLUSTER_NAME`](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/31/files#diff-c074072b66ed05f1b43eb2b0eeb04565R6)) in call to script instead of statically defining.

jeffkpayne:
@nehalecky, do you mean for the scripts that run local to the user or for those that run as init-actions on the cluster?  For the locally run scripts, simply exporting the environment variable in bash is enough for the script to use it.  If you want to "pass in" values to the init-action scripts, there is the `--metadata` argument to the `gcloud beta dataproc clusters create...` command that takes key/value pairs that are accessible from within init-action scripts using the `get_metadata_value` command.  For instance, the call to `create...` in the bash script run locally would be something like:

```
gcloud beta dataproc clusters create $DATAPROC_CLUSTER_NAME \
    --metadata DATAPROC_BUCKET=$DATAPROC_BUCKET \
    --initialization-actions \
        gs://$DATAPROC_BUCKET/dataproc-init-actions/bootstrap-dataproc-cluster.sh \
    --bucket $DATAPROC_BUCKET \
    --num-workers 2
```

and the init-action script would read that value with something like:

`DATAPROC_BUCKET=$(/usr/share/google/get_metadata_value attributes/DATAPROC_BUCKET)`

I'm not sure if there's a better way to do that sort of thing...

nehalecky:
Thanks, I'm working on cleaning up static definitions, `--metadata` seems like the correct place to pass parameters like this. :)

Another issue I'm having is launching Jupyter notebook on cluster launch via call to [`launch-jupyter.sh`](https://github.com/nehalecky/dataproc-initialization-actions/blob/feature/jupyter_notebooks/develop/jupyter/launch-jupyter.sh) in this PR, which I call in a general `bootstrap-dataproc-cluster.sh` script like:

```bash
echo "Installing and configuring Jupyter on cluster master node..."
./dataproc-initialization-actions/jupyter/setup-jupyter.sh
echo "Launching Jupyter on master node..."
./dataproc-initialization-actions/jupyter/launch-jupyter.sh &
```

Any thoughts?

jeffkpayne:
@nehalecky, you still having issues with `launch-jupyter.sh`?  I didn't see any errors when I modified my local bootstrap script to call it, but then I noticed that you pushed some changes to it.  I *did* notice that it looks like lines 12 and 14 are missing the `$` when referencing the `JUPYTER_IP` var.  See my comment on that commit...

nehalecky:
> I didn't see any errors when I modified my local bootstrap script to call it, but then I noticed that you pushed some changes to it.

@jeffkpayne correct, no errors, however, if you check to see if jupyter application launch was successful with a `ps -ef | grep jupyter`, I get no love. :disappointed:  

If I ssh into master and execute same command (i.e., `./dataproc-initialization-actions/jupyter/launch-jupyter.sh`) it launches successfully. :confused: 

:sos:

jlowin:
@nehalecky for Jupyter on Dataproc, we switched to using kernels instead of startup scripts for two reasons: first, it lets users easily "drop down" into Python instead of PySpark by simply selecting the Python kernel, which is sometimes helpful (especially if the executors are monopolized); and second, there were some conflicts running `pyspark` with `PYSPARK_PYTHON=ipython` since it essentially tried to start twice (though in fairness that's only an issue for debugging). With a kernel, the startup script only runs when explicitly instructed by Jupyter (similar to an Ipython profile)

Our setup looks like this, in case it's interesting:

#### Inside `jupiter-run.sh`
```bash
	mkdir -p $JUPYTER_NOTEBOOK_DIR

	echo "Creating Jupyter config..."
	jupyter notebook --generate-config -y --ip=127.0.0.1
	echo "c.NotebookApp.ip = '*'" >> ~/.jupyter/jupyter_notebook_config.py
	echo "c.NotebookApp.open_browser = False" >> ~/.jupyter/jupyter_notebook_config.py
	echo "c.NotebookApp.port = $JUPYTER_NOTEBOOK_PORT" >> ~/.jupyter/jupyter_notebook_config.py
	echo "c.NotebookApp.notebook_dir = '$JUPYTER_NOTEBOOK_DIR'" >> ~/.jupyter/jupyter_notebook_config.py

	# install pyspark kernel
	echo "Installing pyspark Kernel..."
	jupyter kernelspec install /path/to/pyspark/kernel/folder/
	echo "c.MappingKernelManager.default_kernel_name = 'pyspark'" >> ~/.jupyter/jupyter_notebook_config.py

	# install findspark
	pip install findspark

    # Start Jupyter
	echo "Starting Jupyter Notebook on port $JUPYTER_NOTEBOOK_PORT..."
	nohup jupyter notebook --no-browser > /var/log/python_notebook.log &
```

#### `kernel.json` (in a folder called `pyspark`)
```json
{
 "argv": [
    "python", "-m", "ipykernel", "-f", "{connection_file}"],
 "display_name": "PySpark",
 "language": "python",
 "env": {
    "PYTHONSTARTUP": "/path/to/start_pyspark.py"
    }
}
```
#### `start_pyspark.py`
```python
import findspark

findspark.init()

import pyspark
from pyspark.sql import HiveContext

sc = pyspark.SparkContext()
sqlContext = HiveContext(sc)
```

nehalecky:
@jlowin, hey! Thanks for the tip on use of kernels. I've only casually browsed documentation, however, now seeing your implementation it does seems like a nice way to control use of PySpark with Jupyter on Dataproc. A few questions for you, if you don't mind:

- how do you support multiple users on the same cluster (jupyterhub)?
- what is your workflow for collaboration of Jupyter notebooks?
- are you familiar with or have thoughts on [launching Jupyter via PYSPARK_DRIVER_PYTHON](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/30#issuecomment-187680844)?
- do you have any perform any dynamic scaling the cluster up and down to support analysis needs?

Thanks agains, look forward to your reply.

jlowin:
@nehalecky great questions, many of the answers are things we're currently exploring (and not set in stone)! And I'm not trying to argue that the kernel approach is "best" by any means; it has however served our purposes nicely in terms of flexibility. But for now:

1. Don't have a solution on hand; at the moment most interactive/development work is done in ad-hoc clusters so user management isn't an issue
2. We have an initialization action that mounts a GCS bucket using [GCSFuse](https://github.com/GoogleCloudPlatform/gcsfuse) when the Dataproc cluster starts. The Jupyter notebook directory points into that bucket, so notebooks are effectively shared across clusters and persistent. This isn't ideal (partially because GCSFuse was just discontinued!) but works for now. I have looked at harnessing a git-based workflow similar to Google DataLab and would love to hear any other thoughts.
3. Actually have never considered that pattern before! @pquentin what happens when you submit a remote job to the cluster -- does it try to run it in the notebook? Does it even matter? If I understand it correctly, anytime you run pyspark you are also running Jupyter; does that allow alternative kernels? I'm interested in kernels because of projects like [Apache Toree](https://github.com/apache/incubator-toree) (formerly spark-kernel).
4. Haven't crossed that bridge yet due to ad hoc cluster usage as mentioned in 1).

HTH!

dennishuo:
Indeed, `/usr/share/google/get_metadata_value` is the preferred way to fetch parameters/env vars which are used in various places inside init scripts. Currently, there are already several such attributes where you don't have to manually specify `--metadata` and you'll get automatically, all which start with the `dataproc-` prefix, such as:

    dataproc-bucket
    dataproc-cluster-name
    dataproc-cluster-uuid
    dataproc-datanode-enabled  # Whether this machine runs datanotes; preemptible VMs do not
    dataproc-master # Hostname of the master of the cluster
    dataproc-role # Master or Worker
    dataproc-worker-count # Initial number of workers

The `dataproc-worker-count` might not reflect an accurate count after updates, and with actual preemptible VM counts, so it's generally better if setup scripts don't have to worry about worker counts.

When you say you had an issue with calling `launch-jupyter.sh`, what kinds of errors were you running into? If starting things async, you probably need to preface your async command with "nohup", and better yet, integrate with `systemd` so that it's auto-started on reboot.

jeffkpayne:
@dennishuo, Thanks!  Couldn't find the default metadata attributes in the docs and I must have missed them when I checked the cluster UI for my custom attribute being set.  The less metadata managed by the custom scripts, the better.  Thanks again.

nehalecky:
Hey all, been quite busy but will pick this up again tomorrow, and hoping to finish by the end of this week. For us, this is an important milestone for our collaborative analytics needs. On the topic of supporting multiple users via JupyterHub (concurrently running jobs on a single spark dataproc cluster), wanted to point out this writeup: [Spark cluster on OpenStack with multi-user Jupyter Notebook](http://arnesund.com/2015/09/21/spark-cluster-on-openstack-with-multi-user-jupyter-notebook/). 

- @jlowin, they also follow your pattern in use of a custom pyspark kernel, kudos. :)
- @dennishuo, with the `launch-jupyter-interface.sh` of this PR, we're trying to abstract away a lot of need for explicitly having to set bash params and instead [allow a more concise way to pass needed variables](https://github.com/nehalecky/dataproc-initialization-actions/commit/49acdbcdca8bcc7f8cfcfa81163dc8f979d823ac#diff-c074072b66ed05f1b43eb2b0eeb04565R4). Does dataproc have any of these env vars exposed through use of the gcloud sdk? If so, we could default to them in use of these local scripts, which would be very convenient.
- @jeffkpayne, do these recent commits fix the issue of launching Jupyter on cluster creation?

jeffkpayne:
@nehalecky, yes, the recent commits fix the issue of jupyter not being started on cluster launch.  Pushing one more change to the `launch-dataproc-cluster.sh` script itself to use the same branch name logic used by the `build-init-action-branch.sh` in 5 minutes.

dennishuo:
@nehalecky I think cluster name should still be an argument to the `launch-jupyter-interface.sh` script, but beyond that all other info should be made readily available via gcloud commands, by specifying `--metadata` key/value pairs at cluster-creation time which the init actions will read using the `get_metadata_value` helper, and then on the client side any scripts or wrapper programs can easily parse those shared metadata values from the API cluster resource itself.

It'd be hard to guess which cluster is intended without specifying a cluster-name, and making it an argument makes it easy to manage multiple different clusters at once.

Things like `zone` can be fetched from the cluster's `zoneUri`:

    ZONE=$(gcloud dataproc clusters describe ${DATAPROC_CLUSTER_NAME} | grep zoneUri | cut -d '/' -f 9)

And if I created my cluster with `--metadata jupyterport=10000,foo=bar` then my cluster has:

    gceClusterConfig:
      metadata:
        foo: bar
        jupyterport: '10000'

So that can likewise be parsed out into the environment variable. I suppose it might be convenient to have a small helper tool that properly parses YAML or JSON representations of resources and allows retrieving fields by their resource hierarchical path.

jeffkpayne:
@nehalecky I'll work on the changes to `launch-jupyter-interface.sh` recommended directly above, FYI.

jeffkpayne:
@nehalecky I made the suggested changes in my most recent commits.  I also made the corresponding changes to the launch scripts, etc in https://github.com/nehalecky/BomboraAnalytics/tree/environment/update-provisioning. See issue https://github.com/nehalecky/BomboraAnalytics/issues/12.

@dennishuo I included a couple of simple utility functions to make it a little easier to extract values from the `gcloud... describe...` output, but it's far from ideal.  There's a python package (https://github.com/0k/shyaml) that works great from the shell and provides more than enough functionality, but I wasn't sure about adding a dependency to a python package to the client-side `launch-jupyter-interface.sh`.  I imagine the target users are going to be pretty python package savvy, though, so maybe adding that dependency isn't too bad?

dennishuo:
@jeffkpayne we could certainly consider adding dependencies on some lightweight python packages, though if it's possible to use Python built-ins instead it'd be nicer to avoid the dependency.

Would Python's `yaml` package do the trick potentially in this case? For example, the following seemed to work well for me:

    $ gcloud dataproc clusters describe dhuo-meta > cluster.yaml
    $ cat cluster.yaml | python -c "import sys,yaml; cluster = yaml.load(sys.stdin); print cluster['config']['gceClusterConfig']['metadata']"
    {'foo': 'bar', 'jupyterport': '10000'}
    $ cat cluster.yaml | python -c "import sys,yaml; cluster = yaml.load(sys.stdin); print cluster['config']['gceClusterConfig']['metadata']['jupyterport']"
    10000

You could also split out the 3 lines of Python into a separate file or helper bash function so that the call site doesn't have to have the long cat command, something like:

    function get_metadata_property() {
      # $1 is clusterName, $2 is the metadata key you're looking for
      gcloud dataproc clusters describe $1 | python -c "import sys,yaml; cluster = yaml.load(sys.stdin); print cluster['config']['gceClusterConfig']['metadata']['$2']"
    }

    PORT=$(get_metadata_property dhuo-meta jupyterport)

jeffkpayne:
@dennishuo Yep, using the builtin `yaml` package is obviously the better option!  Thanks.  I was also concerned about introducing a dependency on python into the client-side bash scripts at all, but that concern is probably undue since this is going to be used by python users.  I'll integrate your snippet into the `util/utils.sh` bash helper file I already added.

jeffkpayne:
@nehalecky Do we want to move to the kernel based approach for jupyter, using http://arnesund.com/2015/09/21/spark-cluster-on-openstack-with-multi-user-jupyter-notebook/ as an example?

nehalecky:
@jeffkpayne, I think the kernel based approach is a good idea, and like the design proposed by @jlowin above (seems to fit nicely with dataproc), but whatever works. :) Ultimately, the goal is to allow multiple users use of multiple Jupyter notebooks, all tied to a single Spark cluster as the computational resource. Perhaps we can finish this kernel effort, and then break off the JupyterHub functionality efforts to another PR so we can ship the working Jupyter init actions? Thoughts?

Also, FWIW, I believe the scope of this PR is expanding a little with inclusion of the `monitoring` and `utils` modules, which helps abstract common functionality. I totally support this and in general am a huge fan of such, however, in this case I can't help but feel we're reinventing the wheel (vs. adoption of a provisioning framework like mentioned in #15). 

Imagine, instead of trying to mesh together a collection of bash shell scripts for our provisioning efforts, we could simply target something like an [Ansible playbook](http://docs.ansible.com/ansible/playbooks.html) (a yaml-formatted file containing instructions for setting state across different VMs). This would prevent efforts here from being spent on creating and capturing workflows in low-level shell-scripts, and instead leverage a high-level API with much more expressive and targeted provisioning functionality. This could be implemented as another provisioning path here, in the `dataproc-initialization-actions` code base, but long term would be support on dataproc API, perhaps through use of a `--init-action-playbook`, that could be leveraged on a `glcoud dataproc create` and `gcloud dataproc update`. Food for thought.

jeffkpayne:
@nehalecky FYI, kernel-based Jupyter config is working as of my last commit.

nehalecky:
@jeffkpayne, FWIW, see [Spark Docs: Scheduling Across Applications](https://spark.apache.org/docs/latest/job-scheduling.html#scheduling-across-applications). Seems like there is some _possibly_ helpful information there for configuring spark cluster for multi-job / multi-user workloads, specifically under the [Dynamic Resource Allocation](https://spark.apache.org/docs/latest/job-scheduling.html#dynamic-resource-allocation) section.

nehalecky:
@jeffkpayne, Also, I just tested launching multiple notebooks using [the PYSPARK_DRIVER_PYTHON method proposed by](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/30#issuecomment-187680844) @pquentin, and unfortunately, it doesn't resolve the issue with multiple `SparkContext`s.

I guess by method of elimination, we're getting closer? :confused:

jeffkpayne:
Lol. Check your Google chat. I tried that and starting two shells. No dice.

On Sat, Mar 12, 2016, 1:20 PM Nicholaus Halecky <notifications@github.com>
wrote:

> @jeffkpayne <https://github.com/jeffkpayne>, Also, I just tested
> launching multiple notebooks using the PYSPARK_DRIVER_PYTHON method
> proposed by
> <https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/30#issuecomment-187680844>
> @pquentin <https://github.com/pquentin>, and unfortunately, it doesn't
> resolve the issue with multiple SparkContexts.
>
> I guess by method of elimination, we're getting closer? [image:
> :confused:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/31#issuecomment-195810549>
> .
>
-- 
Jeff Payne
jeffkpayne@gmail.com
916-273-0938

nehalecky:
Hey all, just rebased and cleaned up the commit history. We have few more commits with documentation on multiple users and notebooks coming, and after, this PR will be ready! 

@jeffkpayne and I are the authors @googlebot, no worries! :ok_hand:

jlowin:
@nehalecky looking forward to seeing what you have here! 

In the meantime, I've been experimenting in a different direction and wanted to share some progress in case anyone is interested (since I think everyone in this thread is trying to achieve the same thing!). Using `remote_ikernel` (https://bitbucket.org/tdaff/remote_ikernel), I'm testing a *local* Jupyter notebook that hooks in to my DataProc pyspark kernel. This has the big advantage of not requiring any notebook storage management on the cluster.

I run it like this:

```bash
# set up passwordless ssh to DataProc cluster (this might be overkill but it works)
gcloud compute config-ssh

# install remote_ikernel
pip install remote_ikernel

# add remote kernel for DataProc
# start_pyspark.py is the same as described earlier in the thread
# $DATAPROC_CLUSTER is the address of the cluster's master node
remote_ikernel manage --add \
	--kernel_cmd="PYTHONSTARTUP=\"/remote/path/to/start_pyspark.py\" python -m ipykernel -f {connection_file}" \
	--name="DataProc PySpark ($DATAPROC_CLUSTER)" \
	--interface=ssh \
	--host=$DATAPROC_CLUSTER \
	--language="python"

# start local notebook, then select the remote kernel
jupyter notebook
```

Open items:

1. Having to set `PYTHONSTARTUP` in kernel_cmd is ugly but works (along with the kernel approach I outlined higher in this thread). Maybe there's a better way, but `remote_ikernel` doesn't seem to allow setting kernel parameters other than `argv` (via `kernel_cmd`)
1. making sure to release the kernel when done so that pyspark becomes available for submitted jobs

This works locally in limited testing (VERY limited). In the future, I want to explore running JupyterHub on GKE along with a job that periodically scans for DataProc clusters and adds/removes remote_ikernels as needed. Notebooks could live in a persistent volume. Then all users could log into the JH app and not have to worry about accessing the cluster itself (other than making sure its running).

nehalecky:
> Notebooks could live in a persistent volume. Then all users could log into the JH app and not have to worry about accessing the cluster itself (other than making sure its running)

@jlowin, seems like an awesome setup, and I perhaps eventually be a part of this `jupyter` init action after the remaining issues you mentioned are ironed out. Perhaps it could be housed under a `setup_jupyter_remote_ikernel.sh`, or something similar? I think we should open 

A few questions: 
- how does one control dependencies on the local machine running jupyter
- can it link into environment management tools (e.g., conda, virtualenv)?

> I want to explore running JupyterHub on GKE along with a job that periodically scans for DataProc clusters and adds/removes remote_ikernels as needed. 
I'm not familiar, GKE is [Google Container Engine](https://cloud.google.com/container-engine/)?

As @jeffkpayne and I have pretty much wrapped up this PR, I think we should open up another issue to begin development on what you're proposing—seems awesome. :)

jlowin:
@nehalecky I've released a project that I think will be interesting to everyone on this list: https://github.com/jlowin/jupyterhub-dataproc

Appreciate any thoughts!

nehalecky:
@jlowin, that looks like an awesome project. I will hop over there to comment when I can. 

@jeffkpayne, this PR is ready to merge, correct? 

@dennishuo, let us know if we have any outstanding issues with @googlebot? Seems it wasn't too happy with two authors on this PR!

Thanks all!

dennishuo:
@nehalecky Thanks for the ping, and sorry for the delays! Between GCP NEXT and an office shuffle last week I never got around to checking if this had been wrapped up and ready for a final once-over before merging, I'll take a look and poke at it now.

No worries about @googlebot, it's still learning :)  there's precedent for this kind of thing where the rule is that as long as all authors have CLAs on file (and I've verified for you and @jeffkpayne successfully), then the other author's explicit participation in the PR thread is sufficient (but perhaps not strictly *necessary* in all cases) to override the CLA status.

jeffkpayne:
Yep, everything should be good to go IMO.  My last commit was just removing a debug `echo`.

dennishuo:
@nehalecky It appears there are some quirks arising from the way this has been developed (perhaps by `git clone` onto a Dataproc master node and then running the scripts explicitly in an SSH session) and as a result there doesn't appear to be an easy way to use these as an actual initialization action without writing an extra wrapper script and/or editing the scripts somewhat. In particular, the following tend to make some assumptions about being run manually with certain files present on fixed relative paths to the running script:

- File dependencies like `kernel.json`
- Sourcing ../util/utils.sh
- Erroring out instead of exit-early if the script is run on a worker

Additionally, the conda dependency isn't obvious at first so we'll want to at least document it but it might be better if there's a nice way to bundle the dependencies all together somehow.

It should be hard to wrangle all this into a nice turnkey solution though, and I think would be worth it for ease of use if folks can just copy-paste a minimal set of commands to have it all up and working.

For what it's worth, I got it working by creating a wrapper init action like the following, which just uses github itself as the hermetic dependency-distributor:

    #!/bin/bash

    ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
    if [[ "${ROLE}" != 'Master' ]]; then
      exit 0
    fi
    git clone https://github.com/bomboradata/dataproc-initialization-actions.git
    cd dataproc-initialization-actions
    git checkout feature/jupyter_notebooks/develop
    cd jupyter
    chmod 750 *.sh
    ./setup-jupyter-kernel.sh
    ./bootstrap-jupyter-ext.sh
    ./launch-jupyter-kernel.sh


And then my cluster-creation command looked something like this: 

    gcloud dataproc clusters create dhuo-jupyter --metadata JUPYTER_PORT=8124 --initialization-actions gs://hadoop-cloud-dev-dhuo/conda/bootstrap-conda.sh,gs://hadoop-cloud-dev-dhuo/conda/install-conda-env.sh,gs://hadoop-cloud-dev-dhuo/jupyter/init-action-harness.sh

This is definitely an interesting case because as we expand the usability of init actions we'll likely want to provide the right mechanisms to have things like file dependencies and other-init-action dependencies supported in a first-class way. In the meantime, though, there are a few options:

1. Add a script for running on local workstation which wraps `gcloud dataproc` itself so that it can perform setup, staging files, etc., and potentially pass dynamic arguments to the cluster-creation request itself, such as [this cloud-bigtable example](https://github.com/GoogleCloudPlatform/cloud-bigtable-examples/blob/1b7e1ad3373a49b6eff83f3f705452e32dc10d61/java/wordcount-mapreduce/cluster.sh)
2. Modify the scripts that are intended to be run as init actions to no longer assume being run manually in a customized environment; remove `source "$DIR/../util/utils.sh"` calls for scripts intended to be run on init actions, and if only intended to be run on the master, add something that early-returns with exit code 0 when not on a master. Might have to inline `kernel.json` as a here-document in this case and generate it dynamically from one of the init scripts.
3. Go ahead with the idea of github being the distribution mechanism and add a wrapper script like what I have above, and document the way this is intended to be used. This approach would also allow the wrapper script to automatically also run any dependencies like the Conda init actions, rather than requiring the user to stage those as separate init action scripts.

Any thoughts on these or preferences? I have no strong preference at this point for any particular one.

jeffkpayne:
@dennishuo Thanks for the in-depth analysis and response.  We actually do use a set of scripts (in a different repo) that fall into the category of the first option.  Our `launch` script copies another "bootstrap" init script up to GCS and specifies it as the init action script in the `gcloud dataproc clusters create...` command.  That bootstrap init action script calls the scripts under the `jupyter` directory, in much the same way that the script you wrote does.  We are happy to add these "driver" scripts we use to this PR and we apologize for not just including said scripts in the first place.

We are also interested in having things like interdependencies handled in a better way.  I know @nehalecky has brought this issue up before, with regard to using something beyond `bash` scripts as the framework for specifying init actions (https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/15), and the response was, essentially, "we're working on it".

Perhaps one temporary solution which wouldn't require a change to the gcloud sdk itself would be a generic [Ansible](http://www.ansible.com) bash init action script?  Said script would handle installing any system level packages, etc, and then reading in the playbook, etc (I'm not really familiar with Ansible).

At any rate, we're happy to provide the driver scripts we use, which are already fairly user friendly with many parameters around cluster creation specs, etc.

jeffkpayne:
@dennishuo I went ahead and made the changes outlined in the third option.  There is now a single init action `jupyter.sh` which essentially does the same thing as the one you were forced to write (sorry about that), only with a couple of niceties that I found helpful during development and use.  The first is that the git repo/branch that gets cloned is optionally configurable via metadata, defaulting to `https://github.com/GoogleCloudPlatform/dataproc-initialization-actions` and `master`.  The second is that the `notebooks` directory in the `dataproc-bucket` on GCS is copied to the master node and used as jupyter's default notebooks directory to make things a little easier.

I just realized that I forgot to add a real-world example of creating a cluster using this init action with the available options.  I'll do that now.  Thanks again for your time and help on this!

jeffkpayne:
Updated the README with a real world usage example.  Let me know if there are any other issues.  Thanks!

nehalecky:
@jeffkpayne, nice work. 

Quick feedback: now that we have the dataproc init actions repo available at the public bucket `gs://dataproc-initialization-actions/` (see #40), perhaps we can simplify the real world example by modifying the `--initialization-actions` to simply point to  `gs://dataproc-initialization-actions/jupyter/jupyter.sh`, instead? Also, if we do this we can remove the `git clone` and `gsutil cp` before it.

Let me know if that makes sense.

jeffkpayne:
Ah, yes. I was unclear on the availability of that bucket. Of course, that
would make it much simpler. I'll update the example. Thanks.

On Wed, Mar 30, 2016, 7:54 AM Nicholaus Halecky <notifications@github.com>
wrote:

> @jeffkpayne <https://github.com/jeffkpayne>, nice work.
>
> Quick feedback: now that we have the dataproc init actions repo available
> at the public bucket gs://dataproc-initialization-actions/ (see #40
> <https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/40>),
> perhaps we can simplify the real world example by modifying the
> --initialization-actions to simply point to
> gs://dataproc-initialization-actions/jupyter/jupyter.sh, instead? Also,
> if we do this we can remove the git clone and gsutil cp before it.
>
> Let me know if that makes sense.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/31#issuecomment-203472480>
>
-- 
Jeff Payne
jeffkpayne@gmail.com
916-273-0938

dennishuo:
Awesome, thanks for the quick updates! I'll take another look over everything.

jeffkpayne:
Thanks again for your insight and suggestions.  First contribution to any open source project for me :)

dennishuo:
@jeffkpayne Always happy to help! Kudos on open source, hopefully the first of many :+1:

PR #32, opened 2016-02-23
fix attributes/role for dataproc

rangadi:
minor fix to key for dataproc role.


dennishuo:
Merged https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/commit/241ad75a6edd4b8ff4f5308470df957e6c1d1b9f Thanks for catching and fixing this!

PR #33, opened 2016-02-24
zeppelin: Update Hadoop and Spark versions

pquentin:
Don't you think it would be better to automatically retrieve those values from the environment? Calling pyspark --version and hadoop version.

Also, I know Zeppelin does not release that often, but always checking out master sounds quite dangerous.


dennishuo:
Thanks, good catch! I'll go ahead and merge this for now to match the current Dataproc versions. As for having a more stable installation vs checking out from master, @pmkc has actually been hashing out the distro bits for this already, and should be able to strip out the explicit version strings, etc., in favor of `apt-get install` shortly: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/35

dennishuo:
Merged https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/commit/0a916cb5eea2c438cbae6b31c695425a99f812a2

PR #45, opened 2016-03-23
Fix master regexp for preemptible nodes

erikdubbelboer:
Preemptible dataproc nodes have a hostname that differs a bit from the
worker hostnames.


dennishuo:
Nice catch, thanks for submitting the fix!

PR #48, opened 2016-03-31
updating initialization action for apache hbase setup on dataproc clu…

rdoppalapudi:
Added support for hbase setup on the data proc cluster.

### DIFF
--- a/apache-hbase/hbase.sh
+++ b/apache-hbase/hbase.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+#    Copyright 2015 Google, Inc.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.set -x -e
+set -x -e
+
+# Variables for this script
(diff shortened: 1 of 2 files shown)


rdoppalapudi:
I signed it!

apache-hbase/hbase.sh
```diff
@@ -0,0 +1,48 @@
...
+set -x -e
+
+# Variables for this script
+HBASE_VERSION="0.98.18"
+ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+CLUSTER_NAME=$(hostname | sed -r 's/(.*)-[w|m](-[0-9]+)?$/\1/')
+
+# Download and extract ZooKeeper
+cd ~
+wget http://mirrors.advancedhosters.com/apache/hbase/${HBASE_VERSION}/hbase-${HBASE_VERSION}-hadoop2-bin.tar.gz
```
dennishuo: Though it's not automatically started right now, HBase 1.1.2 is actually already pre-installed on Dataproc clusters (in part because we need the hbase-client libraries for Cloud Bigtable connectivity); you should be able to edit /etc/hbase/conf/hbase-site.xml and use /usr/lib/hbase/bin/hbase-daemon.sh and that'll help this work across different Dataproc versions when HBase versions change in the distro as well as cut down on the time it takes to download/install.

dennishuo:
Thanks for looking into this! Left a note inline in the file comments; nowadays we actually already preinstall HBase 1.1.2 but it just isn't started/configured; you should be able to use `/usr/lib/hbase/bin/...` and conf in `/etc/hbase/conf/hbase-site.xml` for version compatibility and faster install.

bischofs:
is anyone going to merge this?

dennishuo:
Our preference would be to use the hbase binaries provided as part of the
Dataproc distro, this probably won't be merged with the existing tarball
download from http://mirrors.advancedhosters.com

PR #49, opened 2016-04-11
Fix presto memory configs, update to latest release.

dennishuo:
Instead of using fixed defaults, configure query-per-node memory,
    distributed total memory, and system reserved memory with help
    from the cluster's spark executor configs.
    
    Update to version 0.144.1.

### DIFF
--- a/presto/presto.sh
+++ b/presto/presto.sh
@@ -20,7 +20,7 @@ HOSTNAME=$(hostname)
 DNSNAME=$(dnsdomainname)
 FQDN=${HOSTNAME}.$DNSNAME
 CONNECTOR_JAR=$(find /usr/lib/hadoop/lib -name 'gcs-connector-*.jar')
-PRESTO_VERSION="0.126"
+PRESTO_VERSION="0.144.1"
 HTTP_PORT="8080"
 
 # Download and unpack Presto server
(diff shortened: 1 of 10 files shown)


presto/presto.sh
```diff
@@ -48,9 +48,35 @@ connector.name=hive-hadoop2
...
+# system MB as a crude approximation of other unaccounted overhead that we need
+# to leave betweenused bytes and Xmx bytes. Rounding down by integer division
+# here also effectively places round-down bytes in the "general" pool.
+PRESTO_QUERY_NODE_MB=$(( ${PRESTO_JVM_MB} * 6 / 10 - ${SPARK_EXECUTOR_OVERHEAD_MB} ))
+PRESTO_RESERVED_SYSTEM_MB=$(( ${PRESTO_JVM_MB} * 4 / 10 - ${SPARK_EXECUTOR_OVERHEAD_MB} ))
+
+# Compute the distributed max-memory by multiplying worker count by the
+# per-worker memory allocations.
+WORKER_COUNT=$(/usr/share/google/get_metadata_value attributes/dataproc-worker-count)
+PRESTO_QUERY_DISTIBUTED_MB=$(( ${WORKER_COUNT} * ${PRESTO_QUERY_NODE_MB} ))
```
pmkc: Do you know if this is necessary? It breaks with scaling, which is why we set the YARN equivalent to 10 TB

pmkc:
Seems reasonable overall.

dennishuo:
Re-opened in https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/50, I can't seem to fix this branch's commit logs.

PR #62, opened 2016-07-12
Making it more clear that install-conda-env.sh does not work alone

feczo:
Making it more clear that install-conda-env.sh does not work on its own as an action


dennishuo:
Closed with https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/commit/62d4a1c7a3fc5fc60629ba5283698d1b94d8475f

PR #67, opened 2016-08-02
Fix apt source urls

mt:
Perform `apt-get update` before installing ganglia packages to make sure that the repository urls are correct.


mt:
`I signed it!`

Issue #70, opened 2016-08-26
Jupyter setup failure for Dataproc 1.1 (Spark 2.0)


gilmar:
Hi,

It seems that the Jupyter initialization scripts are failing to correctly setup Jupyter in a cluster using Dataproc image version 1.1 (Spark 2.0).

Here is how I have created my dataproc cluster:

```
gcloud dataproc clusters create $CLUSTER_NAME \
--image-version 1.1 \
--project $PROJECT_ID \
--bucket $BUCKET \
--zone us-central1-f \
--num-workers 2 \ 
--scopes cloud-platform \
--initialization-actions gs://dataproc-initialization-actions/jupyter/jupyter.sh \
--initialization-action-timeout 60m
```

When I open the Jupyter UI and run "sc.version" in a cell I get the following error:

```
NameErrorTraceback (most recent call last)
<ipython-input-1-fd0e3f469918> in <module>()
----> 1 sc.version

NameError: name 'sc' is not defined
```

I presume this is something you are already working on, but I didn't find any issues opened reporting this problem. Meanwhile I'm trying to solve it by myself.
Let me know if you need further details.
Thanks in advance.

Regards,
Gilmar

gilmar:
Ok, the problem seems to be that Spark 2.0 changed the version of Py4J that is referenced in jupyter/kernels/pyspark/kernel.json
I was able to make it work just by applying this change:

```
<     "PYTHONPATH": "/usr/lib/spark/python/:/usr/lib/spark/python/lib/py4j-0.9-src.zip",
---
>     "PYTHONPATH": "/usr/lib/spark/python/:/usr/lib/spark/python/lib/py4j-0.10.1-src.zip",
```

mobcdi:
Hi @gilmar  where did you make the change (master?,workers?)

gilmar:
Hi @mobcdi ,

I was doing a local test. But I have just created a [pull request](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/71).
Meanwhile, you can try to use the script below as your initialization script. The only difference from the [original](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/jupyter/jupyter.sh) is that this one is pointing to my fork instead of this repo.
Just move it to your GCP bucket and point to it when creating your cluster, like this:
`--initialization-actions gs://$YOUR_BUCKET/jupyter.sh`

```

#!/usr/bin/env bash
[...]

PR #71, opened 2016-08-29
#70 Fixed Py4J path for Dataproc 1.1 (Spark 2.0)

gilmar:


jupyter/internal/setup-jupyter-kernel.sh
```diff
@@ -5,13 +5,22 @@ DIR="${BASH_SOURCE%/*}"
...
+DATAPROC_VERSION=$(gcloud dataproc clusters describe $CLUSTER_NAME | grep imageVersion | tr -d "'" | awk '{print $2}')
 ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
-JUPYTER_KERNEL_DIR="/dataproc-initialization-actions/jupyter/kernels/pyspark"
 JUPYTER_NOTEBOOK_DIR="/root/notebooks"
 JUPYTER_PORT=$(/usr/share/google/get_metadata_value attributes/JUPYTER_PORT || true)
 [[ ! $JUPYTER_PORT =~ ^[0-9]+$ ]] && JUPYTER_PORT=8123
 JUPYTER_IP=*
 
+
+if [[ "${DATAPROC_VERSION}" = "1.1" ]]; then
```
evilsoapbox: Quick note. It's likely you will see Dataproc 1.1.1 or greater in the near future. I am guessing this fix will probably work for those as well. What about making this >= 1.1 to avoid breakage?

evilsoapbox:
Hello!

Thanks for the pull request (awesome, as always!) I left one comment. Let me know what you think and we can go from there.

Cheers!

- James (Dataproc Product Manager)

jupyter/internal/setup-jupyter-kernel.sh
gilmar: Thanks @evilsoapbox ! Really good point. I'm updating the pull request.

evilsoapbox:
Hey @gilmar 

Two points; both of these are based on the fact that I missed you had two kernel JSON files, sorry. I should have caught that and missed it. Ideally, we'd not have to duplicate those files; instead can just use one kernel JSON. To that end:

1. For the .zip file, what about using something like `ls /usr/lib/spark/python/lib/py4j*.zip | sed 's/py4j\(.*\)\.zip/\1/'`? This can set the file across many versions.
2. For the `PYSPARK_SUBMIT_ARGS` what about having the conditional logic to add the `--packages com.databricks:spark-csv_2.10:1.3.0`  near line 17 in ` jupyter/internal/setup-jupyter-kernel.sh`

This would mean we'd probably be able to avoid further breakage when Dataproc 1.2 comes out /if/  something like the `PYTHONPATH` changes again, because the .zip version changes.

dennishuo:
Thanks @gilmar for sending the pull request! We ended up refactoring out a script which dynamically generates the JSON file instead of maintaining separate static JSON files, maybe you can verify if it works next time you need to spin up Jupyter on Dataproc 1.1: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/75

PR #73, opened 2016-09-06
Add Google Cloud Datalab init action.

pmkc:
@ojarjur PTAL


ojarjur:
This generally looks good to me.

The only caveat is that we need to release a new Datalab image with [PR 957](https://github.com/googledatalab/datalab/pull/957) merged before the user will be able to connect their Datalab frontend to the kernel gateway running in the Dataproc cluster.

pmkc:
Confirmed it works as expected at HEAD. merging.

PR #75, opened 2016-09-22
Generate pyspark kernel instead of static json.

dennishuo:
Use a lightweight script to determine appropriate settings based on the
local libraries found, versioning, etc., and dynamically generate
the json file used to install a pyspark kernel.

This fixes support for Dataproc 1.1.*/Spark 2*


evilsoapbox:
LGTM - will merge this afternoon.

PR #78, opened 2016-10-29
Fix typo in zeppelin conf file name

ipostanogov:


pmkc:
Thanks for catching this!

Issue #85, opened 2016-12-28
oozie script is broken


moander:
Vanilla dataproc clusters fails to run this script. I have tried image version 1.0, 1.1, and preview.

Image version `1.0.24`
```
Unpacking oozie (4.2.0-1) ...
dpkg: error processing archive /var/cache/apt/archives/oozie_4.2.0-1_all.deb (--unpack):
 trying to overwrite '/usr/lib/oozie/lib/zookeeper-3.4.6.jar', which is also in package oozie-client 4.2.0-1
dpkg-deb: error: subprocess paste was killed by signal (Broken pipe)
Processing triggers for man-db (2.7.0.2-5) ...
Processing triggers for systemd (215-17+deb8u5) ...
Errors were encountered while processing:
 /var/cache/apt/archives/oozie_4.2.0-1_all.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

Image version `1.1` and `preview`
```
Reading package lists... Done
+ apt-get install oozie oozie-client -y
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package oozie
E: Unable to locate package oozie-client
```

PR #87, opened 2017-01-04
Oozie script repository and sharedlib 

trK54Ylmz:
Hi,

I've made few changes about bigtop repository and sharedlib also tested on `1.1.15`.

I believe restarting to necessary services would be enough because `reboot` does not working for me.

#85 directly relevant with this PR.

### DIFF
--- a/oozie/oozie.sh
+++ b/oozie/oozie.sh
@@ -77,7 +77,7 @@ EOF
   # Required for Oozie sharedlib setup
   # throws ClassNotFoundException about org.apache.commons.io.Charsets
   wget http://central.maven.org/maven2/commons-io/commons-io/2.3/commons-io-2.3.jar
-  mv commons-io-2.3 /usr/lib/oozie/lib/
+  mv commons-io-2.3.jar /usr/lib/oozie/lib/
 
   # HDFS and MapReduce must be cycled; restart to clean things up
   service hadoop-hdfs-namenode restart
--- a/oozie/oozie.sh
+++ b/oozie/oozie.sh
@@ -15,7 +15,7 @@
 # Initialization action for installing Apache Oozie on a Google Cloud
 # Dataproc cluster. This script will install and configure Oozie to run on the
 # master node of a Dataproc cluster. The version of Oozie which is installed
-# comes from the BigTop repository. 
+# comes from the BigTop repository.
 #
 # You can find more information about Oozie at http://oozie.apache.org/
 # For more information in init actions and Google Cloud Dataproc see the Cloud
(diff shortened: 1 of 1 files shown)


trK54Ylmz:
I signed it!

oozie/oozie.sh
```diff
@@ -56,15 +65,31 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+  service hadoop-mapreduce-historyserver restart
+  service hadoop-yarn-resourcemanager restart
+  service hive-server2 restart
+  service oozie start
+
+  # Create sharedlib
+  oozie-setup sharelib create -fs hdfs://${HOSTNAME}
+
+  # Update available sharedlist
+  oozie admin -sharelibupdate -oozie http://localhost:11000/oozie
```
trK54Ylmz: This could be `${HOSTNAME}` but works both ways.

```
oozie admin -sharelibupdate -oozie http://${HOSTNAME}:11000/oozie
```

nloadholtes:
I used this PR and it worked for me. Is it going to be merged into master?

dennishuo:
@trK54Ylmz Thanks for sending this pull request! And sorry it required such workarounds to get it working; we're still working on longer-term more official support in future versions of the bigtop-based distro so that ideally it'd work out of the box and wouldn't require pinning distro versions and having workaround for any significant period of time, but depending on timelines, I suppose it could still be worth having it in as a temporary working solution until we get it cleaned up.

I'll try to find some cycles from our team to assess a bit more and determine how best to proceed.

dossett:
@trK54Ylmz Nice script!  The command `oozie-setup sharelib create -fs hdfs:` fails with this error:

```
org.apache.oozie.service.HadoopAccessorException: E0902: Exception occured: [No FileSystem for scheme: hdfs]
	at org.apache.oozie.service.HadoopAccessorService.createFileSystem(HadoopAccessorService.java:496)
	at org.apache.oozie.tools.OozieSharelibCLI.run(OozieSharelibCLI.java:159)
	at org.apache.oozie.tools.OozieSharelibCLI.main(OozieSharelibCLI.java:57)
```

Have you run into that problem?

dossett:
The issue was interaction between oozie 4.2.0 and hadoop 2.7.3  I had to:

1) add this to core-site:

```xml
<property>
<name>fs.hdfs.impl</name>
<value>org.apache.hadoop.hdfs.DistributedFileSystem</value>
</property>
```

2) edit the oozie-setup script to include `/usr/lib/hadoop/client/*`in the classpath when it launched the oozie setup CLI tool

medb:
This workaround seems to be obsolete.

PR #90, opened 2017-02-06
fix flink version & typo for tarball link

randyh0329:


dennishuo:
Thanks for the pull request; we opted to fix Flink in the base distro instead of having to install from tarballs: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/commit/6ebf8b495c7a797f9317d2a29235b7014634b6cf

PR #91, opened 2017-02-10
Fix flink init action.

bsidhom:
The flink download links were broken and some configuration variables were set incorrectly. This action now installs flink via apt-get and only runs on the master.

### DIFF
--- a/flink/flink.sh
+++ b/flink/flink.sh
@@ -48,6 +48,10 @@ function configure_flink() {
   # Number of worker nodes in your cluster
   local num_workers=$(/usr/share/google/get_metadata_value attributes/dataproc-worker-count)
 
+  # Number of Flink TaskManagers to use. Reserve 1 node for the JobManager.
+  # NB: This assumes > 1 worker node.
+  local num_taskmanagers="$(($num_workers - 1))"
+
   # Determine the number of task slots per worker.
   # TODO: Dataproc does not currently set the number of worker cores on the
   # master node. However, the spark configuration sets the number of executors
@@ -63,7 +67,7 @@ function configure_flink() {
 
   # Determine the default parallelism
   local flink_parallelism=$(python -c \
-      "print ${num_workers} * ${flink_taskmanager_slots}")
+      "print ${num_taskmanagers} * ${flink_taskmanager_slots}")
 
   # Get worker memory from yarn config.
   local worker_total_mem="$(hdfs getconf \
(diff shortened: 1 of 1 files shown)


flink/flink.sh
```diff
@@ -22,80 +22,75 @@
...
 fs.hdfs.hadoopconf: ${HADOOP_CONF_DIR}
 EOF
 
-# Start a Flink YARN session
-HADOOP_CONF_DIR=/etc/hadoop/conf ./bin/yarn-session.sh -n \
-	${FLINK_TASKMANAGER_SLOTS} --detached
+if [ "$START_FLINK_YARN_SESSION" = "true" ] ; then
+  env HADOOP_CONF_DIR="$HADOOP_CONF_DIR" \
+    "$FLINK_INSTALL_DIR/bin/yarn-session.sh" \
+    -n "${num_workers}" \
```
bsidhom: The number of YARN containers requested was previously set to the number of CPU cores on the master (more precisely, on any node on which the init action was run). It seems like setting this to the number of worker nodes is more appropriate since we're consuming almost all of the machine's memory for each TaskManager.

flink/flink.sh
```diff
@@ -22,80 +22,77 @@
...
+
+  # Determine the number of task slots per worker.
+  local flink_taskmanager_slots="$(hdfs getconf \
+    -confKey yarn.nodemanager.resource.cpu-vcores)"
+
+  # Determine the default parallelism
+  local flink_parallelism=$(python -c \
+      "print ${num_workers} * ${flink_taskmanager_slots}")
+
+  # Calculate the memory allocations, MB, using 'free -m'. Floor to nearest MB.
```
pmkc: Update comment

flink/flink.sh
```diff
@@ -22,80 +22,77 @@
...
-cd ${FLINK_INSTALL_DIR}
-wget ${FLINK_TARBALL_URI}
-tar -zxvf ${FLINK_TARBALL_NAME}
-mv ${FLINK_EXTRACT_DIRECTORY}/* .
-rm ${FLINK_TARBALL_NAME}
-rmdir ${FLINK_EXTRACT_DIRECTORY}
-
-# Apply Flink settings by appending them to the default config
-cat << EOF >> ${FLINK_INSTALL_DIR}/conf/flink-conf.yaml
+readonly FLINK_TASKMANAGER_MEMORY_FRACTION='0.8'
```
pmkc: This might be a bit wasteful, but we can change it later.

flink/flink.sh
bsidhom: Done.

flink/flink.sh
bsidhom: Adding a TODO.

bsidhom:
I manually tested and looked through the code and it turns out that `taskmanager.numberOfTaskSlots` is ignored by `yarn-session.sh`. Instead, this value _must_ be passed via the `-s` option.

This doesn't appear to be documented, so users creating their own sessions by CLI will need to know this.

PR #93, opened 2017-03-02
Add a volume mount for notebooks

### LINKED ISSUE #92: The Datalab initialization action should include a volume mount for /content/datalab
The Datalab initialization action [script](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/datalab/datalab.sh) specifies volume mounts for the spark libraries, but none for mapping any part of the VM's persistent disk to the '/content/datalab/' directory in the container.

That means notebooks won't be saved to the VM's disk

ojarjur:
This change adds an additional volume mount to the Datalab initialization action that maps "/content/datalab" from the Datalab container into the VM's persistent disk.

This fixes #92

### DIFF
--- a/datalab/datalab.sh
+++ b/datalab/datalab.sh
@@ -38,6 +38,9 @@ if [[ "${ROLE}" == 'Master' ]]; then
     echo 'spark.sql.warehouse.dir=/root/spark-warehouse' >> "${SPARK_CONF}"
   fi
 
+  DATALAB_DIR="${HOME}/datalab"
+  mkdir -p "${DATALAB_DIR}"
+
   # Expose every possbile spark libary to the container.
   VOLUME_FLAGS=$(echo {/usr/bin,/usr/lib,/etc,/etc/alternatives,/var/log}/{hadoop*,hive*,*spark*} \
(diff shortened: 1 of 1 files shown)


datalab/datalab.sh
```diff
@@ -38,6 +38,9 @@ if [[ "${ROLE}" == 'Master' ]]; then
     echo 'spark.sql.warehouse.dir=/root/spark-warehouse' >> "${SPARK_CONF}"
   fi
 
+  DATALAB_DIR="${HOME}/datalab"
```
pmkc: HOME is /root for an initialization action.
Should there be a metadata key of a GCS directory to copy here?

ojarjur: > Should there be a metadata key of a GCS directory to copy here?

I don't know of any reason to add such a thing, but maybe I'm overlooking something.

What makes you ask?

pmkc: We don't preserve disks between clusters, so those notebooks would need to be manually up/downloaded.

I suppose this FR makes sense in any case, if anyone reloads the docker container, but the invocation to reload is arcane and undocumented.

I'll merge this as is and we can think about notebook staging later.

PR #96, opened 2017-03-21
Feature/sql proxy fix

### LINKED ISSUE #95: Google Cloud SQL Proxy setup failed for Dataproc 1.1
Hi,

It seems that the Google Cloud SQL Proxy initialization script (**cloud-sql-proxy.sh**) is not working in a cluster using Dataproc image version 1.1

Using this dataproc creation command:

gcloud dataproc clusters create [CLUSTER_NAME] \
        --zone [ZONE] \
        --master-machine-type [MASTER_MACHINE_TYPE] \
        --master-boot-disk-size [MASTER_BOOT_DISK_SIZE] \
        --num-workers [NUM_WORKERS] \
        --worker-machine-type [WORKER_MACHINE_TYPE] \
        --worker-boot-disk-size [WORKER_BOOT_DISK_SIZE] \
        --scopes cloud-platform \
        --project [PROJECT_ID] \
        --properties hive:hive.metastore.warehouse.dir=gs://[SEGMENT]/hive-warehouse \
        --metadata hive-metastore-instance=[PROJECT_ID]:[REGION]:[SQL_INSTANCE_NAME] \
        --initialization-actions 'gs://[INIT_SEGMENT]/initialization-scripts/cloud-sql-proxy.sh'

The proxy service seems to be up and running but it is not reachable.

Going into [Cloud SQL proxy parameters and flags](https://cloud.google.com/sql/docs/mysql/sql-proxy#flags) I realized that some parameters are wrong in the initialization script, all parameters has to be written with the format -[PARAM]=[VALUE] (i.e.: **-dir=/cloudsql**).

After fixing them everything runs as expected.

Please let me know if you need more details

Thanks

danivzq:
This resolves #95 and also fix some format details for README file

### DIFF
--- a/cloud-sql-proxy/README.MD
+++ b/cloud-sql-proxy/README.MD
@@ -12,8 +12,10 @@ You can use this initialization action to create a Dataproc cluster using a shar
         --tier db-n1-standard-1 \
         --activation-policy=ALWAYS
     ```
-  a. Optionally create (or already have) other 2nd generation instances, which you wish to be accessible.
+    a. Optionally create (or already have) other 2nd generation instances, which you wish to be accessible.
+
 2. Uploading a copy of this initialization action (`cloud-sql-proxy.sh`) to [Google Cloud Storage](https://cloud.google.com/storage).
+
 3. Using the `gcloud` command to create a new cluster with this initialization action.
 
     ```bash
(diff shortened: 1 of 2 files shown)


dennishuo:
Thanks! Looks like the flag parsing was permissive so the old way also worked, but this will now be more in-line with documentation.

Issue #98, opened 2017-03-23
Improvements and future-proofing for Apache Drill init action


dennishuo:
A few ideas for making the Drill init action more friendly for debugging in case of failure and better future-proofing for version changes or different configurations of dependencies (like zookeeper being already running in HA mode):

- If server status still fails in drill, cat out /var/log/drill/* for easier debugging; possibly poll in a loop for up to a limited number of times for server to come up instead of failing later when trying tin add storage plugins in case the configuration is broken for some reason
- Find the GCS connector by wildcard for version-proofing
- Maybe derive Hive and HDFS settings from *-site.xml files
- Derive zookeeper node list and client port from /etc/zookeeper/conf
- Could be something like:

```
    ZOOKEEPER_CLIENT_PORT=$(grep clientPort /etc/zookeeper/conf/zoo.cfg | \
        cut -d '=' -f 2
    ZOOKEEPER_LIST=$(grep "^server\." /etc/zookeeper/conf/zoo.cfg | \
        cut -d '=' -f 2 | cut -d ':' -f 1 | sed "s/$/:${ZOOKEEPER_CLIENT_PORT}/" | \
        xargs echo | sed "s/ /,/g")
```

- Move drill profile to a subdirectory of dataproc bucket that has cluster uuid

dennishuo:
@vicenteg What do you think of these possible enhancements? Mostly I'm thinking of ensuring the init action supports "dataproc high-availability" mode and version updates for the GCS connector. FYI there's a handy tool called "bdconfig" available on Dataproc clusters as well to make it easier to pull out config values from XML configs, for example:

    bdconfig get_property_value --configuration_file /etc/hive/conf/hive-site.xml --name hive.metastore.uris 2>/dev/null

vicenteg:
These enhancements make sense to me. I'll work on it. Thanks!

PR #99, opened 2017-03-24
enhancements from #98

vicenteg:
Made the following changes recommended from #98:

- Find the GCS connector by wildcard for version-proofing
- Maybe derive Hive and HDFS settings from *-site.xml files
- Derive zookeeper node list and client port from /etc/zookeeper/conf
- Move drill profile to a subdirectory of dataproc bucket that has cluster uuid
- on a failure after starting the drillbit, cat the logs for troubleshooting


vicenteg:
I signed it! - I'm a googler.

dennishuo:
Nice, thanks! Tested cleanly with preexisting Hive tables and new drill cluster.

Issue #102, opened 2017-04-05
Jupyter setup failure since April 5


vadym-wix:
Hi, since today started to encounter jupyter setup failure.

 For the previous 2 months everything worked just fine

Now, I'm getting an error:

```
jupyter-1.0.0-   0% |                              | ETA:  --:--:--   0.00  B/s
jupyter-1.0.0- 100% |###############################| Time: 0:00:00   5.96 MB/s
jupyter-1.0.0- 100% |###############################| Time: 0:00:00   3.71 MB/s
Traceback (most recent call last):
  File "/usr/lib/google-cloud-sdk/platform/gsutil/gsutil", line 22, in <module>
    gsutil.RunMain()
  File "/usr/lib/google-cloud-sdk/platform/gsutil/gsutil.py", line 112, in RunMain
    import gslib.__main__
  File "/usr/lib/google-cloud-sdk/platform/gsutil/gslib/__main__.py", line 39, in <module>
    import boto
  File "/usr/lib/google-cloud-sdk/platform/gsutil/third_party/boto/boto/__init__.py", line 1216, in <module>
    boto.plugin.load_plugins(config)
  File "/usr/lib/google-cloud-sdk/platform/gsutil/third_party/boto/boto/plugin.py", line 93, in load_plugins
    _import_module(file)
  File "/usr/lib/google-cloud-sdk/platform/gsutil/third_party/boto/boto/plugin.py", line 75, in _import_module
    return imp.load_module(name, file, filename, data)
  File "/usr/lib/python2.7/dist-packages/google_compute_engine/boto/boto_config.py", line 29, in <module>
    from google_compute_engine import config_manager
ImportError: No module named google_compute_engine
```

dennishuo:
I got stuck at first on the `ImportError: No module named google_compute_engine` error since I couldn't figure out how it ever worked :)

Basically, the clobbering of `PATH` to have conda first always breaks `gsutil`, so the little `gsutil stat` trick trying to fetch notebooks from GCS just never actually worked (but also was untested and apparently unused).

But it's in a conditional so that's actually just a red herring. @nehalecky is correct that https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/103 should resolve this.

dennishuo:
Merged https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/103 - should work out of the box once again.

nehalecky:
[...]

PR #103, opened 2017-04-05
BUG: (DS-699) Add —allow-root flag to jupyter command runs for config…

nehalecky:
… generation and launching, supporting changes in Jupyter 5.0.0

### DIFF
--- a/jupyter/internal/launch-jupyter-kernel.sh
+++ b/jupyter/internal/launch-jupyter-kernel.sh
@@ -2,5 +2,5 @@
 set -e
 
 echo "Starting Jupyter notebook..."
-nohup jupyter notebook --no-browser > /var/log/jupyter_notebook.log 2>&1 &
+nohup jupyter notebook --allow-root --no-browser > /var/log/jupyter_notebook.log 2>&1 &
 
--- a/jupyter/internal/setup-jupyter-kernel.sh
+++ b/jupyter/internal/setup-jupyter-kernel.sh
@@ -18,7 +18,7 @@ echo "Creating notebook directory $JUPYTER_NOTEBOOK_DIR if necessary..."
 [[ ! -d $JUPYTER_NOTEBOOK_DIR ]] && mkdir -p $JUPYTER_NOTEBOOK_DIR
 
 echo "Creating Jupyter config..."
-jupyter notebook --generate-config -y --ip=127.0.0.1
+jupyter notebook --allow-root --generate-config -y --ip=127.0.0.1
 echo "c.Application.log_level = 'DEBUG'" >> ~/.jupyter/jupyter_notebook_config.py
 echo "c.NotebookApp.ip = '*'" >> ~/.jupyter/jupyter_notebook_config.py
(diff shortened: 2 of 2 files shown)


nehalecky:
might also resolve #102?

dennishuo:
Ah, indeed, good catch. Thanks! I'll test and merge as soon as possible.

dennishuo:
Tested cleanly:

    gcloud dataproc clusters create dhuo-jupyter-fixed \
        --initialization-actions gs://dataproc-initialization-actions/jupyter/jupyter.sh \
        --metadata "INIT_ACTIONS_REPO=https://github.com/bomboradata/dataproc-initialization-actions.git,INIT_ACTIONS_BRANCH=fix-jupyter-5.0.0-config-and-launch"

Thanks for sending the fix!

nehalecky:
@dennishuo, np! Glad to help out. This morning, when my analysis cluster didn't start, I knew it was likely related to something around the Jupyter 5.0.0 update. Sorry I didn't get it out sooner!

devpog:
Hi there,

Doesn't seem like modules are being properly loaded with pyspark kernel, although the init process boots ok now
screen shot 2017-04-06 at 11 10 25 am

nehalecky:
@devpog, that's correct (not due to Jupyter update) and unfortunate. Like many Py3 code bases, this is due to the removal of tuple unpacking in Python 3.6, which not one of us understands. :/

On the dataproc provisioning side, the issue manifests due to the most version of miniconda3 (e.g. the env var leverages in `bootstrap-conda.sh` defaulting to `MINICONDA_VER`='latest'`, which in turn now defaults to Python 3.6. To circumvent this, you have a few options available within use of this repo:
1)  explicitly target an older miniconda version when provisioning with `bootstrap-conda.sh` via the evn var as `export MINICONDA_VER='4.2.12'`), which currently resolves to Python 3.5.2
2) passing in an additional init action script that creates and uses a custom conda environment, which has python set  (e.g., `python<=3.5.3`)
3) My least favorite option, because it's a step backwards: set miniconda variant to 2 via `export MINICONDA_VARIANT='2'`, which resolves to install most recent version of Python 2 (i.e., `2.7.13`). Note, this won't this won't help if your code base is written in Python 3.

HTH, and if needed, I can probably find some time to submit a PR that has a sample script, using option 1 above (and later a combination of 2). :)

devpog:
@nehalecky thanks! was thinking about 3.5 conda's env )

dennishuo:
I'm going to take a stab at reintroducing the metadata options for MINICONDA_VARIANT and MINICONDA_VERSION (it was broken before due to using "-v" for variable being set instead of "-z" for variable being empty whether or not it was set), and also pin default to 4.2.12 for now.

dennishuo:
FYI - merged https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/106/files

nehalecky:
@dennishuo, great. Until it is fixed upstream in Spark, that's definitely the best solution. Thanks much!

PR #106, opened 2017-04-07
Pin default miniconda version to 4.2.12.

dennishuo:
-Allow setting MINICONDA_VARIANT by metadata again
-Fix variable existence checks to accommodate for this
-Also add MINICONDA_VERSION as metadata option
-Rename MINICONDA_VER to MINICONDA_VERSION

### DIFF
--- a/conda/bootstrap-conda.sh
+++ b/conda/bootstrap-conda.sh
@@ -7,6 +7,9 @@ set -e
 #ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
 #if [[ "${ROLE}" == 'Master' ]]; then
 
+MINICONDA_VARIANT=$(/usr/share/google/get_metadata_value attributes/MINICONDA_VARIANT || true)
+MINICONDA_VERSION=$(/usr/share/google/get_metadata_value attributes/MINICONDA_VERSION || true)
+
 if [[ ! -v CONDA_INSTALL_PATH ]]; then
     echo "CONDA_INSTALL_PATH not set, setting ..."
     CONDA_INSTALL_PATH="/opt/conda"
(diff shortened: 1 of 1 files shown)


surjikal:
Just fyi this just broke my stuff, for some reason my pyspark jobs are using python3 now. Trying to find a fix now and will post here.

surjikal:
Alright I was setting `MINICONDA_VARIANT` through an environment variable in a customized initialization-action script.

Like this:
```bash
MINICONDA_VARIANT=2 ./dataproc-initialization-actions/conda/bootstrap-conda.sh
```

But that doesn't work anymore. I am now setting `MINICONDA_VARIANT` with a metadata option.

dennishuo:
Sorry about that, I figured it'd be safe since prior to merging https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/104/files there was already the `MINICONDA_VARIANT=$(/usr/share/google/get_metadata_value attributes/MINICONDA_VARIANT || true)` line which is what ends up "blocking" the environment variable settings from a caller, so the behavior should've been the same as prior to yesterday.

Sounds like in your case it was a series of coincidences:

1. You had been exporting MINICONDA_VARIANT=2 since before yesterday, but it was actually a no-op due to the metadata fetch line causing it to be an empty string instead, and then the bug of using "-v MINICONDA_VARIANT" in the presence of empty strings made it point at a non-python-version-prefixed install, e.g. "Miniconda-latest.*" which happens to map to Python 2.
2. Yesterday, the cleanup patch got merged, which removed the metadata line, such that your env variable started actually being used. So it started explicitly specifying "Miniconda2-latest".
3. My update, despite preserving the intended default of Python 3, thus broke this case by "blocking" the  environment variable.

Generally using metadata options is probably going to be the preferred route though as we progress towards official support of init actions, where users ideally won't need to maintain separate forks too much and instead and customize options through metadata values.

In your case, is the metadata option a reasonable long-term solution?

surjikal:
Yeah metadata works fine! Not sure why I was using an environment variable for this, since I was already using metadata for `PYTHON_PACKAGES`...

devpog:
@dennishuo Hi Dennis, 

What is the actual fix for juniper.sh?
When I spin a cluster, providing an init action file juniper.sh, it goes well. But when I actually try to use Jupyter, invoking something primitive such as spark.sql(" select * from parquet.'...' "), I get the following error: 

NameError                                 Traceback (most recent call last)
<ipython-input-1-0fbbf4ecf81f> in <module>()
----> 2 imps = spark.sql("select * from parquet.`gs://...`")

NameError: name 'spark' is not defined

Which was not the case last Friday, when I simply forked a repo under my account and created a branch with fixed MINICONDA_VER=4.2.12

How does it work now?


Thanks!

dennishuo:
Hmm, strange; I tested this change with out-of-the-box settings as part of fixing https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/105 - are you passing any other settings like a specific `--image-version` or other metadata values? Are you using the hosted jupyter.sh, or your own version of it?

devpog:
Hi @dennishuo 

Thanks for following up. Yes I do pass extra options, such as below:

gcloud dataproc clusters create cl01-jupyter --bucket=*** --zone us-east1-d \
--metadata "JUPYTER_CONDA_PACKAGES=numpy:pandas:scikit-learn:matplotlib,INIT_ACTIONS_REPO=https://github.com/***/dataproc-initialization-actions.git,INIT_ACTIONS_BRANCH=miniconda_4.2.12" \
--initialization-actions gs://***/dataproc-initialization-actions/jupyter/jupyter.sh \
--properties \
spark:spark.executorEnv.PYTHONHASHSEED=0,\
spark:spark.executor.extraJavaOptions=-XX:+UseG1GC,\
spark:spark.dynamicAllocation.enabled=false,\
spark:spark.executor.cores=7,\
spark:spark.executor.memory=8g,\
spark:spark.yarn.executor.memoryOverhead=4g,\
spark:spark.driver.memory=8g,\
spark:spark.driver.maxResultSize=6g,\
spark:spark.kryoserializer.buffer.max=1g,\
spark:spark.yarn.driver.memoryOverhead=4g,\
spark:spark.driver.cores=7,\
spark:spark.master=yarn,\
spark:spark.master.DeployMode=cluster,\
spark:spark.executor.instances=28 \
--image-version=1.1 \
--master-machine-type=n1-standard-4 \
--worker-machine-type=custom-8-40960 \
--master-boot-disk-size=10GB \
--worker-boot-disk-size=10GB \
--num-workers=4 \
--num-worker-local-ssds 4 \
--project ***


It works because I use a repo that I forked last week. Also I have noticed when I pass options for dynamic allocation, it failed to recognize spark sessions, giving me the error that I mentioned earlier e.g:
...
spark:spark.executorEnv.PYTHONHASHSEED=0,\
spark:spark.executor.extraJavaOptions=-XX:+UseG1GC,\
spark:spark.shuffle.service.enabled=true,\
spark:spark.dynamicAllocation.enabled=true,\
spark:spark.dynamicAllocation.minExecutors=4,\
spark:spark.dynamicAllocation.maxExecutors=48,\
spark:spark.executor.cores=7,\
spark:spark.yarn.executor.memoryOverhead=4g,\
spark:spark.driver.maxResultSize=6g,\
spark:spark.kryoserializer.buffer.max=1g,\
spark:spark.yarn.driver.memoryOverhead=4g,\
spark:spark.driver.cores=7,\
spark:spark.master=yarn,\
spark:spark.master.DeployMode=cluster \
...

PR #109, opened 2017-04-28
Fix Kafka data loss on reboot

ipostanogov:
The default log (data) folder gets wiped on reboot in Debian/Ubuntu. It is better to move it to somewhere else (see http://stackoverflow.com/a/32439885).


dennishuo:
Thanks for the contribution! FYI we're probably going to polish this up to use a cleaner bigtop-based install in the near future, but this kind of configuration will be relevant nonetheless.

PR #121, opened 2017-06-07
Gate flink yarn session on GCE metadata

bsidhom:
This allows the default behavior (whether or not a detached yarn session should be started) to be overridden based on the value of attributes/flink-start-yarn-session.

This change also fixes the jobmanager RPC address. This value was previously not being filled because the MASTER_HOSTNAME environment variable was unbound.


AngusDavis:
LGTM

PR #122, opened 2017-06-08
Only install Flink on primary master

bsidhom:
The Flink init action currently installs Flink on all masters and attempts to start a YARN session on all masters. Because this default session is configured to consume all cluster resources, all but one of the master init actions hang indefinitely while trying to allocate containers.

This change only installs and runs Flink on the primary master.

### DIFF
--- a/flink/flink.sh
+++ b/flink/flink.sh
@@ -112,7 +112,9 @@ EOF
 
 function main() {
 local role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
-if [[ "${role}" == 'Master' ]]; then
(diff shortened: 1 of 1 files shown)


bsidhom:
Alternatively, we could _install_ Flink on all masters, but only run the detached session from the primary one.

AngusDavis:
> Alternatively, we could install Flink on all masters, but only run the detached session from the primary one.

I like this idea - it prevents any master from being a pet.

bsidhom:
Done.

AngusDavis:
LGTM

PR #124, opened 2017-06-19
Fixes the "too many symlinks" automounting error

jmthibault79:


jmthibault79:
signed

datalab/datalab.sh
```diff
@@ -85,16 +84,14 @@ EOF
...
-  for i in {1..10}; do
-    if docker run -d --net=host -v "${DATALAB_DIR}:/content/datalab" ${VOLUME_FLAGS} \
-        datalab-pyspark; then
-      echo 'Cloud Datalab Jupyter server successfully deployed.'
-      exit
-    fi
-    sleep 1
-  done
+  # Docker gives a "too many symlinks" error if volumes are not yet automounted.
+  # Ensure that the volumes are mounted to avoid the error.
```
karth295: I'm a little confused -- does `touch ${VOLUMES}` allow you to safely mount the directories when invoking `docker run`? I couldn't find a related docker bug.

jmthibault79: `docker run` requires directories to be mounted.  `touch` does that mounting.

https://serverfault.com/questions/640895/why-do-some-host-volumes-in-docker-containers-give-the-error-too-many-levels-of

karth295: Ah, thanks.

PR #144, opened 2017-09-28
Deprecate the ipython-notebook init action in favor of Jupyter Notebook

### LINKED ISSUE #143: ipython-notebook init action does not work
The `ipython notebook` run command fails with the error message:

```
[C 22:02:58.810 NotebookApp] Running as root is not recommended. Use --allow-root to bypass.
```

When I add `--allow-root`, it starts, but then prints out a token that I have to use to log in.

```
    Copy/paste this URL into your browser when you connect for the first time,
    to login with a token:
        http://localhost:8123/?token=4d404beccaafb9605e5663dff0349341bae4bebdd23073d8
```

Since the init action is run with `nohup`, it's stderr (with these messages) is completely lost. Only stdout is redirected to a file.

It seems like this init action should just be removed since we have a jupyter init action.

karth295:
Close #143


dennishuo:
LGTM

PR #148, opened 2017-09-29
Update README with single node information

karth295:
Importantly, document init actions that do not work on single node clusters.


evilsoapbox:
lgtm!

Issue #131, opened 2017-08-24
Starting Presto server fails in Dataproc 1.2


karth295:
In `/var/presto/data/var/log/launcher.log`:

```
Invalid maximum heap size: -Xmxm
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

The script sets:

```
-Xmx${PRESTO_JVM_MB}m
```

Where PRESTO_JVM_MB comes from spark properties parsed from spark-defaults.conf: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/presto/presto.sh#L53

The init action assumes spark-defaults.conf has `key=value` pairs, but does not accept `key value` with a space in the middle. Looking at my (unchanged) spark-defaults.conf in Dataproc 1.2:

```
spark.yarn.executor.memoryOverhead 465
```

```
$ grep spark.yarn.executor.memoryOverhead /etc/spark/conf/spark-defaults.conf | tail -n 1 | cut -d '=' -f 2
spark.yarn.executor.memoryOverhead 465
```

PR #151, opened 2017-10-13
Use sed instead of cut to extract Spark settings

rekotan:
In /etc/spark/conf/spark-defaults.conf, some settings use "=" and other use
whitespace to define Spark's settings. With the Dataproc image
version upgrade, one of them changed to use whitespace and that
prevented this script from working properly. To add a bit of robustness
to the script, use regex to make sure this works both "=" and
whitespace. In addition, if spark.yarn.executor.memoryOverhead is not
defined, it uses the default Spark property value.


rekotan:
I signed it!

karth295:
Ah nice, this fixes #131

PR #154, opened 2017-11-22
Write GCS-provided Jupyter notebook changes back to GCS

waltherg:
Dear maintainers. Thanks very much for these startup scripts!

The proposed change ensures that Jupyter notebook changes are pushed back to GCS for those notebooks that were provided through GCS to begin with.
Please advise whether this behavior may be desired within the scope of this startup script.


karth295:
This change seems reasonable to me given the current way of uploading notebooks.

But consider trying https://github.com/src-d/jgscm as an alternative. I haven't tried it myself, but it loads/saves directly to/from GCS, which is better than our current hackery of copying files around. There is a caveat in that project's readme that it can make the Jupyter UI slow -- which may be a dealbreaker. If you do try that out, let us know what you find!

Otherwise, I'm happy to merge this PR.

PR #156, opened 2017-11-27
Only store Jupyter notebooks in GCS.

karth295:
This is a counterproposal to #154 (cc @waltherg). Instead of copying the files back and forth to GCS, just use this [cool project](https://github.com/src-d/jgscm) to read/save files directly in GCS.

CC @vmarkovtsev as the maintainer of jgscm. You mention that jgscm is in in alpha state -- I just want to check whether you have any concerns about this PR. For what it's worth, at least Python3 "hello world" worked for me.

### DIFF
--- a/jupyter/README.MD
+++ b/jupyter/README.MD
@@ -55,7 +55,7 @@ gcloud dataproc clusters create my-dataproc-cluster \
   - configures `jupyter` to use auth token `JUPYTER_AUTH_TOKEN`, with a default of none.
       - Dataproc does not recommend opening firewall ports to access Jupyter, but rather using a proxy. See [connecting to web interfaces](https://cloud.google.com/dataproc/docs/concepts/cluster-web-interfaces)
       - This proxy access is automated by [launch-jupyter-interfaces.sh](#launch-jupyter-interfacesh).
-  - targets a default notebooks directory `/root/notebooks`, into which the directory found at `gs://$DATAPROC_BUCKET/notebooks/` is copied, where `$DATAPROC_BUCKET` is the value stored in the metadata key `dataproc-bucket` (set by default upon cluster creation and overridable)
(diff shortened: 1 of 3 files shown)


vmarkovtsev:
@karth295 jgscm is listed as alpha because it works without any issues for my company and some other people, but I haven't gathered any massive feedback yet. AFAIK listing very large directories (> 1000 files) can be slow sometimes.

I am always online and watching for the issues, so it is maintained and alive. Will be happy to make it beta if this PR is accepted.

waltherg:
Excellent @karth295 - if jgscm holds up this makes far more sense than the hacky workaround we had in `jupyter.sh`.

dennishuo:
+1, sounds like a good idea. One nit though is we should be moving toward more hermetic setups (even though there's existing legacy stuff throughout our init actions examples that don't do so at the moment), so let's start with mirroring some pinned version(s) of the packages into GCS and pip install from there by default. We can still make efforts to proactively sync up mirrored packages as updates are made upstream.

This would help save latency and bandwidth avoiding having to fetch things from outside Google's network during cluster deployments, and once we move all the package installs to point to artifacts mirrored in GCS we'll also then support "private-IP only" clusters where nodes have no external connectivity at all, and UI access could be controlled behind something like [Google Cloud Identity-Aware Proxy](https://cloud.google.com/iap/docs/). This is in-line with many of the other existing init actions which use Bigtop-based packages and thus install from Dataproc's deb packages hosted in GCS.

karth295:
Talked offline and decided to just pin to a version of jgscm for now, so that the init action is hermetic. We'll mirror packages later.

dennishuo:
LGTM

karth295:
Sorry for the delay @waltherg and @vmarkovtsev it's live now!

PR #169, opened 2018-01-24
Run apt-get update on all init actions with retries

### LINKED ISSUE #168: Init actions occasionally fail on `apt-get update`
Looking at a failure to run the datalab init action, for example:

```
+ apt-get update
...
W: Failed to fetch http://repo.stackdriver.com/apt/dists/jessie/main/binary-amd64/Packages  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.
```

Apparently this is an issue in APT itself -- the hash file and package are two separate files, so when one is being updated, there is a race where the hashes will not match.

More information: https://blog.packagecloud.io/eng/2016/09/27/fixing-apt-hash-sum-mismatch-consistent-apt-repositories/

Since Debian 8 does not have this new version of apt, the easiest thing to do is just re-run `apt-get update` until it succeeds.

karth295:
Fixes #168 (init actions occasionally fail because `apt-get update` fails)

I added `update_apt_get` to all init actions that run `apt-get install`, as we have also seen occasional init action failures because apt-get update was not run.

### DIFF
--- a/datalab/datalab.sh
+++ b/datalab/datalab.sh
@@ -18,13 +18,23 @@
 
 set -e -x
 
+function update_apt_get() {
+  for ((i = 0; i < 10; i++)); do
+    if apt-get update; then
+      return 0
+    fi
+    sleep 5
+  done
+  return 1
+}
+
 ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
 PROJECT=$(/usr/share/google/get_metadata_value ../project/project-id)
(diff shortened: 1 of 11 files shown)


karth295:
(Note to self: DO NOT MERGE UNTIL PR IS TESTED)

karth295:
Alright, I tested all of them.

AngusDavis:
LGTM - thanks for the nudge.

Issue #174, opened 2018-01-31
Installing Jupyter with "--metadata 'MINICONDA_VARIANT=2" breaks gsutil


spark2ignite:
When installing Jupyter using initialisation action and specifying --metadata 'MINICONDA_VARIANT=2, it breaks Google's gsutil. Without "MINICONDA_VARIANT=2" everything works fine.

How to reproduce:
```
gcloud dataproc --region us-east1 clusters create jupytertest --subnet 'projects/operations-1266/regions/us-east1/subnetworks/xpn-us-east1-shared1' --no-address --image-version 1.2 --project research-cluster-186207 --initialization-actions 'gs://dataproc-initialization-actions/jupyter/jupyter.sh' --metadata 'MINICONDA_VARIANT=2'
```

Will result in this:
![image](https://user-images.githubusercontent.com/5726975/35646357-9facaea6-06d7-11e8-874e-043b30f6a54a.png)

karth295:
Weird. If you just `pip install google-compute-engine` it's okay, so somehow gsutil's environment got destroyed when switching to conda.

As a temporary workaround, you can use an init action to run `/opt/conda/bin/pip install google-compute-engine`. (It's not in the root user's $PATH by default, so use the absolute path)

karth295:
For posterity, folks who like Python 2 can use the `jupyter2` init action: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/tree/master/jupyter2

karth295:
The alternative fix is to `export CLOUDSDK_PYTHON=/usr/bin/python`

PR #180, opened 2018-02-20
Proposal: A simpler Jupyter init action (design review wanted)

karth295:
The Jupyter init action does its job, but is a little confusing to use:

* It invokes the Conda init action(s) by git cloning this repo. It also invokes sub-scripts through this cloned copy of the repo. So any changes to the sub-scripts involve pointing the main init action script at a different git repository.
* It switches to Python 3. Because of this you lose all of the packages pre-installed by Dataproc, like `numpy`. So a script that works in `gcloud dataproc jobs submit pyspark` does not work out of the box in Jupyter. Issue #174 is an example of when this bites you.
  * Note that Jupyter comes with the `six` library, which lets you run most python 2 or 3 code against either version of python. So leaving python at python2 hopefully shouldn't be a big deal for python3 users.
* The jupyter "kernel" for pyspark does not actually invoke the `pyspark` command, so misses out on some of the configuration in `spark-env.sh`, for example. If you set `PYTHONPATH=...` in spark-env.sh, it works for pyspark, but not jupyter.

So this PR is my proposal for a simpler, one-file-only init action (credit to https://blog.sicara.com/get-started-pyspark-jupyter-guide-tutorial-ae2fe84f594f):

* Use `pip` to avoid invoking the Conda init action. Pip is probably more familiar for folks anyway. Jupyter supports installation through pip, even if it recommends using Conda.
* Stay on python2
* Use PYSPARK_DRIVER_PYTHON=jupyter. This makes Jupyter just like Zeppelin, which uses `pyspark` under the hood.
* Continues to save notebooks to GCS in the same location

Some things that my proposal does not support, but do work right now:

* `JUPYTER_CONDA_PACKAGES` to install extra packages. I do see value in this btw -- and we could support a similar flag for installing pip packages.
* Installing IPyNB extensions and RISE (turning notebooks into presentations). These both seem pretty cool, but are sort of fringe use cases that people can add if they need.
* Separate interpreters for "Python 2" and "PySpark". In my proposal, there is only an interpreter called "Python 2" that's actually just the `pyspark` shell (I'll admit that's a little confusing).
* Installing `com.databricks:spark-csv_2.10:1.3.0` for Spark < 2.0. That seems like something that should be in Dataproc itself, not in this init action.

This would obviously be a very breaking change. Maybe this should be a separate `jupyter-pyspark` init action?

I'll write a README if this looks good to all of you ( @AngusDavis @pmkc @dennishuo )

### DIFF
--- a/jupyter2/jupyter2.sh
+++ b/jupyter2/jupyter2.sh
@@ -2,16 +2,14 @@
 
 set -euxo pipefail
 
-ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
-DATAPROC_BUCKET=$(curl -f -s -H Metadata-Flavor:Google http://metadata/computeMetadata/v1/instance/attributes/dataproc-bucket)
-JUPYTER_PORT=$(/usr/share/google/get_metadata_value attributes/JUPYTER_PORT || true)
-[[ ! $JUPYTER_PORT =~ ^[0-9]+$ ]] && JUPYTER_PORT=8123
-JUPYTER_AUTH_TOKEN=$(/usr/share/google/get_metadata_value attributes/JUPYTER_AUTH_TOKEN || true)
+readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+readonly DATAPROC_BUCKET="$(/usr/share/google/get_metadata_value attributes/dataproc-bucket)"
+readonly JUPYTER_AUTH_TOKEN="$(/usr/share/google/get_metadata_value attributes/JUPYTER_AUTH_TOKEN || true)"
+readonly JUPYTER_PORT="$(/usr/share/google/get_metadata_value attributes/JUPYTER_PORT || echo 8123)"
 
-# Only run on master
-if [[ "${ROLE}" != 'Master' ]]; then
-  exit 0;
-fi
+# GCS directory in which to store notebooks.
+# DATAPROC_BUCKET must not include gs:// and bucket must exist
+readonly NOTEBOOK_DIR="${DATAPROC_BUCKET}/notebooks"
 
 function update_apt_get() {
   for ((i = 0; i < 10; i++)); do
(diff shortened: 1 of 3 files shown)


karth295:
Oooh "JupyterLab" (a more complete dev environment) was just released to beta today: https://news.ycombinator.com/item?id=16420142. Might be worth looking into since we're significantly changing the init action anyway.

pmkc:
It looks good to me. I would stick with Jupyter for now at least.

Also interested in Angus's and Dennis's feedback.

karth295:
Latest commit adds matplotlib, similar to #184

karth295:
Offline from Dennis: let's make this a separate init action to avoid breaking existing users. Even if users had copied jupyter.sh into their own bucket, that script by default clones this repo to find its subscripts.

karth295:
Alright, the latest commit adds a `jupyter2` init action. It adds a note to the `jupyter` init action that Python 2 and `pip` users should use `jupyter2`, but does not mark is as legacy/deprecated.

jupyter2/jupyter2.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
pmkc: This should probably follow the new style guide

jupyter2/jupyter2.sh
```diff
@@ -0,0 +1,75 @@
...
+
+set -euxo pipefail
+
+ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
```
DariuszAniszewski: all constants here should probably be declared as `readonly` and quoted

jupyter2/jupyter2.sh
```diff
@@ -0,0 +1,75 @@
...
+DATAPROC_BUCKET=$(curl -f -s -H Metadata-Flavor:Google http://metadata/computeMetadata/v1/instance/attributes/dataproc-bucket)
```
DariuszAniszewski: this can be obtained via `get_metadata_value`, no need for `curl`

jupyter2/jupyter2.sh
```diff
@@ -0,0 +1,75 @@
...
+  return 1
+}
+
+update_apt_get
+apt-get install -y python-dev python-tk
+easy_install pip
+pip install --upgrade jupyter jgscm matplotlib
+
+# Must not include gs:// and bucket must exist
+NOTEBOOK_DIR="$DATAPROC_BUCKET/notebooks"
```
DariuszAniszewski: this can be moved to top and marked as `readonly`

jupyter2/jupyter2.sh
```diff
@@ -0,0 +1,75 @@
...
+EOF
+echo "Jupyter setup!"
+
+INIT_SCRIPT="/usr/lib/systemd/system/jupyter.service"
+cat << EOF > ${INIT_SCRIPT}
+[Unit]
+Description=PySpark-based Jupyter Notebook Server
+After=hadoop-yarn-resourcemanager.service
+
+[Service]
```
DariuszAniszewski: consider adding `Restart=on-failure`

karth295:
@DariuszAniszewski done -- PTAL

DariuszAniszewski:
@karth295 Fine for me, thanks.

karth295:
From @pmkc offline: we should make a separate Jupyter kernel (which under the hood runs `pyspark`). There are a couple benefits: a) The UI clearly differentiates between vanilla "Python 2" and "PySpark", b) we don't always leave a `pyspark` driver running (and thus consuming YARN resources).

The latest commit refactors the init action a bit to create a Jupyter kernel, yet still runs `pyspark` so picks up `spark-env.sh`.

PR #181, opened 2018-02-21
Add BigDL init action

karth295:


karth295:
Hah this doesn't actually work out of the box on 1.0 (as the README implies). Maybe I should just remove that note and say this only works on 1.2.

karth295:
The latest commits 1) Fix the init action for Dataproc 1.0, 2) update/fix the READMEs.

bigdl/bigdl.sh
```diff
@@ -0,0 +1,72 @@
...
+cd /opt/intel-bigdl
+
+wget "${BIGDL_DOWNLOAD_URL}"
+unzip *.zip
+
+JAR=`realpath lib/*.jar`
+PYTHON_ZIP=`realpath lib/*.zip`
+
+echo "SPARK_DIST_CLASSPATH=\"\$SPARK_DIST_CLASSPATH:$JAR\"" >> /etc/spark/conf/spark-env.sh
+echo "PYTHONPATH=\"\$PYTHONPATH:$PYTHON_ZIP\"" >> /etc/spark/conf/spark-env.sh
```
pmkc: nit: cat << EOF >> spark-env.sh is more legible

bigdl/bigdl.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# Config changes only need to happen on the master.
+# This way we also avoid DOSing the resource manager when running yarn node -list
+
+if [[ "${ROLE}" == "Master" ]]; then
+  cat conf/spark-bigdl.conf >> /etc/spark/conf/spark-defaults.conf
+  echo "spark.dynamicAllocation.enabled=false" >> /etc/spark/conf/spark-defaults.conf
```
pmkc: You could also combine the spark-defaults into a cat << EOF

PR #185, opened 2018-02-22
More meaningful error handling in Flink. Made Flink script consistent with Shell Sty…

adam-stasiak:
Changes:
 1. Moved functions to the top of the file just below constants declaration.
 2. Added err() function for printing more detailed error info.
 3. Moved start_flink_yarn_session to boolean instead of string.
 4. Hdfs usage replaced with recommended bdconfig.
 5. Added "." in comments to make script consistent.

### DIFF
--- a/flink/flink.sh
+++ b/flink/flink.sh
@@ -137,9 +137,9 @@ EOF
 function main() {
 local role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
 if [[ "${role}" == 'Master' ]] ; then
-  update_apt_get || (err "Unable to update apt-get" && exit 1)
-  apt-get install -y flink || (err "Unable to install flink" && exit 1)
-  configure_flink || (err "Flink configuration failed" && exit 1)
+  update_apt_get || err "Unable to update apt-get"
+  apt-get install -y flink || err "Unable to install flink"
+  configure_flink || err "Flink configuration failed"
 fi
 }
 
--- a/flink/flink.sh
+++ b/flink/flink.sh
@@ -46,6 +46,7 @@ readonly START_FLINK_YARN_SESSION_DEFAULT=true
 
 function err() {
   echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
+  return 1
 }
 
 function update_apt_get() {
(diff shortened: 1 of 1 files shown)


flink/flink.sh
```diff
@@ -70,39 +74,41 @@ function configure_flink() {
...
+  # Determine the default parallelism.
   local flink_parallelism=$(python -c \
-      "print ${num_taskmanagers} * ${flink_taskmanager_slots}")
+    "print ${num_taskmanagers} * ${flink_taskmanager_slots}")
 
   # Get worker memory from yarn config.
-  local worker_total_mem="$(hdfs getconf \
-    -confKey yarn.nodemanager.resource.memory-mb)"
+  local worker_total_mem="$(bdconfig get_property_value --configuration_file \
+    /etc/hadoop/conf/yarn-site.xml \
```
bsidhom: This is strictly more brittle than using hdfs getconf as it depends on a hard-coded hadoop conf path. Why redirect stderr to /dev/null? If we want better error messaging, this should be captured with normal stderr.

flink/flink.sh
```diff
@@ -70,39 +74,41 @@ function configure_flink() {
...
+    "print int(${worker_total_mem} * ${FLINK_JOBMANAGER_MEMORY_FRACTION})")
   local flink_taskmanager_memory=$(python -c \
-      "print int(${worker_total_mem} * ${FLINK_TASKMANAGER_MEMORY_FRACTION})")
+    "print int(${worker_total_mem} * ${FLINK_TASKMANAGER_MEMORY_FRACTION})")
 
   # Fetch the primary master name from metadata.
   local master_hostname="$(/usr/share/google/get_metadata_value attributes/dataproc-master)"
   local hostname="$(hostname)"
 
+  # Variable for check if yarn session launch is required.
```
bsidhom: I don't think that this comment increases readability.

flink/flink.sh
bsidhom: @dennishuo What do you think about using `bdutil` with a hard-coded configuration file rather than `hdfs getconf`?

dennishuo: I agree hdfs getconf is probably better here; I think the confusion stems from my guidance in https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/CONTRIBUTING.md saying to use bdconfig instead of grep or sed against the XML files; in general bdconfig is still preferred vs string-replace hacks, but for cases where we can use Hadoop's tooling, that's indeed better than bdconfig. There are some edge cases where we might need to get values from specific files instead of relying on implicit import ordering of different xml files though, such as when we want to inspect distcp-default.xml and its overrides vs mapred-site.xml.

bsidhom: Ah, that makes sense. I've updated the contributing guide to mention Hadoop tools in #189.

adam-stasiak:
@bsidhom  what do you think now? I removed not needed comment and restored previous way of  worker_total_mem assignment.

PR #188, opened 2018-02-22
Kafka script updated

szewi:
…to follow https://google.github.io/styleguide/shell.xml. 

As described in https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/187

### DIFF
--- a/kafka/kafka.sh
+++ b/kafka/kafka.sh
@@ -18,6 +18,8 @@
 
 set -euxo pipefail
 
+readonly KAFKA_PROP_FILE='/etc/kafka/conf/server.properties'
+
 function update_apt_get() {
   for ((i = 0; i < 10; i++)); do
     if apt-get update; then
@@ -30,17 +32,14 @@ function update_apt_get() {
 
 function err() {
   echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
+  return 1
 }
 
 function install_and_configure_kafka_server() {
   # Find zookeeper list first, before attempting any installation.
-  local ZOOKEEPER_CLIENT_PORT
-  local ZOOKEEPER_LIST
-  readonly KAFKA_PROP_FILE='/etc/kafka/conf/server.properties'
-
-  ZOOKEEPER_CLIENT_PORT=$(grep 'clientPort' /etc/zookeeper/conf/zoo.cfg \
+  local ZOOKEEPER_CLIENT_PORT=$(grep 'clientPort' /etc/zookeeper/conf/zoo.cfg \
     | cut -d '=' -f 2)
-  ZOOKEEPER_LIST=$(grep '^server\.' /etc/zookeeper/conf/zoo.cfg \
+  local ZOOKEEPER_LIST=$(grep '^server\.' /etc/zookeeper/conf/zoo.cfg \
     | cut -d '=' -f 2 \
     | cut -d ':' -f 1 \
     | sed "s/$/:${ZOOKEEPER_CLIENT_PORT}/" \
@@ -58,12 +57,11 @@ function install_and_configure_kafka_server() {
   # If all attempts failed, error out.
   if [[ -z "${ZOOKEEPER_LIST}" ]]; then
     err 'Failed to find configured Zookeeper list; try --num-masters=3 for HA'
-    exit 1
   fi
 
   # Install Kafka from Dataproc distro.
-  apt-get install -y kafka-server \
-    || (err 'Unable to install kafka-server' && exit 1)
+  apt-get install -y kafka-server || dpkg -l kafka-server \
+    || err 'Unable to install and find kafka-server on worker node.'
 
   mkdir -p /var/lib/kafka-logs
   chown kafka:kafka -R /var/lib/kafka-logs
@@ -85,20 +83,19 @@ function install_and_configure_kafka_server() {
 
 function main() {
   local ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
-  update_apt_get || (err 'Unable to update packages lists.' && exit 1)
+  update_apt_get || err 'Unable to update packages lists.'
 
   # Only run the installation on workers; verify zookeeper on master(s).
   if [[ "${ROLE}" == 'Master' ]]; then
     service zookeeper-server status \
-      || (err 'Required zookeeper-server not running on master!' && exit 1)
+      || err 'Required zookeeper-server not running on master!'
     # On master nodes, just install kafka libs but not kafka-server.
     apt-get install -y kafka \
-      || (err 'Unable to install kafka libraries on master!' && exit 1)
-    exit 0
+      || err 'Unable to install kafka libraries on master!'
+  else
+    # Run installation on workers.
+    install_and_configure_kafka_server
   fi
-
-  # Run installation on workers.
-  install_and_configure_kafka_server
 
 }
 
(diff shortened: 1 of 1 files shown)


szewi:
@dennishuo could you take a look ?

kafka/kafka.sh
```diff
@@ -24,49 +28,78 @@ function update_apt_get() {
...
-ZOOKEEPER_CLIENT_PORT=$(grep clientPort /etc/zookeeper/conf/zoo.cfg | cut -d '=' -f 2)
-ZOOKEEPER_LIST=$(grep "^server\." /etc/zookeeper/conf/zoo.cfg | \
-        cut -d '=' -f 2 | cut -d ':' -f 1 | sed "s/$/:${ZOOKEEPER_CLIENT_PORT}/" | \
-        xargs echo | sed "s/ /,/g")
-if [[ -z "${ZOOKEEPER_LIST}" ]]; then
-  # Didn't find zookeeper quorum in zoo.cfg, but possibly workers just didn't
-  # bother to populate it. Check if YARN HA is configured.
-  ZOOKEEPER_LIST=$(bdconfig get_property_value --configuration_file \
+function err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
```
bsidhom: It appears that this method is almost always followed by exit code 1. Since we're using `set -e` above, consider adding `return 1` at the bottom of this function. That will cause the entire script to exit with code 1 on failure. If you still want to use `err` as a non-terminal function, consider adding the `return 1` variant as a new function `errAndFail` or similar.

kafka/kafka.sh
```diff
@@ -24,49 +28,78 @@ function update_apt_get() {
...
+  update_apt_get || (err 'Unable to update packages lists.' && exit 1)
+
+  # Only run the installation on workers; verify zookeeper on master(s).
+  if [[ "${ROLE}" == 'Master' ]]; then
+    service zookeeper-server status \
+      || (err 'Required zookeeper-server not running on master!' && exit 1)
+    # On master nodes, just install kafka libs but not kafka-server.
+    apt-get install -y kafka \
+      || (err 'Unable to install kafka libraries on master!' && exit 1)
+    exit 0
```
bsidhom: Please use `if`/`else` rather than using `exit` for control flow.

kafka/kafka.sh
```diff
@@ -24,49 +28,78 @@ function update_apt_get() {
...
-  # Didn't find zookeeper quorum in zoo.cfg, but possibly workers just didn't
-  # bother to populate it. Check if YARN HA is configured.
-  ZOOKEEPER_LIST=$(bdconfig get_property_value --configuration_file \
+function err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
+}
+
+function install_and_configure_kafka_server() {
+  # Find zookeeper list first, before attempting any installation.
+  local ZOOKEEPER_CLIENT_PORT
```
bsidhom: Prefer initializing local variables immediately since they aren't initialized in `if`/`else` branches.

kafka/kafka.sh
```diff
@@ -24,49 +28,78 @@ function update_apt_get() {
...
+  local ZOOKEEPER_LIST
+  readonly KAFKA_PROP_FILE='/etc/kafka/conf/server.properties'
```
bsidhom: Please move constants to the top of the file.

kafka/kafka.sh
```diff
@@ -24,49 +28,78 @@ function update_apt_get() {
...
+  fi
+
+  # If all attempts failed, error out.
+  if [[ -z "${ZOOKEEPER_LIST}" ]]; then
+    err 'Failed to find configured Zookeeper list; try --num-masters=3 for HA'
+    exit 1
+  fi
+
+  # Install Kafka from Dataproc distro.
+  apt-get install -y kafka-server \
```
bsidhom: This is a change of behavior. `dpkg -l` appears to be checking if the package is already installed and allowing configuration to continue. Please keep the previous behavior or explain why this change is necessary.

kafka/kafka.sh
szewi: This perfectly makes sense to me to add return 1 at the bottom of err function, as I would like to terminate script after error.

kafka/kafka.sh
szewi: ok, wasn't sure how to handle this situation where KAFKA_PROP_FILE is only used by single function however it's a constant and we should be consistent here with moving it to the top of the file.

kafka/kafka.sh
szewi: Could you double check that as in [guide](https://google.github.io/styleguide/shell.xml?showone=Use_Local_Variables#Use_Local_Variables) there is: `Declaration and assignment must be separate statements when the assignment value is provided by a command substitution`

kafka/kafka.sh
szewi: Sure.

kafka/kafka.sh
szewi: This doesn't change behaviour as if kafka-server installs successfully, then `dpkg -l` will not even run, but when kafka-server installation fails that dpkg -l would be run and if it finds already installed packaged that it will allow to continue with initialization script execution.. I assumed that it's rather unlikely that machines would already have kafka-server installed and installation will fail in the same time, but I must admin that my change modifies behaviour. I will keep that old flow, but add modified err message.

kafka/kafka.sh
bsidhom: Sorry, I hadn't considered that these were set using command substitution. Please separate the declarations where this is the case.

kafka/kafka.sh
```diff
@@ -24,49 +30,73 @@ function update_apt_get() {
...
-if [[ $(/usr/share/google/get_metadata_value attributes/dataproc-role) == 'Master' ]]; then
-  service zookeeper-server status || \
-      (echo 'Required zookeeper-server not running on master!' && exit 1)
-  # On master nodes, just install kafka libs but not kafka-server.
-  apt-get install -y kafka
-  exit 0
-fi
+function install_and_configure_kafka_server() {
+  # Find zookeeper list first, before attempting any installation.
+  local ZOOKEEPER_CLIENT_PORT=$(grep 'clientPort' /etc/zookeeper/conf/zoo.cfg \
```
bsidhom: One more small note about naming: local variables should be lower cased.

szewi: Ok, I applied to all local variables.

kafka/kafka.sh
```diff
@@ -37,25 +37,28 @@ function err() {
 
 function install_and_configure_kafka_server() {
   # Find zookeeper list first, before attempting any installation.
-  local ZOOKEEPER_CLIENT_PORT=$(grep 'clientPort' /etc/zookeeper/conf/zoo.cfg \
+  local zookeeper_client_port
```
szewi: I was't sure if it would be better to have list of declarations of local variables at the top of the function. WDYT @bsidhom ?

bsidhom: This isn't explicitly called out in the style guide, but I prefer having local variables declared as late as possible so we can see their use in context. This is in keeping with the [Java style guide](https://google.github.io/styleguide/javaguide.html#s4.8.2.2-variables-limited-scope).

bsidhom:
Looks good. Thanks for the changes!

bsidhom:
Just tested this with HA and it appears to work. Unfortunately, it doesn't work in non-HA mode due to a problem with the zookeeper service. I don't think this is a regression, so I'll go ahead and merge this as-is, but we'll probably need to fix the single-master zookeeper init action.

szewi:
This PR was part of work on the issue #200

PR #191, opened 2018-02-23
Adjust Zeppelin init action to conform the guidelines.

DariuszAniszewski:
As per https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/190 following topics are covered:
- consistent indentation
- moved constants to top of the file
- add err function and log installation problems
- add systemd config
- main function

I'm also wondering what to do with this commented-out block of code related to `mplot` and `rCharts` . Was it ever used? It could be changed to be run only if some appropriate metadata key is set, but currently it doesn't work when uncommented.

### DIFF
--- a/zeppelin/zeppelin.sh
+++ b/zeppelin/zeppelin.sh
@@ -97,23 +97,7 @@ function configure_zeppelin(){
     # TODO(pmkc): Add googlevis to Zeppelin Package recommendations
     apt-get install -y r-cran-googlevis
 
-## Uncomment here to compile and install 'mplot' and 'rCharts'.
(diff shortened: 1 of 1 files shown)


pmkc:
As to the R packages it used to work, but non-run code stagnates especially with at HEAD references like in R packages (and Zeppelin upgrades).

I would support replacing the code with a TODO / issue reference to fix SparkR in Zeppelin.

DariuszAniszewski:
@pmkc thanks for suggestion, I followed your idea and put a TODO.

DariuszAniszewski:
linking this to #200

Issue #200, opened 2018-02-27
Initialization scripts should be consistent with Shell Style Guide


szewi:
This task will be used for future PRs when checking for scripts consistency with Shell Style Guide. 

So far for each init script there was separate issue created. This issue will aggregate all releated PRs. 
Link to guide:

https://google.github.io/styleguide/shell.xml

karth295:
Closing this issue for now, as we have fixed at least the highly used initialization actions.

Issue #199, opened 2018-02-27
Apache Tez not working on HA mode


szewi:
The initialization action fails for apache tez in HA mode, with an multiple errors of:
> copyFromLocal: `/apps/tez/hadoop-shim-0.8.4.jar': File exists
> copyFromLocal: `/apps/tez/hadoop-shim-2.8-0.8.4.jar': File exists

Problem is connected with copying of tez jar files to ${tez_hdfs_path} by each of master nodes. 
`hadoop fs -copyFromLocal ${tez_jars}/* ${tez_hdfs_path}/
` when second master is trying to copy jar files they are already there.

Sometimes additionally following errors occur, but they are all related to this issue:
>copyFromLocal: File does not exist: /apps/tez/tez-yarn-timeline-history-with-acls.jar._COPYING_ (inode 16549) Holder DFSClient_NONMAPREDUCE_-1172794459_1
>copyFromLocal: File does not exist: /apps/tez/tez-yarn-timeline-history-with-fs-0.8.4.jar._COPYING_`

Steps to reproduce:
`gcloud dataproc clusters create tez-kamil --initialization-actions gs://dataproc-initialization-actions/tez/tez.sh --initialization-action-timeout 5m --num-workers 2 --num-masters 3 --worker-machine-type n1-standard-4 --master-machine-type n1-standard-4
`

Using `hadoop fs -copyFromLocal -f` that forces write operation seem to help, but doesn't solve the issue with HA mode.

karth295:
So the reason this happens is that `hadoop fs -copyFromLocal` writes a temp file, then renames to the final location. It does this so that if you fail writing a file halfway, you never write the final file (but just leave a temp file around). The problem here is that it always uses the same temp file. There's a race condition where 2 writers overwrite the same temp file, 1 writers succeeds in renaming it, and the 2nd writer complains that the file is not found.

One reasonable solution is to only run this code if `$HOSTNAME == dataproc-master` (aka -m-0), and put a comment explaining why that's necessary.

szewi:
Thx @karth295 I was also thinking about doing `hadoop fs -test` on  `${tez_hdfs_path}`, but this seem to be a simple solution. I tested your proposal and ended up with known issue about hive2.server  on multiple masters. I fixed both. I will submit a PR and you will see.

PR #201, opened 2018-02-28
Tez initialization script updated to follow https://google.github.io/…

szewi:
…styleguide/shell.xml.

### DIFF
--- a/tez/tez.sh
+++ b/tez/tez.sh
@@ -25,6 +25,7 @@ readonly TEZ_JARS='/usr/lib/tez'
 readonly TEZ_CONF_DIR='/etc/tez/conf'
 readonly HADOOP_CONF_DIR='/etc/hadoop/conf'
 readonly HIVE_CONF_DIR='/etc/hive/conf'
+readonly SPARK_CONF_DIR='/etc/spark/conf'
 
 function update_apt_get() {
   for ((i = 0; i < 10; i++)); do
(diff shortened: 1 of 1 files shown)


szewi:
This is releated with issue #200 .

Could be easy tested:
1. This initialization script is located in `gs://polidea-dataproc-utils/tez.sh` and could be used when creating cluster
2. It's possible to spin Tez job on cluster master node by running validation script:
`gcloud compute ssh MASTER_NODE_NAME -- -t "curl -s https://storage.googleapis.com/polidea-dataproc-utils/tez/validate.sh | bash"`

This PR is not resolving #199 so don't expect Tez working on HA mode.

tez/tez.sh
```diff
@@ -126,6 +134,23 @@ if [[ "${role}" == 'Master' ]]; then
...
-fi
+}
+
+function main() {
+  local role
+  role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+
+  # Let spark continue using the existing hive configuration, as it will
+  # not want to use hive on tez.
+  cp /etc/hive/conf/* /etc/spark/conf/
```
karth295: Nit: use HIVE_CONF_DIR and create a SPARK_CONF_DIR

szewi: Ok, here also this should be substituted with proper variables.

PR #202, opened 2018-03-01
refactor datalab init-action to follow shell script guidelines

DariuszAniszewski:
This PR is related to #200. 

Changed items:
- moved constants to top of the file
- add err function and check return values when applicable
- copy only files needed by `Dockerfile`, not all of them in directory

following issues are identified, but left as-is as changing them to strictly follow guidelines will degrade readability
- `set -u` flag can't be set as script uses standard environment variables that are not declared within the script
- Usages `sed` could probably be replaced by bash equivalent but it's more readable as is 
- Using symlinks instead of `cp` is not possible, as files must exists in the same directory as `Dockerfile`. Symlinks are not followed.
- there is no systemd configuration, but `docker run --restart=always` seems good enough
- `main` function doesn't seem to be needed here

### DIFF
--- a/datalab/datalab.sh
+++ b/datalab/datalab.sh
@@ -24,9 +24,12 @@ readonly SPARK_PACKAGES="$(/usr/share/google/get_metadata_value attributes/spark
 readonly SPARK_CONF='/etc/spark/conf/spark-defaults.conf'
 readonly DATALAB_DIR="${HOME}/datalab"
 readonly PYTHONPATH="/env/python:$(find /usr/lib/spark/python/lib -name '*.zip' | paste -sd:)"
+readonly DOCKER_IMAGE="$(/usr/share/google/get_metadata_value attributes/docker-image || \
+  echo 'gcr.io/cloud-datalab/datalab:local')"
 
-DOCKER_IMAGE="$(/usr/share/google/get_metadata_value attributes/docker-image || true)"
-
+# Expose every possible spark configuration to the container.
+readonly VOLUMES="$(echo /etc/{hadoop*,hive*,*spark*} /usr/lib/hadoop/lib/{gcs,bigquery}*)"
+readonly VOLUME_FLAGS="$(echo "${VOLUMES}" | sed 's/\S*/-v &:&/g')"
 
 function update_apt_get() {
   for ((i = 0; i < 10; i++)); do
@@ -43,24 +46,13 @@ function err() {
   return 1
 }
 
-if [[ "${ROLE}" == 'Master' ]]; then
+function configure_master(){
   update_apt_get || err 'Failed to update apt-get'
   mkdir -p "${DATALAB_DIR}"
 
-  apt-get install -y -q docker.io
-  if [ $? != 0 ]; then
-    err 'Failed to install Docker'
-  fi
+  apt-get install -y -q docker.io || err 'Failed to install Docker'
 
-  if [[ -z "${DOCKER_IMAGE}" ]]; then
-    DOCKER_IMAGE="gcr.io/cloud-datalab/datalab:local"
-  fi
-  readonly DOCKER_IMAGE
-
-  gcloud docker -- pull ${DOCKER_IMAGE}
-  if [ $? != 0 ]; then
-    err 'Failed to pull ${DOCKER_IMAGE}'
-  fi
+  gcloud docker -- pull ${DOCKER_IMAGE} || err "Failed to pull ${DOCKER_IMAGE}"
 
   # For some reason Spark has issues resolving the user's directory inside of
   # Datalab.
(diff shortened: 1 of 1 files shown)


datalab/datalab.sh
```diff
@@ -16,7 +16,17 @@
...
+set -exo pipefail
+
+readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+readonly PROJECT="$(/usr/share/google/get_metadata_value ../project/project-id)"
+readonly SPARK_PACKAGES="$(/usr/share/google/get_metadata_value attributes/spark-packages || true)"
+readonly SPARK_CONF='/etc/spark/conf/spark-defaults.conf'
+readonly DATALAB_DIR="${HOME}/datalab"
+readonly PYTHONPATH="/env/python:$(find /usr/lib/spark/python/lib -name '*.zip' | paste -sd:)"
+
+DOCKER_IMAGE="$(/usr/share/google/get_metadata_value attributes/docker-image || true)"
```
karth295: You can make this also `readonly` if you make this something like `DOCKER_IMAGE=$(...get_metadata_value... || echo 'gcr.io/cloud-datalab/datalab:local')`. And you can remove the logic later in this script.

datalab/datalab.sh
```diff
@@ -28,53 +38,65 @@ function update_apt_get() {
...
+  return 1
+}
 
 if [[ "${ROLE}" == 'Master' ]]; then
-  update_apt_get
+  update_apt_get || err 'Failed to update apt-get'
+  mkdir -p "${DATALAB_DIR}"
+
   apt-get install -y -q docker.io
+  if [ $? != 0 ]; then
```
karth295: Your convention seems to be `<command> || err 'Reason'`, which is a little bit shorter

datalab/datalab.sh
```diff
@@ -28,53 +38,65 @@ function update_apt_get() {
...
+    err 'Failed to install Docker'
+  fi
+
   if [[ -z "${DOCKER_IMAGE}" ]]; then
     DOCKER_IMAGE="gcr.io/cloud-datalab/datalab:local"
   fi
+  readonly DOCKER_IMAGE
+
   gcloud docker -- pull ${DOCKER_IMAGE}
+  if [ $? != 0 ]; then
```
karth295: Same here

datalab/datalab.sh
```diff
@@ -28,53 +38,65 @@ function update_apt_get() {
...
-  SPARK_CONF='/etc/spark/conf/spark-defaults.conf'
   if ! grep -q '^spark\.sql\.warehouse\.dir=' "${SPARK_CONF}"; then
     echo 'spark.sql.warehouse.dir=/root/spark-warehouse' >> "${SPARK_CONF}"
   fi
 
-  DATALAB_DIR="${HOME}/datalab"
-  mkdir -p "${DATALAB_DIR}"
-
   # Expose every possible spark configuration to the container.
   VOLUMES=$(echo /etc/{hadoop*,hive*,*spark*} /usr/lib/hadoop/lib/{gcs,bigquery}*)
```
karth295: Can these four lines be pulled out into a `readonly VOLUME` and `readonly VOLUME_FLAGS` at the top? I don't think the `echo` line is necessary because you are running with `set -x`

datalab/datalab.sh
```diff
@@ -16,7 +16,17 @@
 # This init script installs Google Cloud Datalab on the master node of a
 # Dataproc cluster.
 
-set -e -x
+set -exo pipefail
```
karth295: Other scripts seem to be using `set -euxo pipefail`

datalab/datalab.sh
```diff
@@ -28,53 +38,65 @@ function update_apt_get() {
...
-ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
-PROJECT=$(/usr/share/google/get_metadata_value ../project/project-id)
-DOCKER_IMAGE=$(/usr/share/google/get_metadata_value attributes/docker-image || true)
-SPARK_PACKAGES=$(/usr/share/google/get_metadata_value attributes/spark-packages || true)
+function err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
+  return 1
+}
 
 if [[ "${ROLE}" == 'Master' ]]; then
```
karth295: I noticed some of the other scripts are moving to using a `main` method at the bottom, and in `main` saying `if role = master run configure_master`. That might be nice here too.

E.g. https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/191/files

datalab/datalab.sh
DariuszAniszewski: Good point

datalab/datalab.sh
DariuszAniszewski: `-u` is not possible - this script uses environment variables from the system that are not declared in this script

DariuszAniszewski:
@karth295 thanks for review. I pushed my changes, PTAL.

PR #207, opened 2018-03-05
Init actions that update GCS and BQ connectors.

medb:
New GCS (1.7.0) and BQ (0.11.0) connectors have been released, but they will not be included in Dataproc until 1.3 image release.

This init actions allows to use one or both of the new connectors with existing Dataproc images.

### DIFF
--- a/connectors/README.md
+++ b/connectors/README.md
@@ -17,14 +17,16 @@ gcloud dataproc clusters create <CLUSTER_NAME> \
 This script downloads specified versions of GCS connector and matching version of BigQuery connector (if BigQuery connector
 was installed before) from the same [release](https://github.com/GoogleCloudPlatform/bigdata-interop/releases).
 
-To specify connector version, find the needed released GCS connector version from the
+To specify connector version, find the needed released GCS connector version on the
 [connectors releases page](https://github.com/GoogleCloudPlatform/bigdata-interop/releases),
-and set it as `gcs-connector-version` metadata key value.
+and set it as the `gcs-connector-version` metadata key value.
 
 If cluster already has BigQuery connector installed then it will be updated together with GCS connector, otherwise BigQuery
 connector will not be updated.
 
-For example, if GCS connector 1.7.0 version will be specified then BigQuery connector will be updated to 0.11.0 version, because they are from the same [release](https://github.com/GoogleCloudPlatform/bigdata-interop/releases).
+For example, if GCS connector 1.7.0 version is specified and BigQuery connector is installed on a Dataproc cluster before
+execution of the init action then BigQuery connector will be updated to 0.11.0 version, because they are from the same
+[release](https://github.com/GoogleCloudPlatform/bigdata-interop/releases).
 
 You can find more information about using initialization actions with Dataproc in the
 [Dataproc documentation](https://cloud.google.com/dataproc/init-actions).


connectors/connectors.sh
```diff
@@ -0,0 +1,55 @@
...
+   *) echo "Error: invalid connector name - '$name'"; return 1;;
+  esac
+  if [[ $version ]] && [[ ! "${valid_versions[@]}" =~ "$version" ]]; then
+    local IFS=, # print versions concatenated by comma
+    echo "ERROR: $name-connector version should be one of [${valid_versions[*]}], but was '$version'"
+    return 1
+  fi
+}
+
+upload_connector() {
```
karth295: Nit: `download_connector`

connectors/README.md
```diff
@@ -0,0 +1,30 @@
...
+gcloud dataproc clusters create <CLUSTER_NAME> \
+    --initialization-actions gs://dataproc-initialization-actions/connectors/connectors.sh \
+    --metadata 'gcs-connector-version=1.7.0'
+```
+
+This script downloads specified versions of BigQuery and/or GCS connectors.
+To specify connector version, find the needed released connector versions from the
+[BigQuery and GCS connectors releases page](https://github.com/GoogleCloudPlatform/bigdata-interop/releases),
+and set one or both metadata keys `gcs-connector-version` and/or `bigquery-connector-version`.
+If version is set only for one connector (BigQuery or GCS) then only this connector will be installed, but both old BigQuery and GCS connectors will be removed.
```
karth295: This seems more complicated as a user -- if I want to update the bigquery connector, I also need to specify a GCS connector version to keep writing to GCS?

Also, doesn't the BQ connector depend on the GCS connector?

I would vote to either mandate both are specified, or use the version map you have.

connectors/connectors.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+VM_CONNECTORS_DIR=/usr/lib/hadoop/lib
+
+declare -A GCS_TO_BQ_VERSIONS_MAP
+GCS_TO_BQ_VERSIONS_MAP=(
+  ["1.7.0"]="0.11.0")
```
karth295: Is this just to avoid dependency conflicts? Otherwise it seems like "ask for forgiveness rather than permission" is better -- just try to download the version the user specified. If it does not exist, they will get an error anyway. And that way you don't need to update this map for every release.

connectors/connectors.sh
medb: Done.

connectors/connectors.sh
medb: Yes, both connectors have the same classes and they could conflict with each other if connectors are from different releases.

connectors/README.md
medb: Agree, it's not very user friendly, but it has explicit behavior and will fail-fast, if user didn't specify needed connector version during cluster creation.

In future Dataproc releases BQ connector could be (will be?) excluded from the image, so it will be weird to install BQ connector in init action if only GCS connector version is specified or require to specify both of them. But, because GCS connector will be always in Dataproc image and more widely used, it could make sense to update GCS connector even if only BQ connector version specified, but this looks illogical.

BQ connector doesn't need GCS connector to function, it has all GCS-related classes that it needs in it.

medb: @karth295 I have modified init action to update BigQuery connector to version that matches specified GCS connector version if BigQuery connector is installed, and do nothing if BigQuery connector not installed. PTAL.

PR #210, opened 2018-03-06
Revert "Adjust Zeppelin init action to conform the guidelines. (#191)"

karth295:
This reverts commit 0e82573e84de0d6b10ee2eefbba76353ff99441c.

### DIFF
--- a/zeppelin/zeppelin.sh
+++ b/zeppelin/zeppelin.sh
@@ -17,12 +17,7 @@
 # Dataproc cluster. Zeppelin is also configured based on the size of your
 # cluster and the versions of Spark/Hadoop which are installed.
 
-set -euxo pipefail
-
-
-readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
-readonly INTERPRETER_FILE='/etc/zeppelin/conf/interpreter.json'
-readonly INIT_SCRIPT='/usr/lib/systemd/system/zeppelin-notebook.service'
+set -x -e
 
 function update_apt_get() {
   for ((i = 0; i < 10; i++)); do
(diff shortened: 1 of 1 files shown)


karth295:
CC @DariuszAniszewski 

Reverting because this breaks in dataproc 1.0, because `zeppelin-daemon.sh upstart` was not available in Zeppelin 0.5. You can try it on a 1.0 cluster by passing `--image-version=1.0`. (I'll admit I only test on the latest image version, but our internal integration tests run on all versions).

Also, this PR created two `systemd` configs -- one that comes with `apt-get` and one that was explicitly created in the PR. This is a little bit confusing. You can instead just change the line that says `service start zeppelin` to something like `systemctl enable zeppelin` `systemctl start zeppelin`.

DariuszAniszewski:
@karth295 OK, thanks for info. I'll check it on Dataproc 1.0 and 1.1 as well.

As for the additional config. Well, yes - I'm explicitly creating new config file, but the config created by `apt-get` isn't configured for auto-restart. I was able to find following `.service` files related to Zeppelin:
```
# find / -name "*.service" | grep zeppelin
/run/systemd/generator.late/zeppelin.service
/run/systemd/generator.late/runlevel5.target.wants/zeppelin.service
/run/systemd/generator.late/runlevel4.target.wants/zeppelin.service
/run/systemd/generator.late/runlevel3.target.wants/zeppelin.service
/run/systemd/generator.late/runlevel2.target.wants/zeppelin.service
/sys/fs/cgroup/systemd/system.slice/zeppelin.service
```

I assume that the first one `/run/systemd/generator.late/zeppelin.service` is the one that I should be considering. It has 
```
...
[Service]
Type=forking
Restart=no
...
```
so if zeppelin stops (for any reason) it won't be started. 

`systemctl enable zeppelin` and `systemctl start zeppelin` works well, but in the end it's equivalent to `service zeppelin start` - a session of Zeppelin is started, but not configured to restart on failure.

Maybe an option is to `sed` the config here: `/run/systemd/generator.late/zeppelin.service` to add auto-restart. WDYT?

PR #215, opened 2018-03-07
refactored Zeppelin init action

DariuszAniszewski:
This PR is follow-up for #191 that was reverted in #209. As per discussion in #209 I removed additional systemd config file in favor of the one installed by apt-get.

I'm however still not sure if we should simply abandon it. The main benefit of custom `zeppelin.service` is that it contained `Restart=always` clause that is not there in the pre-installed configs. What's more it seems to be hard to change (i.e. via sed) because `Restart=no` seems to be hardcoded in sysv-generator as can be seen here: 
https://github.com/systemd/systemd/blob/master/src/sysv-generator/sysv-generator.c#L179
(it's pointig to master, but this line was the same for last 4 years ;)

I didn't find a way to change this configuration and if we want to have auto-restarting Zeppelin I'd suggest to bring back previous changes and modify them for Dataproc 1.0 on which `zeppelin-daemon.sh upstart` is not available. WDYT?

(part of #200)

### DIFF
--- a/zeppelin/zeppelin.sh
+++ b/zeppelin/zeppelin.sh
@@ -103,27 +103,10 @@ function configure_zeppelin(){
 
 function launch_zeppelin(){
   # Start Zeppelin as systemd job
-
-  cat << EOF > ${INIT_SCRIPT}
-[Unit]
-Description=Zeppelin Notebook
-
-[Service]
-Type=simple
-ExecStart=/usr/lib/zeppelin/bin/zeppelin-daemon.sh upstart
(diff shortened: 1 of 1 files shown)


DariuszAniszewski:
Long story short: I can't modify those configs in persistent way. I can append `Restart=always` or even use `sed` to replace it, but it will be gone when `.service` files will be re-generated and change is lost.

Yes, notebook will start after reboot. Question is - what do if notebook process crashes? In current configuration it won't be restarted. This was main benefit of `Restat=always` in custom service script.

If we're OK with no automatic restart upon Zeppelin crash, no further changes are needed and current setup works.

Issue #220, opened 2018-03-12
Presto not working on Single-node and HA


DariuszAniszewski:
Init action for Presto is not working on single-node and HA configurations of Dataproc.

Single node failure is because by default, master doesn't consider itself as worker, thus number of workers is 0
> node-scheduler.include-coordinator=false

HA fails because of broken reference to master(s). Prefix `m` is used, which isn't good name for HA:
>	MASTER_NODE_NAME=$(hostname | sed 's/\(.*\)-s\?w-[a-z0-9]*.*/\1/g')
>	PRESTO_MASTER_FQDN=${MASTER_NODE_NAME}-m.${DNSNAME}

PR #221, opened 2018-03-12
Fix Single-Node and HA configurations for Presto

DariuszAniszewski:
This PR fixes issue #220.

### DIFF
--- a/presto/presto.sh
+++ b/presto/presto.sh
@@ -16,6 +16,7 @@ set -x -e
 
 # Variables for running this script
 ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+HOSTNAME=$(hostname -s)
 PRESTO_MASTER_FQDN=$(/usr/share/google/get_metadata_value attributes/dataproc-master)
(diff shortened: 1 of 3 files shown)


presto/presto.sh
```diff
@@ -96,10 +97,15 @@ EOF
 
 if [[ "${ROLE}" == 'Master' ]]; then
 	# Configure master properties
-	PRESTO_MASTER_FQDN=${FQDN}
+	if [[ $WORKER_COUNT == 0 ]]; then
```
karth295: Nit: The indentation looks off here.

DariuszAniszewski:
@karth295 thanks for insights. I was naively thinking than in case of `m-0` failure other master node will take its place and there will be some automatic routing from broken `m-0` to working `m-N`

You are right, other masters were considered as workers and in HA number of workers was 4. I followed you idea of running coordinator only on `m-0` and workers on actual workers, PTAL.

I also modified Readme as HA doesn't really makes sense - it's basically equivalent of Regular setup with other master nodes idle.

PR #223, opened 2018-03-13
Adjust Jupyter script to Shell style guidelines:

DariuszAniszewski:
- consistent indentation
- unified shebang
- inline `;then` with `if`s
- unified how `ROLE` is obtained
- added top-level comments for files
- removed curl in favor of get_metadata_value
- rearranged constants
- consistent quoting
- use built-in string replacement instead of external tr

part of #200


DariuszAniszewski:
one comment on renaming files - I wanted to be aligned with [this requirement](https://google.github.io/styleguide/shell.xml?showone=Source_Filenames#Source_Filenames) but git doesn't see this as move but rather as new file (even when I used `git mv`) and by this we loose some (possibly important) history of file changes. 
In such case I'd consider un-renaming files and break the requirement on purpose, just in favor of keeping history. WDYT?

karth295:
I'm a little nervous to move setup_jupyter_kernel and bootstrap_jupyter_ext -- user often copy a snapshot of an init action to their own GCS bucket to avoid running into problems if we make a breaking change. So if a user copied `jupyter.sh` without changing the `git clone` url in the script, their init action will break.

But if we are going to rename the files, I'd prefer to combine all of them into one `jupyter.sh`, similar to the one in #180 . After installing conda, it should be virtually the same, substituting conda for pip. But that's orthogonal to this PR.

For now, I'd vote to avoid renaming the files and breaking users.

To answer your question, you could avoid this problem by pushing a commit that just renames the files (and has the appropriate changes to `jupyter.sh`). Then, another commit that actually touches the files. That's a good rule of thumb with git in general -- whitespace changes and renaming files should generally be their own commit. We can then use github's "Rename and merge" to preserve the commits.

DariuszAniszewski:
thanks @karth295, I didn't thought about the copying snapshot use-case - this is even better reason not to rename the files. I reverted this change already - PTAL.

I was thinking about two commits as walk-around for this rename problem but to be honest I was afraid that you may squash-rebase this branch and in the end it will end up with one change.

jupyter/launch-jupyter-interface.sh
```diff
@@ -1,56 +1,61 @@
...
-    echo "If the appropriate environment variables are not set and the appropriate command"
-    echo "line arguments are not given, then the usage message will be displayed and the "
-    echo "script will exit."
-    echo ""
-    echo "usage: $0 [-h] [-c=cluster-name] [-z=zone]"
-    echo "    -h                 display help"
-    echo "    -z=zone            specify cloud zone for cluster"
-    echo "    -c=cluster-name    specify unique dataproc cluster name to launch"
-    exit 1
+  echo "Creates an SSH tunnel and socks proxy and launches Chrome, using the environment "
```
karth295: Nit: this is more readable as:

```
cat << EOF
...
EOF
exit 1
```

jupyter/launch-jupyter-interface.sh
```diff
@@ -3,7 +3,7 @@
 # This script sets SSH tunnel and opens Chrome Browser with SOCKS proxy
 # configured for direct access to the Jupyter cluster.
 
-set -euxo pipefail
+set -euo pipefail
```
karth295: Any reason to remove `-x`?

DariuszAniszewski: One actually - readability. This script is used locally to connect to Jupyter via SOCKS proxy and with `-x` all commands were additionally printed which doesn't really makes sense when you use it:
```
$ ./launch-jupyter-interface.sh -h
+ DIR=.
+ [[ ! -d . ]]
+ readonly DIR
+ source ./../util/utils.sh
+ for i in '"$@"'
+ case $i in
+ usage
+ cat
Creates an SSH tunnel and socks proxy and launches Chrome, using the environment
variable DATAPROC_CLUSTER_NAME for the unique cluster name. The cluster metadata
must contain a value for the key 'JUPYTER_PORT'.

If the appropriate environment variables are not set and the appropriate command"
line arguments are not given, then the usage message will be displayed and the "
script will exit."

usage: ./launch-jupyter-interface.sh [-h] [-c=cluster-name] [-z=zone]"
  -h                 display help"
  -z=zone            specify cloud zone for cluster"
  -c=cluster-name    specify unique dataproc cluster name to launch"
+ exit 1
```
After removal, output is clean:
```
$ ./launch-jupyter-interface.sh -h
Creates an SSH tunnel and socks proxy and launches Chrome, using the environment
variable DATAPROC_CLUSTER_NAME for the unique cluster name. The cluster metadata
must contain a value for the key 'JUPYTER_PORT'.

If the appropriate environment variables are not set and the appropriate command"
line arguments are not given, then the usage message will be displayed and the "
script will exit."

usage: ./launch-jupyter-interface.sh [-h] [-c=cluster-name] [-z=zone]"
  -h                 display help"
  -z=zone            specify cloud zone for cluster"
  -c=cluster-name    specify unique dataproc cluster name to launch"
```

PR #224, opened 2018-03-13
Implement RStudio Server init-action

gliptak:


gliptak:
@medb @karth295 @pmkc Please review. Thanks

rstudio/README.MD
```diff
@@ -0,0 +1,21 @@
+# RStudio Server
+
+This initialization action installs the latest version of [RStudio Server](https://www.rstudio.com/products/rstudio/#Server) on a master node within a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster.
+
+## Using this initialization action
+You can use this initialization action to create a new Dataproc cluster with RStudio Server installed by:
+
+1. Uploading a copy of this initialization action (`rstudio.sh`) to [Google Cloud Storage](https://cloud.google.com/storage).
```
karth295: Heh I know a lot of init action README's have instructions to copy a script to your own GCS bucket. I'd rather have all of them avoid that and just have a `clusters create` command that uses our mirror of this repo on GCS:

```
gcloud dataproc clusters create <CLUSTER_NAME> \
  --initialization-actions gs://dataproc-initialization-actions/rstudio/rstudio.sh
```

rstudio/rstudio.sh
```diff
@@ -0,0 +1,49 @@
...
+  apt-get install -y software-properties-common
+  apt-key adv --keyserver keys.gnupg.net --recv-key 'E19F5F87128899B192B1A2C2AD5F960A256A04AF'
+  add-apt-repository 'deb http://cran.r-project.org/bin/linux/debian jessie-cran34/'
+  update_apt_get
+  apt install -y r-base r-base-dev gdebi-core
+  
+  cd /tmp; wget https://download2.rstudio.org/rstudio-server-${RSTUDIO_VERSION}-amd64.deb
+  gdebi -n /tmp/rstudio-server-${RSTUDIO_VERSION}-amd64.deb
+  
+  groupadd --gid 5501 rstudio
```
karth295: Is there a particular reason for hardcoding a uid/gid? You can just leave that out otherwise.

rstudio/rstudio.sh
```diff
@@ -0,0 +1,49 @@
...
+  useradd --create-home --gid 5501 --uid 5501 rstudio
```
karth295: This user doesn't need a home directory

rstudio/rstudio.sh
```diff
@@ -0,0 +1,49 @@
...
+  echo "rstudio:rstudio" | chpasswd
```
karth295: This is unnecessary -- users already have `sudo` access to the VMs, so they can just run `sudo su rstudio` if really necessary. Regardless, hardcoding passwords in scripts scares me a little.

rstudio/rstudio.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Only run on the master node
+ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+RSTUDIO_VERSION=1.1.442
+
+if [[ "${ROLE}" == 'Master' ]]; then
+  # Install RStudio Server
+  update_apt_get
+  apt-get install -y software-properties-common
+  apt-key adv --keyserver keys.gnupg.net --recv-key 'E19F5F87128899B192B1A2C2AD5F960A256A04AF'
```
karth295: Just doing my due diligence -- is there a source saying to use this repo/key? I googled around and found slightly different keys/repo urls.

rstudio/README.MD
gliptak: I will update README after the script updates are done.

rstudio/rstudio.sh
gliptak: Removed hardcoded uid/gid

rstudio/rstudio.sh
gliptak: The user's home director is used for script strorage and R package installs, so I recommend that we create a home directory.

rstudio/rstudio.sh
gliptak: I'm not disagreeing. The reason to create this user so this non-root login is used on the RStudio Server webtop. I do not know if/how to disable that login (which seem to be the approach used for other webtops installed).

rstudio/rstudio.sh
gliptak: Yes, I have seen a number of various keys too ... These are the instructions i used:

https://linode.com/docs/development/r/how-to-install-r-on-ubuntu-and-debian/

gliptak:
@karth295 Would you like to see other updates (with the exception of the README one)? Thanks

karth295:
@gliptak sorry for the late reply -- this looks good to me! If you update the README I'll go ahead and merge it.

gliptak:
@karth295 Thank you. I will update README in this PR

rstudio/README.MD
gliptak: Updated

karth295:
^I added a note about RStudio's AGPL license to the README

karth295:
@gliptak sorry to add more work for you, but can you add details in the README about how to use Spark + RStudio on Dataproc. I see two ways to run it with radically different instructions:

* https://spark.apache.org/docs/latest/sparkr.html#starting-up-from-rstudio
* http://spark.rstudio.com/

One issue with both is that they say to use `master=local[*]`, which you wouldn't want to do on Dataproc except on single node clusters. The README should be clear that you want to just use the pre-configured options (aka just `spark_connect()`).

Also, it would be nice if the init action pre-installed / configured the needed libraries. Ideally you would be able to run the init action and immediately be able to run:

```
sc <- spark_connect()
```

@pmkc knows more about R and might have more insights here.

gliptak:
@karth295 At this time, I won't be able to test proposed updates to README. Maybe this can be committed as WIP and @pmkc add clarifications? Thanks

PR #229, opened 2018-03-26
Set option to allow any origin option on jupyter

DanSedov:
This helps with #230, accessing jupyter from webshell issues.


karth295:
Rejecting this since we cannot use Jupyter with Cloud Shell Preview (see discussion in #230 ). Removing security measures unnecessarily makes me nervous :)

PR #231, opened 2018-03-27
Fixing Tez init action issues with copying tez jars to hdfs in HA mode.

szewi:
This fixes issue described in #199 .

### DIFF
--- a/tez/tez.sh
+++ b/tez/tez.sh
@@ -51,10 +51,13 @@ function configure_master_node() {
   apt-get install tez hadoop-yarn-timelineserver nodejs -y \
     || err 'Failed to install required packages.'
 
-  # Stage Tez
-  hadoop fs -mkdir -p ${TEZ_HDFS_PATH}
-  hadoop fs -copyFromLocal ${TEZ_JARS}/* ${TEZ_HDFS_PATH}/ \
-    || err 'Unable to copy tez jars to hdfs destination.'
+  # Copy to hdfs from one master only to avoid race
+  if [[ "${HOSTNAME}" == "${master_hostname}" ]]; then
+    # Stage Tez
+    hadoop fs -mkdir -p ${TEZ_HDFS_PATH}
+    hadoop fs -copyFromLocal ${TEZ_JARS}/* ${TEZ_HDFS_PATH}/ \
+      || err 'Unable to copy tez jars to hdfs destination.'
+  fi
 
   # Update the hadoop-env.sh
   {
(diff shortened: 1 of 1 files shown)


tez/tez.sh
```diff
@@ -132,14 +135,18 @@ function configure_master_node() {
   systemctl restart hadoop-yarn-timelineserver
   systemctl status hadoop-yarn-timelineserver  # Ensure it started successfully
 
-  # Restart hive server2
-  systemctl restart hive-server2
-  systemctl status hive-server2  # Ensure it started successfully
+  if [[ "${HOSTNAME}" == "${master_hostname}" ]]; then
```
karth295: In case all masters run a hive-server2 in the future, I'd rather restart hive-server2 if it's installed/enabled in systemd, rather than only restarting on -m-0..

szewi: Sure. It may be possible that all masters will run hive-server2 in future. I could use `systemctl show --property=LoadState` to check if it's installed or not. The result of that command could be "loaded" or "not-found".

tez/tez.sh
```diff
@@ -132,14 +135,24 @@ function configure_master_node() {
   systemctl restart hadoop-yarn-timelineserver
   systemctl status hadoop-yarn-timelineserver  # Ensure it started successfully
 
-  # Restart hive server2
-  systemctl restart hive-server2
-  systemctl status hive-server2  # Ensure it started successfully
+  # Check hive-server2 status
+  local hive_server2_status
+  hive_server2_status=$(systemctl show --property=LoadState hive-server2.service \
```
karth295: This will work for hive-server2 because it's either installed or not installed. I'm worried about creating a precedent that won't work for namenodes -- on `-m-2`, the namenode package is installed but not enabled.

I'd prefer to use `systemctl list-unit-files --state=enabled | grep <whatever>.service` to only worry about services we plan to run.

szewi: Yes, I checked that and your proposal will work only for native services, so for instance `hive-server2.sevice` will not appear on the list when checking using `systemctl list-unit-files --state=enabled` command. It will be on the list when we search using `systemctl list-unit-files --type=service` however by its state in the output we can't determine its right status. 
I think that, if we could consider that services we plan to run must be enabled (to allow them automatically start after the reboot ), then the nicest way to check whether service is enabled is by running: 
`systemctl is-enabled --quiet hive-server2.service`

karth295: Awesome, didn't know about that one!

PR #232, opened 2018-03-28
refactored presto init-action to conform with shell-style guidelines

DariuszAniszewski:
(part of #200)

### DIFF
--- a/presto/presto.sh
+++ b/presto/presto.sh
@@ -129,14 +129,14 @@ EOF
 }
 
 function configure_master(){
-	# Configure master properties
-	if [[ ${WORKER_COUNT} == 0 ]]; then
-	  # master on single-node is also worker
-	  include_coordinator='true'
-	else
-	  include_coordinator='false'
-	fi
-	cat > presto-server-${PRESTO_VERSION}/etc/config.properties <<EOF
+  # Configure master properties
+  if [[ ${WORKER_COUNT} == 0 ]]; then
+    # master on single-node is also worker
+    include_coordinator='true'
+  else
+    include_coordinator='false'
+  fi
+  cat > presto-server-${PRESTO_VERSION}/etc/config.properties <<EOF
 coordinator=true
 node-scheduler.include-coordinator=${include_coordinator}
 http-server.http.port=${HTTP_PORT}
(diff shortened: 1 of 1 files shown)


presto/presto.sh
```diff
@@ -95,17 +126,17 @@ cat > presto-server-${PRESTO_VERSION}/etc/jvm.config <<EOF
...
-if [[ "${HOSTNAME}" == "${PRESTO_MASTER_FQDN}" ]]; then
-  # Configure master properties
-  if [[ $WORKER_COUNT == 0 ]]; then
-    # master on single-node is also worker
-    include_coordinator='true'
-  else
-    include_coordinator='false'
-  fi
-  cat > presto-server-${PRESTO_VERSION}/etc/config.properties <<EOF
+function configure_master(){
```
karth295: Indentation is a little off here.

presto/presto.sh
```diff
@@ -12,75 +12,106 @@
...
+    # executorMemory * 0.10, with minimum of 384
+    local min_executor_overhead=384
+    spark_executor_overhead_mb=$(( ${spark_executor_mb} / 10 ))
+    spark_executor_overhead_mb=$(( ${spark_executor_overhead_mb}>${min_executor_overhead}?${spark_executor_overhead_mb}:${min_executor_overhead} ))
+  fi
+  local spark_executor_count;
+  spark_executor_count=$(( $(nproc) / ${spark_executor_cores} ))
+
+  # Add up overhead and allocated executor MB for container size.
+  local spark_container_mb;
```
karth295: Can all of these `local foo; foo = 'bar'` be declared and assigned in one line? (Or is that against the style guide?)

presto/presto.sh
```diff
@@ -12,75 +12,106 @@
...
+}
+
+function get_presto(){
+  # Download and unpack Presto server
+  wget https://repo1.maven.org/maven2/com/facebook/presto/presto-server/${PRESTO_VERSION}/presto-server-${PRESTO_VERSION}.tar.gz
+  tar -zxvf presto-server-${PRESTO_VERSION}.tar.gz
+  mkdir -p /var/presto/data
+}
+
+function calculate_memory(){
```
karth295: Ah, sorry to make you refactor such an ugly function. There are easier ways to calculate memory on a node that don't involve spark. For example, just `bdconfig get_property_value --configuration_file /etc/hadoop/conf/yarn-site.xml --name yarn.nodemanager.resource.memory-mb` would tell you how much memory is configured on a worker node manager. This number is 80% of the node's total memory (in MB).

I think you can use that number to calculate the 60%/40% thing below. Presto's instructions for configuring memory are here: https://prestodb.io/presto-admin/docs/current/installation/presto-configuration.html.

If you want to do this in a different PR, feel free to push back.

presto/presto.sh
DariuszAniszewski: indeed, I missed this. Fixed

presto/presto.sh
DariuszAniszewski: Syntactically they can, but according to [shell style guide](https://google.github.io/styleguide/shell.xml?showone=Use_Local_Variables#Use_Local_Variables) they shouldn't:
> Declaration and assignment should be on different lines.

presto/presto.sh
DariuszAniszewski: I'll be happy to do it, but I suggest another PR as this isn't just a style change

DariuszAniszewski:
@karth295 Thanks for review, I pushed small change and responded to your comments, PTAL.

PR #233, opened 2018-03-28
Fix HA mode in cloud-sql-proxy. Use hive-metastore commands only on h…

### LINKED ISSUE #222: cloud-sql-proxy doesn't initialize in HA-mode
[cloud-sql-proxy.sh](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/cloud-sql-proxy/cloud-sql-proxy.sh) only initialize on m-0 when running in HA.

Steps to reproduce:
1. Create cloud SQL database:
`gcloud sql instances create my-db --tier db-n1-standard-1 --activation-policy=ALWAYS`

2. Use this database when creating HA cluster:

``` 
gcloud dataproc clusters create sql-ha \
--initialization-actions gs://dataproc-initialization-actions/cloud-sql-proxy/cloud-sql-proxy.sh \
--initialization-action-timeout 5m \
--num-workers 2 \
--num-masters 3 \
--worker-machine-type n1-standard-4 \
--master-machine-type n1-standard-4 \
--scopes sql-admin \ 
--properties hive:hive.metastore.warehouse.dir=${my_warehouse_bucket} \
--metadata "hive-metastore-instance=my-project:my-zone:my-db"
```

As a result in dataproc-initialization-script-0_output from m-1 and m2 we can see at the end:
``` 
+ [[ Master == \M\a\s\t\e\r ]]
+ ((  ENABLE_CLOUD_SQL_METASTORE  ))
+ systemctl stop hive-metastore
 Failed to stop hive-metastore.service: Unit hive-metastore.service not loaded. ```

### PR DESCRIPTION
**Pull request by szewi**
…osts where hive-metastore service is installed.

This fixes #222

### DIFF
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -161,7 +161,7 @@ EOF
       || err 'Failed to set mysql schema.'
   fi
 
-  if [[ "${hive_metastore_status}" == 'loaded' ]]; then
+  if ( systemctl is-enabled --quiet hive-metastore ); then
     # Start metastore back up.
     systemctl start hive-metastore \
       || err 'Unable to start hive-metastore service'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by szewi** [comment]
@karth295 Could you take a look? This should allow to handle situation when in future hive-metastore will be installed on all masters instead of single one. 
As a result if job is submitted to such ha cluster, it will be run successfully if will be executed on master that have yarn RM running, otherwise it will fail over to another RM on another master. 
I tested it with 3 masters and in my example it was executed on m-2 after two fail overs.

**Comment by karth295** [comment]
Awesome, thanks!

File: cloud-sql-proxy/cloud-sql-proxy.sh [hunk 1]
```diff
@@ -213,7 +217,15 @@ function main() {
   if [[ "${role}" == 'Master' ]]; then
     # Disable Hive Metastore and MySql Server.
     if (( ENABLE_CLOUD_SQL_METASTORE )); then
-      systemctl stop hive-metastore
+      local hive_metastore_status
+      hive_metastore_status=$(systemctl show --property=LoadState hive-metastore.service \
```
karth295: Same comment here -- let's set a precedent for checking whether a service is enabled, rather than whether it's installed.

PR #243, opened 2018-04-06
Increasing the number of open files in Presto.

adamderusha:
The default limit of open files in Presto appears to be 4096. However, in newer versions of Presto, this limit is too low, resulting in "java.io.IOException: Too many open files." This prevents new connections from being opened and manifests itself to the user as a failure to list a file.

This change raises the default limit so that this doesn't happen.


adamderusha:
I don't have any concrete reason other than Presto seems to need it but the others did not. My only worry would be something I don't know about that would break existing clusters if defaulted to a high number regardless of Presto installation.

medb:
Open files limit will be set to 65535 for all Systemd services in Datparoc starting from these image versions: 1.0.88, 1.1.79, 1.2.43, 1.3.3

PR #253, opened 2018-04-24
[Cloud SQL Proxy] Add metadata keys to enable/disable Hive Cloud SQL metastore

jphalip:
This makes it easier to control those parameters when creating a new Dataproc cluster


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -20,11 +20,11 @@ set -euxo pipefail
 
 # Whether to configure the Hive metastore to point to a Cloud SQL database.
 # This is not required for Hive & Spark I/O.
-readonly ENABLE_CLOUD_SQL_METASTORE=1 # 0 -> false
+readonly ENABLE_CLOUD_SQL_METASTORE="$(/usr/share/google/get_metadata_value attributes/enable-cloud-sql-hive-metastore || 1)"
```
karth295: I think `$()` is going to expect a command instead of `|| 1`. Maybe use the `false` command (and same suggestion for the line below)

cloud-sql-proxy/README.MD
```diff
@@ -49,6 +49,6 @@ You can use this initialization action to create a Dataproc cluster using a shar
...
-* This initalization action can be reconfigured to only install a proxy on the master, if you only wish to have a shared metastore. Set `ENABLE_PROXY_ON_WORKERS` to 0.
-* It can be also be reconfigured to istall proxies without changing the Hive metasore, if you only wish to have the proxies. Set `ENABLE_CLOUD_SQL_METASTORE` to 0 and do not set the `hive-metastore-instance` metadata key.
+* This initalization action can be reconfigured to only install a proxy on the master, if you only wish to have a shared metastore. Set the `enable-cloud-sql-proxy-on-workers` metadata key to `0`.
+* It can be also be reconfigured to istall proxies without changing the Hive metasore, if you only wish to have the proxies. Set the `enable-cloud-sql-hive-metastore` metadata key to `0` and do not set the `hive-metastore-instance` metadata key.
```
karth295: Since you're touching this line anyway, "istall" -> "install"

jphalip: Sure, I'll update the PR with the typo fix

cloud-sql-proxy/cloud-sql-proxy.sh
jphalip: Ok yeah I'll use a boolean value instead of a digit

jphalip:
@karth295 Thanks for the review!

I have one reservation about the variable naming. I think generally it's best to use a positive name for boolean variables, so I think "enable" would be best even if it defaults to true. See the reasoning here: http://truelogic.org/wordpress/2009/02/27/how-not-to-name-a-boolean-variable/

What do you think?

As a aside, I also think it would have been better to NOT enable the Hive metastore by default since the init action is simply called "cloud-sql-proxy" (i.e. the name does not refer to Hive). But that's probably too late to change, otherwise that would break legacy code 😀

cloud-sql-proxy/cloud-sql-proxy.sh
karth295: Well `false` is a command (likely `/bin/false`) that exits with a false-y code. I don't think it would have worked as a digit.

karth295:
Sure -- `enable-foo` sounds good to me.

I kinda understand your aside. This init action predates me joining the team, but my understanding is that it was written to let you use `hive` but keep your data outside the cluster. Reading/writing to cloud sql was a secondary use case.

jphalip:
@karth295 Thanks for the feedback.

I've updated the handling of variables and used `|| echo 'true'` instead of `|| true` since the `true` command returns no output, and the user may set the metadata to the `"false"` string.

Let me know what you think.

karth295:
^Pushed a commit clarifying how to run init action without configuring hive metastore

PR #255, opened 2018-04-25
Fix typos in README

findepi:


presto/README.MD
```diff
@@ -13,13 +13,13 @@ You can use this initialization action to create a new Dataproc cluster with Pre
     --initialization-actions gs://<GCS_BUCKET>/presto.sh   
     --initialization-action-timeout 5m
     ```
-1. Once the cluster has been created, Presto is configured to run on port `8080` (though you can change this int he script) on the master node in a Cloud Dataproc cluster. To connect to the Presto web interface, you will need to create an SSH tunnel and use a SOCKS 5 Proxy as described in the [dataproc web interfaces](https://cloud.google.com/dataproc/cluster-web-interfaces) documentation. You can also use the [Presto command line interface](https://prestodb.io/docs/current/installation/cli.html) using the `presto` command on the master node.
```
findepi: GH doesn't show that, but here, `int he` was changed to `in the`

PR #259, opened 2018-05-01
Add an initialization action to install OpenSSL 1.0.2 from backports.

AngusDavis:
NPN has been deprecated for quite some time with ALPN being its
replacement. OpenSSL 1.0.1, which is included in Debian 8, does not
support ALPN, only NPN. OpenSSL 1.0.2 supports ALPN.

This initialization action installs OpenSSL 1.0.2 from jessie backports.
Using this initialization action can be useful if using gRPC with
netty-tcnative-dynamic libraries and encountering NPN / ALPN negotation
failures when communicating with Google services from Dataproc clusters.

### DIFF
--- a/openssl/README.md
+++ b/openssl/README.md
@@ -5,15 +5,13 @@ from Jessie-backports for Dataproc clusters running Dataproc 1.0 through 1.2.
 
 ## Using this initialization action
 You can use this initialization action to create a new Dataproc cluster with the backports version
-of OpenSSL using the following instructions:
+of OpenSSL using the following command:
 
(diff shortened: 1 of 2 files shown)


karth295:
^ @AngusDavis if you don't mind I pushed a commit modifying the README to suggest creating a cluster using our GCS mirror. I've been suggesting that for all new init actions to make them faster to try out.

I guess I should add a note to the top level readme explaining that you should copy init actions to avoid being broken by breaking changes.

AngusDavis:
Ah, nice change to the README. I was cribbing off of another init action for that.

PR #263, opened 2018-05-07
Fixed overwrite problem with zookeeper

adam-stasiak:
Fixed problem with /var/lib/zookeeper/myid file that contains node name instead just simple number.

For valid zookeeper configuration it should be myid prepared as there https://stackoverflow.com/a/11498597


adam-stasiak:
@karth295 could you check this change?

DariuszAniszewski:
@karth295 is there something stopping you from merging this change?

karth295:
@DariuszAniszewski just time haha

PR #264, opened 2018-05-07
[Breaking change] Run Tez UI in App Timeline Server

karth295:
Changes Tez UI from http://clustername-m:9999 to
http://clustername-m:8188/tez.

### DIFF
--- a/tez/README.MD
+++ b/tez/README.MD
@@ -2,23 +2,17 @@
 
 This initialization action installs the latest version of [Apache Tez](https://tez.apache.org/) on a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster. In this initialization action:
 
-1. Required libraries such as [`protobuf`](https://github.com/google/protobuf) are installed
-2. Tez is compiled
-3. Tez is put into HDFS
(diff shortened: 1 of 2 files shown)


karth295:
This avoids needing to run a separate webserver for the Tez UI

karth295:
The nodejs install script does not retry on `apt-get update` failures, which causes the init action to fail occasionally. This PR may also fix that problem, since we do retry `apt-get update`.

```
W: Failed to fetch https://deb.nodesource.com/node_8.x/dists/jessie/main/binary-amd64/Packages  Hash Sum mismatch

E: Some index files failed to download. They have been ignored, or old ones used instead.
Error executing command, exiting
+ err 'Failed to setup node 8.'
++ date +%Y-%m-%dT%H:%M:%S%z
+ echo '[2018-05-15T23:32:39+0000]: Failed to setup node 8.'
[2018-05-15T23:32:39+0000]: Failed to setup node 8.
+ return 1
```

karth295:
For posterity, I ended up setting the web ui path to `/tez-ui` to match Tez in Dataproc 1.3 preview.

Issue #268, opened 2018-05-16
Add Sqoop 


szewi:
We should add sqoop to dataproc. 

Sqoop is tool that allows to import/export data between different types of databases, it also allows to import data to hadoop to perform MR jobs. 
There is `cloud-sql-proxy` init action working, so we should configure sqoop to allow to pull the data from cloud SQL, we should also enable Sqoop jobs for BigTable.

Bidyadharbarik84:
Can you upload some sample script from cloud dataproc using sqoop

szewi:
@Bidyadharbarik84 Please take a look on some example workflow. I assume you copy sqoop.sh from PR to your GCS bucket:
1. Create cloud SQL database mysql and name it `test-sql`
2. Create Bigtable dev instance and name it `test-bigtable`
3. Create cluster with sqoop:
`gcloud dataproc clusters create sqoop-single --initialization-actions gs://dataproc-initialization-actions/cloud-sql-proxy/cloud-sql-proxy.sh,gs://dataproc-initialization-actions/hive-hcatalog/hive-hcatalog.sh,gs://dataproc-initialization-actions/bigtable/bigtable.sh,gs://<your-bucket-with-sqoop-init-action>/sqoop.sh         --scopes cloud-platform         --properties hive:hive.metastore.warehouse.dir=gs://<your-bucket-for-hive>/hive-warehouse --metadata "hive-metastore-instance=<you-project>:<you-region>:test-sql" --metadata "bigtable-project=<you-project>" --metadata "bigtable-instance=test-bigtable" --single-node`
4. Log in: `gcloud compute ssh sqoop-single-m`
5. Clone some test db: `git clone https://github.com/datacharmer/test_db` and `cd test_db`
6. Upload some test _db to Cloud SQL with command `mysql -u root < employees.sql`
7. Run sqoop import: 
/usr/lib/sqoop/bin/sqoop import --connect jdbc:mysql://localhost/employees --username root --table employees --columns "emp_no,first_name" --hbase-table employees --column-family my-column-family -hbase-row-key emp_no --m 1 --hbase-create-table

This command will import some columns from Cloud SQL employees db to BigTable instance using hbase binaries. You should see sth like this in the output:

> 19/02/19 16:01:50 INFO mapreduce.ImportJobBase: Transferred 0 bytes in 20.5951 seconds (0 bytes/sec)
[...]

PR #269, opened 2018-05-16
Sqoop init action (#268).

szewi:


sqoop/README.MD
```diff
@@ -0,0 +1,74 @@
+# Sqoop 
+
+This initialization action installs [Sqoop](http://sqoop.apache.org/) on a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster. Sqoop is being compiled from trunk branch of official [github](https://github.com/apache/sqoop) mirror.
+
+## Using this initialization action
+You can use this initialization action to create a new Dataproc cluster with Sqoop installed by:
+
+1. Uploading a copy of this initialization action (`sqoop.sh`) to [Google Cloud Storage](https://cloud.google.com/storage).
```
karth295: Sorry I haven't merged my README updates, but I'm planning to change the READMEs to just point at our GCS mirror: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/262.

Please follow the style from that PR :)

sqoop/README.MD
```diff
@@ -0,0 +1,74 @@
...
+
+You can find more information about using initialization actions with Dataproc in the [Dataproc documentation](https://cloud.google.com/dataproc/init-actions).
+
+## Using Sqoop with BigTable
+1. Sqoop can be used to import data into BigTable. In order to make it working, bigtable init action must be added to initialization-actions list.
+
+    ```bash
+    gcloud dataproc clusters create <CLUSTER_NAME> \
+        --initialization-action-timeout 15m --initialization-actions \
+        gs://dataproc-initialization-actions/cloud-sql-proxy/cloud-sql-proxy.sh,\
```
karth295: Nit: you don't need to `cloud-sql-proxy` init action here

sqoop/README.MD
```diff
@@ -0,0 +1,74 @@
...
+        gs://dataproc-initialization-actions/hive-hcatalog/hive-hcatalog.sh,\
+        gs://dataproc-initialization-actions/bigtable/bigtable.sh,\
+        gs://dataproc-initialization-actions/sqoop/sqoop.sh \
+        --single-node --master-machine-type n1-standard-4 --scopes sql-admin,\
```
karth295: Similarly omit `sql-admin` here

sqoop/sqoop.sh
```diff
@@ -0,0 +1,73 @@
...
+# Dataproc cluster. Hive-Hcatalog is not installed automatically and if you want
+# to run hive jobs than please use this script with hive-hcatalog.sh init action.
+# HBase libraries are not installed automatically so in order to run imports to 
+# BigTable using sqoop, please run this init action with bigtable.sh.
+
+set -euxo pipefail
+
+readonly SQOOP_HOME='/usr/lib/sqoop'
+readonly SQOOP_CODE_LINK='https://github.com/apache/sqoop'
+readonly BIGTABLE_HBASE_CLIENT='bigtable-hbase-1.x-hadoop-1.3.0.jar'
```
karth295: If you're requiring the BT init action, is this still necessary?

sqoop/README.MD
```diff
@@ -0,0 +1,74 @@
+# Sqoop 
+
+This initialization action installs [Sqoop](http://sqoop.apache.org/) on a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster. Sqoop is being compiled from trunk branch of official [github](https://github.com/apache/sqoop) mirror.
```
karth295: Compiling from `trunk` is a little scary. We should at least compile from a release tag and just cherry-pick the necessary commit in.

@pmkc will we be able to backport [the commit](https://github.com/apache/sqoop/commit/e13dd21209c26316d43350a23f5d533321b61352) into Dataproc so we don't need to compile Sqoop in this init action?

sqoop/sqoop.sh
```diff
@@ -0,0 +1,73 @@
...
+
+function err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
+  return 1
+}
+
+function install_sqoop() {
+  git clone ${SQOOP_CODE_LINK} ${SQOOP_HOME} && cd ${SQOOP_HOME} && ant \
+    || err 'Compiling Sqoop from source code failed.'
+  echo "export PATH=\"${SQOOP_HOME}/bin:$PATH\"" >> /etc/profile
```
pmkc: Source builds are pretty much forbidden in our init actions for performance reasons.

If Sqoop is not available as a binary distribution I would drop this PR and resubmit it when it's fixed in our Bigtop repo.

sqoop/README.MD
pmkc: Yes you do if you are using Cloud SQL as the RDBMS

sqoop/README.MD
pmkc: I always recommending using cloud-admin instead.

sqoop/README.MD
pmkc: This CL should not be compiling at all. It should certainly ignore BigTable until Sqoop is fixed in Bigtop with that patch

sqoop/README.MD
szewi: ok

sqoop/README.MD
szewi: I am using sqoop to export data from cloud SQL and import data to BigTable in single sqoop job so I need it for that example.

sqoop/README.MD
szewi: Do you mean `cloud-platform` scope? 
cloud-admin results in an error: `Invalid service account scope: 'cloud-admin'`

sqoop/README.MD
szewi: Let me know when Sqoop will be available in Bigtop. I will test it and update this PR properly, so far I will just update docs.

sqoop/sqoop.sh
szewi: After adding HBase libs to classpath in bigtable init actions it is not needed now.

sqoop/sqoop.sh
szewi: @pmkc if you would like as temporary solution I can pack compiled sqoop in some lightweight docker container ? WDYT?

szewi:
@karth295 This is ready for another round of reviewing. Please take a look.

szewi:
@karth295 I have pushed docs fixes. 
@pmkc is the any update regarding adding scoop to Bigtop?

sqoop/sqoop.sh
karth295: @szewi I think it will be a little while until this is in Bigtop. For now, I suggest we host a mirror on GCS.

I'd prefer not to build from trunk or any non-release commit, since non-release commits are not guaranteed to be stable. Please run `git cherry-pick e13dd21209c26316d43350a23f5d533321b61352` onto `release-1.4.7-rc0`, resolve the merge conflicts, run Sqoop's tests, and build a jar. I can then put that jar in the appropriate bucket.

If that's more complicated than just minor merge conflicts, then we can build directly from `e13dd21209c26316d43350a23f5d533321b61352` and host that.

Either way, this init action should be hermetic and should not build from trunk.

szewi: @karth295 I cherry-picked `e13dd21209c26316d43350a23f5d533321b61352` onto release-1.4.7-rc0 resolved conflicts, but tests were not passing. I investigated this and decided to first cherry-pick commit `18445290810b1df035e06fb074064d6b9c1d6e90` which was solving the issues with tests and also allowed to apply then `e13dd21209c26316d43350a23f5d533321b61352` without conflicts. Then I compiled sqoop on a clean dataproc cluster and uploaded sqoop to publicly available bucket. Most probably you would like to move sqoop.jar to your own place on google storage, so please let me know about final location, then I will update URL properly. 

Sqoop is being delivered with `bin/` directory containing required executable scripts, so right now sqoop code is downloaded to /usr/lib/sqoop, but no compilation is done. I changed the source code location to my sqoop fork with the branch `sqoop-bigtable-client` containing changes I have applied, so the end user will have exact sources used for creating sqoop jar in /usr/lib/sqoop. Let me know if you are fine with this.

sqoop/sqoop.sh
szewi: Right now when sqoop is not compiled on the cluster, but downloaded as compiled jar from somewhere else, we need to provide a minimal set of dependencies, because during compilation several jars were downloaded and added to classpath [see: bin/configure-sqoop](https://github.com/szewi/sqoop/blob/sqoop-bigtable-client/bin/configure-sqoop#L206). There are several already available on dataproc clusters, but the only missing is avro-1.8.1, so I downloaded it manually and added to init action script.

szewi: `bigtable-hbase-1.x-hadoop-1.3.0.jar` connector is a different story. When I use BigTable init action along with sqoop, the BigTable init action already installs this jar file in `/usr/lib/hbase/lib` and put those on the classpath. Then it's being passed to sqoop in [bin/configure-sqoop](https://github.com/szewi/sqoop/blob/sqoop-bigtable-client/bin/configure-sqoop#L175), but it doesn't work with class not found exception.(I even printed sqoop classpath and bigtable-hbase jar is there!) 
Solution is either to symlink or copy `bigtable-hbase-1.x-hadoop-1.3.0.jar` to /usr/lib/sqoop/lib directory, which is a natural way of installing sqoop connectors. 
I was thinking about automatically detecting which init actions were used to create a cluster and then if there was a `bigtable` init action getting/linknig right jar file, but this doesn't seem to me like a proper way of handling that, explicitily installing connector will present to users how to easily add more connectors, what do you think @karth295 ?

szewi:
Hi, @karth295 I have updated init action and README file, added comments and questions for decisions I made(not sure why not visible on main GitHub page, but just go to latest commit).

Issue #238, opened 2018-04-02
Support Jupyterlab in Jupyter init action


karth295:
Theoretically you can `pip install jupyterlab` and it's available at `<masterip>:8123/lab`. I'd rather keep Juypter as the default, and Jupyterlab accessible at `/lab`, at least until Jupyterlab is GA.

It didn't work for me on the first try -- I got a 404 error popup. Digging in, it is `{"message": "No such file: <cluster-gcs-bucket>/notebooks", "reason": null}`. Since we are using a GCS content manager, I think this is actually this Jupyter bug: https://github.com/jupyterlab/jupyterlab/issues/3796.

karth295:
Ah I see this was actually a bug in `jgscm`, the GCS content manager, and it was fixed in https://github.com/src-d/jgscm/commit/d1f1f4bda80cf8559358926f9c6250b2ce18d247

So we just need to `jgscm` to cut a new version and it should just work.

karth295:
(The jgscm PR was https://github.com/src-d/jgscm/pull/9)

CarloLucibello:
Creating the cluster with this command
```bash

    gcloud dataproc clusters create cluster-1 --bucket storage-north1 --subnet default \
           --zone europe-north1-b --master-machine-type n1-standard-1 --master-boot-disk-size 100 \
           --num-workers 2 --worker-machine-type n1-standard-1 --worker-boot-disk-size 100 \
           --image-version 1.3-deb9 --scopes 'https://www.googleapis.com/auth/cloud-platform' \
           --project bigquery-220317 \
           --initialization-actions  'gs://dataproc-initialization-actions/jupyter/jupyter.sh'\
           --metadata "JUPYTER_CONDA_PACKAGES=jupyterlab:numpy:pandas:matplotlib:pyspark"
```
I can access jupyterlab at `localhost:8123/lab` but I immediatly get a pop up window with the error
```
 Directory not found: ""`
```
Then I can create new notebooks, but the file explorer pane on the left is empty and unusable.

Can we add back support to jupyterlab in jupyter init action, i.e. re-revert #250 (if that's possible), and fix this bug?

MaxAVF:
Iḿ having the same problem installing Jupyter Lab
```
Directory not found: ""
```
But if you open a notebook from Jupyter notebook, it shows up as running on JupyterLab and you can open it from there... There is a way to fix this?

karth295:
[...]

PR #273, opened 2018-05-29
Revert "Install Jupyterlab in jupyter init actions"

karth295:
Reverts GoogleCloudPlatform/dataproc-initialization-actions#250


CarloLucibello:
do we need to add the conda-forge channel for this to  work?
https://anaconda.org/conda-forge/jupyterlab
Cf. #238

PR #274, opened 2018-05-30
Add package of tools for customizing python environment

DanSedov:


python/pip-install.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -exo pipefail
+
+readonly PACKAGES=$(/usr/share/google/get_metadata_value attributes/PIP_PACKAGES || true)
+
+function main() {
+  if [[ -z "${PACKAGES}" ]]; then
+    echo "ERROR: Must specify PIP_PACKAGES metadata key"
```
aniket486: How about keeping this optional?

python/pip-install.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -exo pipefail
+
+readonly PACKAGES=$(/usr/share/google/get_metadata_value attributes/PIP_PACKAGES || true)
```
aniket486: nit: rename PACKAGES to PIP_PACKAGES.

python/conda-install.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -exo pipefail
+
+readonly PACKAGES=$(/usr/share/google/get_metadata_value attributes/CONDA_PACKAGES || true)
```
aniket486: s/PACKAGES/CONDA_PACKAGES

PR #276, opened 2018-06-05
GC Dataproc image version 1.2.37 expects zkServer.sh 

mirohla:
script in /usr/share/zookeeper/bin instead of /usr/lib/zookeeper/bin.

### DIFF
--- a/zookeeper/zookeeper.sh
+++ b/zookeeper/zookeeper.sh
@@ -53,4 +53,4 @@ server.3=${CLUSTER_NAME}-w-1:2888:3888
 EOF
 
 # Start ZooKeeper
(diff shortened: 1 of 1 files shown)


mirohla:
CLA signed

ekampf:
Had the same problem on my cluster, this fix worked.

soelu:
Any chance of getting this approved soon? This somehow blocks our deployment atm.

karth295:
Thanks for the patch and feedback folks. You can (and should) copy the initialization action to your own GCS bucket for your production pipelines: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions#how-initialization-actions-are-used. 1) That protects your automation in case we make a breaking change, and 2) you don't need to wait on us merging PRs.

PR #278, opened 2018-06-07
Add a check for schema compatibility to cloud-sql-proxy

DanSedov:
When switching Dataproc versions, metastore schema may become incompatible. When this happens, we can use schematool to report a concise error and fail.

$ /usr/lib/hive/bin/schematool -dbType mysql -info
Hive distribution version:       2.3.0
Metastore schema version:        2.1.0
org.apache.hadoop.hive.metastore.HiveMetaException: Metastore schema version is not compatible. Hive Version: 2.3.0, Database Schema Version: 2.1.0
*** schemaTool failed ***

User can upgrade their schema by running:
$ /usr/lib/hive/bin/schematool -dbType mysql -upgradeSchemaFrom 2.1.0

I am not sure if automating upgrade is a welcome change. Could we potentially upgrade schema in a way that it could no longer be used with previous Dataproc versions?

### DIFF
--- a/cloud-sql-proxy/README.MD
+++ b/cloud-sql-proxy/README.MD
@@ -146,6 +146,23 @@ Google Cloud's key management service. You will only need to encrypt and provide
         --metadata "db-hive-password-uri=gs://<SECRETS_BUCKET>/hive-password.encrypted"
     ```
 
+8. Upgrading schema (create cluster step failed on new Dataproc version):
+
+    When changing Dataproc versions, metastore may detect schema as obsolete. In this case a schema upgrade may be necessary. Log into master node on a cluster with new Dataproc version, and run these commands:
+    
+    ```bash
+    $ /usr/lib/hive/bin/schematool -dbType mysql -info
+    Hive distribution version: 2.3.0
+    Metastore schema version: 2.1.0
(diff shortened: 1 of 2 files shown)


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -202,6 +202,9 @@ EOF
     echo "Service hive-metastore is not loaded"
   fi
 
+  # Check that metastore schema is compatible.
+  /usr/lib/hive/bin/schematool -dbType mysql -info
```
karth295: Can you add something like `|| err 'Run /usr/lib/hive/bin/schematool -dbType mysql -upgradeSchemaFrom 2.1.0 to upgrade the schema. Note that this may break Hive metastores that depend on the old schema'`

DanSedov: Done

cloud-sql-proxy/README.MD
```diff
@@ -146,6 +146,23 @@ Google Cloud's key management service. You will only need to encrypt and provide
         --metadata "db-hive-password-uri=gs://<SECRETS_BUCKET>/hive-password.encrypted"
     ```
 
+8. Upgrading schema (create cluster step failed on new Dataproc version):
```
karth295: Nit: write out the error message you'll see

DanSedov:
Done

PR #288, opened 2018-06-28
skip hive metastore in cloud sql setup for masters other than master0 in HA mode

memoryoflife:
When creating a HA cluster, mysql cli is uninstalled on all masters other than master0, so we can't use mysql cli to setup hive metastore on those nodes. Since all the masters point to the same db in cloud sql as the hive metastore, we need only master0 to setup the metastore, and the other masters only need to update "hive.metastore.warehouse.dir", which can be fetched using beeline to query hive server directly.

### DIFF
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -96,7 +96,7 @@ function configure_proxy_flags() {
     elif ! [[ "${metastore_instance}" =~ .+:.+:.+ ]]; then
       err 'hive-metastore-instance must be of form project:region:instance'
     elif ! [[ "${metastore_instance}" =~ =tcp:[0-9]+$ ]]; then
-      metastore_instance+="=tcp:0.0.0.0:${metastore_proxy_port}"
+      metastore_instance+="=tcp:${metastore_proxy_port}"
     else
       metastore_proxy_port="${metastore_instance##*:}"
     fi
@@ -208,7 +208,7 @@ EOF
 function run_validation() {
   if ( systemctl is-enabled --quiet hive-metastore ); then
     # Start metastore back up.
-    systemctl start hive-metastore \
+    systemctl restart hive-metastore \
       || err 'Unable to start hive-metastore service'
   else
     echo "Service hive-metastore is not loaded"
@@ -256,18 +256,18 @@ function run_with_retries() {
 
 
 function configure_hive_warehouse_dir(){
-  run_with_retries nc -zv "${MASTER_HOSTNAMES[0]}" 3306
+  run_with_retries run_validation
 
-  HIVE_WAREHOURSE_URI=$(beeline -u jdbc:hive2://localhost:10000 -e "describe database default;" 2>/dev/null \
+  HIVE_WAREHOURSE_URI=$(beeline -u jdbc:hive2://localhost:10000 \
+    -e "describe database default;" 2>/dev/null \
     | sed '4q;d' | cut -d "|" -f4 | tr -d '[:space:]')
+  echo $HIVE_WAREHOURSE_URI
 
   bdconfig set_property \
     --name 'hive.metastore.warehouse.dir' \
     --value "${HIVE_WAREHOURSE_URI}" \
     --configuration_file /etc/hive/conf/hive-site.xml \
     --clobber
-
-  run_validation
 }
 
 function main() {
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -224,7 +224,7 @@ EOF
       || err 'Failed to set mysql schema.'
   fi
 
-  run_validation
+  run_with_retries run_validation
 }
 
 function run_validation() {
@@ -243,8 +243,9 @@ function run_validation() {
   # Validate it's functioning.
   if ! beeline -u jdbc:hive2://localhost:10000 -e 'SHOW TABLES;' >& /dev/null; then
     err 'Failed to bring up Cloud SQL Metastore'
+  else
+    echo 'Cloud SQL Hive Metastore initialization succeeded' >&2
   fi
-  echo 'Cloud SQL Hive Metastore initialization succeeded' >&2
 
 }
 
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -76,6 +76,14 @@ readonly PROXY_BIN='/usr/local/bin/cloud_sql_proxy'
 readonly INIT_SCRIPT='/usr/lib/systemd/system/cloud-sql-proxy.service'
 readonly ADDITIONAL_INSTANCES_KEY='attributes/additional-cloud-sql-instances'
 
+# Dataproc master nodes information
+readonly DATAPROC_MASTER=$(/usr/share/google/get_metadata_value attributes/dataproc-master)
+readonly DATAPROC_MASTER_ADDITIONAL=$(/usr/share/google/get_metadata_value attributes/dataproc-master-additional || true)
+readonly MASTER_HOSTNAMES=($DATAPROC_MASTER ${DATAPROC_MASTER_ADDITIONAL//,/ })
+readonly MASTER_COUNT=${#MASTER_HOSTNAMES[@]}
+readonly MY_HOSTNAME=$(hostname -s)
+
+
 function err() {
   echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
   return 1
@@ -88,7 +96,7 @@ function configure_proxy_flags() {
     elif ! [[ "${metastore_instance}" =~ .+:.+:.+ ]]; then
       err 'hive-metastore-instance must be of form project:region:instance'
     elif ! [[ "${metastore_instance}" =~ =tcp:[0-9]+$ ]]; then
-      metastore_instance+="=tcp:${metastore_proxy_port}"
+      metastore_instance+="=tcp:0.0.0.0:${metastore_proxy_port}"
     else
       metastore_proxy_port="${metastore_instance##*:}"
     fi
@@ -194,6 +202,10 @@ EOF
       || err 'Failed to set mysql schema.'
   fi
 
+  run_validation
+}
+
+function run_validation() {
   if ( systemctl is-enabled --quiet hive-metastore ); then
     # Start metastore back up.
     systemctl start hive-metastore \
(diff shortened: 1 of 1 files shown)


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -214,6 +226,50 @@ EOF
...
+
+  if ! (( ${update_succeeded} )); then
+    echo "Final attempt of '${cmd[*]}'..."
+    # Let any final error propagate all the way out to any error traps.
+    "${cmd[@]}"
+  fi
+}
+
+
+function configure_hive_warehouse_dir(){
```
mikaylakonst: Style nit: I think it would be clearer if you pulled the part of configure_proxy_flags that extracts the hive warehouse uri out of that method and put it here instead, so that this method looks like if I'm master 0, extract the uri one way, else extract it the other way

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -88,7 +96,7 @@ function configure_proxy_flags() {
     elif ! [[ "${metastore_instance}" =~ .+:.+:.+ ]]; then
       err 'hive-metastore-instance must be of form project:region:instance'
     elif ! [[ "${metastore_instance}" =~ =tcp:[0-9]+$ ]]; then
-      metastore_instance+="=tcp:${metastore_proxy_port}"
+      metastore_instance+="=tcp:0.0.0.0:${metastore_proxy_port}"
```
mikaylakonst: Why is this necessary?

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -214,6 +226,50 @@ EOF
...
+  if ! (( ${update_succeeded} )); then
+    echo "Final attempt of '${cmd[*]}'..."
+    # Let any final error propagate all the way out to any error traps.
+    "${cmd[@]}"
+  fi
+}
+
+
+function configure_hive_warehouse_dir(){
+  run_with_retries nc -zv "${MASTER_HOSTNAMES[0]}" 3306
```
mikaylakonst: I'm not sure this is what you want. This will succeed at the very beginning of master 0's setup when mysql server is still running on 3306. You might actually just be able to run the next line with retries

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -214,6 +226,50 @@ EOF
...
+
+  HIVE_WAREHOURSE_URI=$(beeline -u jdbc:hive2://localhost:10000 -e "describe database default;" 2>/dev/null \
+    | sed '4q;d' | cut -d "|" -f4 | tr -d '[:space:]')
```
mikaylakonst: Wondering if there is a more elegant way to do this - describe database default gives you a table of info about the database - maybe you could use the table as input and do select location from (the table)

cloud-sql-proxy/cloud-sql-proxy.sh
memoryoflife: this method is not called for master0, I'd rather keep the flow control in main(), and use this method for master1,2... while keep the configure_sql_client method for master0 unchanged, this way it's easier to figure out what's been run on master0 and other master.

cloud-sql-proxy/cloud-sql-proxy.sh
memoryoflife: Checked, it's only possible to query the table in metastore db, in hive server, the meta data is not stored in any table.

cloud-sql-proxy/cloud-sql-proxy.sh
memoryoflife: it was needed because otherwise it'll just listen on localhost, and the other machines won't be able to connect to it. But it's no longer needed now.

cloud-sql-proxy/cloud-sql-proxy.sh
memoryoflife: Good point. After a few test run, I managed to figure out the correct order, should run_validation until it succeeds before moving on

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -259,12 +323,18 @@ function main() {
...
+        systemctl disable mysql
+      fi
     fi
     install_cloud_sql_proxy
     if [[ $enable_cloud_sql_metastore = "true" ]]; then
-      configure_sql_client
+      if [[ "${MASTER_INDEX?}" == 0 ]]; then
+        configure_sql_client
+      else
+        configure_hive_warehouse_dir
```
karth295: I think you need to run configure_hive_warehouse_dir on `-m-0` as well. If not, you should add a comment on why it isn't necessary.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -214,6 +226,50 @@ EOF
 
 }
 
+
+# Helper to run any command with default bdutil settings with Fibonacci
+# backoff. If all retries fail, returns last attempt's exit code.
+# Args: "$@" is the command to run.
+function run_with_retries() {
```
karth295: I have a shorter version of this utility specifically for running `apt-get update` with retries. https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/zeppelin/zeppelin.sh#L27

Since we already run init actions with set -x, this additional logging is generally unnecessary. In addition, I don't think exponential backoff is necessary here.

karth295: Also, consider putting this at the top of the file with the other utilities (like `err`)

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -214,6 +226,50 @@ EOF
...
+  if ! (( ${update_succeeded} )); then
+    echo "Final attempt of '${cmd[*]}'..."
+    # Let any final error propagate all the way out to any error traps.
+    "${cmd[@]}"
+  fi
+}
+
+
+function configure_hive_warehouse_dir(){
+  run_with_retries run_validation
```
karth295: Why does this require retries? (If it is necessary, please add a comment)

If this is just to block until the metastore comes back up, you may want to split restarting the metastore into its own method.

Also, does the invocation on line 205 need retries too?

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -259,12 +323,18 @@ function main() {
...
-      systemctl disable mysql
+      if ( systemctl is-enabled --quiet mysql ); then
+        systemctl stop mysql
+        systemctl disable mysql
+      fi
     fi
     install_cloud_sql_proxy
     if [[ $enable_cloud_sql_metastore = "true" ]]; then
-      configure_sql_client
+      if [[ "${MASTER_INDEX?}" == 0 ]]; then
```
karth295: You can get rid of several variables by making this `if [[ "${HOSTNAME}" == "${DATAPROC_MASTER}" ]]`. We do that in a couple other places, such as https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/jupyter2/jupyter2.sh#L109.

cloud-sql-proxy/cloud-sql-proxy.sh
memoryoflife: Ok, will move it to the top and remove logging. But I think fibonacci backoff could be better than fix rate here, since we are using that for the same purpose in some other scripts. But there's no specific justification whether to use it or not here.

cloud-sql-proxy/cloud-sql-proxy.sh
memoryoflife: because if you point to an existing db as hive metastore it'll just succeed, otherwise it needs to wait for master 0 to create it first, otherwise when you restart hive-metastore service it tries to connect to the db instance that's not yet created and it'll fail. 

line 205 doesn't need retry, because it runs on master0 only, and it runs after the db instance is created(if needed)

cloud-sql-proxy/cloud-sql-proxy.sh
memoryoflife: no, it's not needed, hive warehouse url will be set for master0 as part of configure_sql_client

cloud-sql-proxy/cloud-sql-proxy.sh
memoryoflife: thanks for the tips, I've applied the changes.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -76,11 +76,42 @@ readonly PROXY_BIN='/usr/local/bin/cloud_sql_proxy'
 readonly INIT_SCRIPT='/usr/lib/systemd/system/cloud-sql-proxy.service'
 readonly ADDITIONAL_INSTANCES_KEY='attributes/additional-cloud-sql-instances'
 
+# Dataproc master nodes information
+readonly DATAPROC_MASTER=$(/usr/share/google/get_metadata_value attributes/dataproc-master)
+readonly DATAPROC_MASTER_ADDITIONAL=$(/usr/share/google/get_metadata_value attributes/dataproc-master-additional || true)
```
mikaylakonst: looks like DATAPROC_MASTER_ADDITIONAL is never used?

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -76,11 +76,42 @@ readonly PROXY_BIN='/usr/local/bin/cloud_sql_proxy'
...
+
 function err() {
   echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
   return 1
 }
 
+# Helper to run any command with default bdutil settings with Fibonacci
```
mikaylakonst: nix "with default bdutil settings"

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -259,12 +310,20 @@ function main() {
       else
         echo "Service hive-metastore is not enabled"
       fi
-      systemctl stop mysql
-      systemctl disable mysql
+      if ( systemctl is-enabled --quiet mysql ); then
+        systemctl stop mysql
+        systemctl disable mysql
+      fi
```
mikaylakonst: else echo "Service mysql is not enabled"

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -259,12 +310,20 @@ function main() {
...
     fi
     install_cloud_sql_proxy
     if [[ $enable_cloud_sql_metastore = "true" ]]; then
-      configure_sql_client
+      if [[ "${HOSTNAME}" == "${DATAPROC_MASTER}" ]]; then
+        #initialize metastore db instance and set hive.metastore.warehouse.dir on master 0
```
mikaylakonst: nit: comment style here and throughout. Space after '#', capitalize first word, period at end

PR #295, opened 2018-07-19
Upgrade presto to use latest released version.

aniket486:
Also, add missing configuration properties to avoid crash looping of presto
server.


aniket486:
Error messages show up in the presto error logs.

PR #298, opened 2018-07-24
Drill url needs to be updated because the old one has expired and is not valid anymore

adam-stasiak:
I replaced apache.mirrors.hoobly.com with a link to the archive apache sources. Now init action passes. 
Related to the #299

### DIFF
--- a/drill/drill.sh
+++ b/drill/drill.sh
@@ -228,7 +228,7 @@ function main() {
   mkdir -p ${DRILL_HOME} && chown ${DRILL_USER}:${DRILL_USER} ${DRILL_HOME}
 
   # Download and unpack Drill as the pseudo-user.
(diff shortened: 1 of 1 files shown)


drill/drill.sh
```diff
@@ -228,7 +228,7 @@ function main() {
   mkdir -p ${DRILL_HOME} && chown ${DRILL_USER}:${DRILL_USER} ${DRILL_HOME}
 
   # Download and unpack Drill as the pseudo-user.
-  sudo wget http://apache.mirrors.hoobly.com/drill/drill-${DRILL_VERSION}/apache-drill-${DRILL_VERSION}.tar.gz
+  sudo wget http://archive.apache.org/dist/drill/drill-${DRILL_VERSION}/apache-drill-${DRILL_VERSION}.tar.gz
```
aniket486: Can you use one of the links from HTTP section of this page instead?
http://www.apache.org/dyn/closer.cgi/drill/drill-1.13.0/apache-drill-1.13.0.tar.gz

PR #304, opened 2018-07-27
Set CLOUDSDK_PYTHON to system python

### LINKED ISSUE #246: Conda init action should put `pip` and `conda` on $PATH, including for root user
It's often handy to ssh into a cluster and install extra packages after creating a cluster, but you pretty much need to use the absolute paths `/opt/conda/bin/conda` and `/opt/conda/bin/pip`.

karth295:
Fixes #246


karth295:
Oops actually fixes #174

PR #306, opened 2018-07-30
Install jupyter on all nodes to ensure consistent python version

### LINKED ISSUE #300: Bug in dataproc-initialization-actions/jupyter/ causes different version on master and worker
I'm getting the following error using a basic cluster initialised using the `dataproc-initialization-actions/jupyter/` scripts. 

> Exception: Python in worker has different version 3.6 than that in driver 3.7, PySpark cannot run with different minor versions.Please check environment variables PYSPARK_PYTHON and PYSPARK_DRIVER_PYTHON are correctly set.

I've confirmed this is the case: on the master version is 3.7.0 and on workers its 3.6.5. 

I've tracked down the issue to this segment of code: 
https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/dced28ebd3780307418789ac51bfe8303030cd9a/jupyter/jupyter.sh#L44-L52

`conda` defaults to upgrading python to 3.7.0 for installations of `jupyter` causing the different versions. 

I think wither the init script should prevent conda from installing the new python version, or perhaps upgrading the workers (I did this manually for now). I'm not familiar with conda (use pip instead), so unsure if there is a straightforward fix for this.

karth295:
Fixes #300

### DIFF
--- a/jupyter/jupyter.sh
+++ b/jupyter/jupyter.sh
@@ -41,12 +41,18 @@ if [ -n "${JUPYTER_CONDA_PACKAGES}" ]; then
   conda install ${JUPYTER_CONDA_PACKAGES//:/ }
 fi
 
+# Install jupyter on all nodes to start with a consistent python environment
+# on all nodes. Also, pin the python version to ensure that conda does not
+# update python because the latest version of jupyter supports a higher version
+# than the one already installed. See issue #300 for more information.
+PYTHON="$(ls /opt/conda/bin/python || which python)"
+PYTHON_VERSION="$(${PYTHON} --version 2>&1 | cut -d ' ' -f 2)"
+conda install jupyter matplotlib "python==${PYTHON_VERSION}"
+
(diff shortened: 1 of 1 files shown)


karth295:
^New commit also pins python version to the version already installed.

This lets you  can use `--metadata="CONDA_PACKAGES='python===3.7.0'" to pick your python version.

jupyter/jupyter.sh
```diff
@@ -41,12 +41,18 @@ if [ -n "${JUPYTER_CONDA_PACKAGES}" ]; then
   conda install ${JUPYTER_CONDA_PACKAGES//:/ }
 fi
 
-if [[ "${ROLE}" == 'Master' ]]; then
```
pmkc: You could still do this on master only. Up to you.

karth295: I still like the idea of a consistent environment, so leaving it on all nodes

PR #308, opened 2018-08-01
Add basic Docker init action

bsidhom:
This installs Docker CE from the public Docker Debian repository at a fixed version and configures it for use by YARN applications.

### DIFF
--- a/docker/docker.sh
+++ b/docker/docker.sh
@@ -1,0 +1,125 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+# TODO: Allow this to be configured by metadata.
+readonly DOCKER_VERSION='18.06.0~ce~3-0~debian'
+
+function is_master() {
+  local role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+  if [[ "$role" == 'Master' ]] ; then
+    true
+  else
+    false
+  fi
+}
+
+function get_docker_gpg() {
+  # Hard code the GPG key to prevent malicious payloads. The key was originally
+  # fetched by running:
+  # curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo
+  # "$ID")/gpg
+  cat <<EOF
+-----BEGIN PGP PUBLIC KEY BLOCK-----
+
+mQINBFit2ioBEADhWpZ8/wvZ6hUTiXOwQHXMAlaFHcPH9hAtr4F1y2+OYdbtMuth
+lqqwp028AqyY+PRfVMtSYMbjuQuu5byyKR01BbqYhuS3jtqQmljZ/bJvXqnmiVXh
+38UuLa+z077PxyxQhu5BbqntTPQMfiyqEiU+BKbq2WmANUKQf+1AmZY/IruOXbnq
+L4C1+gJ8vfmXQt99npCaxEjaNRVYfOS8QcixNzHUYnb6emjlANyEVlZzeqo7XKl7
+UrwV5inawTSzWNvtjEjj4nJL8NsLwscpLPQUhTQ+7BbQXAwAmeHCUTQIvvWXqw0N
+cmhh4HgeQscQHYgOJjjDVfoY5MucvglbIgCqfzAHW9jxmRL4qbMZj+b1XoePEtht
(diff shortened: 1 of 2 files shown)


bsidhom:
R: @axelmagn

docker/docker.sh
```diff
@@ -0,0 +1,125 @@
...
+    true
+  else
+    false
+  fi
+}
+
+function get_docker_gpg() {
+  # Hard code the GPG key to prevent malicious payloads. The key was originally
+  # fetched by running:
+  # curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo
```
pmkc: you hard code debian below, just hard code it here

docker/docker.sh
```diff
@@ -0,0 +1,125 @@
...
+  local role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+  if [[ "$role" == 'Master' ]] ; then
+    true
+  else
+    false
+  fi
+}
+
+function get_docker_gpg() {
+  # Hard code the GPG key to prevent malicious payloads. The key was originally
```
pmkc: Storing a key in a heredoc is pretty unreadable.

In general we trust HTTPS and re-download keys over it on a per installation basis.

If you really don't want to trust docker.com to not be compromised, why not host this key in a GCS bucket we control.

karth295:
Also fwiw GKE uses a much newer version of docker -- I would just use `curl | bash`. According to Docker's github, this is the command they suggest: https://github.com/docker/docker-install.

```
curl -fsSL get.docker.com -o get-docker.sh
sh get-docker.sh
```

pmkc:
Docker actually recommends: https://docs.docker.com/install/linux/docker-ce/debian/#install-docker-ce for more principled installations, which I think this is based on.

docker/docker.sh
bsidhom: Done.

docker/docker.sh
bsidhom: Storing it in a GCS bucket seems reasonable, though I do prefer keeping it inline to make the init action more atomic. Why do you find the heredoc unreadable? The function just does one thing and makes that pretty clear.

Alternatively, I _could_ just curl inline here. I didn't want to do that mostly because I'm not sure what happens with `apt-key add` if the payload is something other than a well-formed public key. I don't have a way to verify the key outside of using HTTPS anyway, so all I did here was check by hand that this was in fact the shape we expected.

bsidhom:
@karth295 The intent is not to enable Docker containers in YARN, but to make the Docker command available to the master and all workers. If we want to add that sort of functionality through configuration, we can do that in a follow-up PR. However, I would like to make that configuration optional.

As Patrick mentioned, this installs the latest Docker from the same repository that the `get-docker.sh` script does.

docker/docker.sh
pmkc: I vote inline then. Curl into | apt-key is pretty standard.

docker/docker.sh
```diff
@@ -0,0 +1,124 @@
...
+      return 0
+    fi
+    sleep 5
+  done
+  return 1
+}
+
+function install_docker() {
+  update_apt_get
+  apt-get install -y apt-transport-https ca-certificates curl gnupg2 \
```
pmkc: I'm 90% we have all of these besides apt-transport-https (which we should arguably add, and I don't see the need for HTTPs when we have a signing key).

If this could be done in one install (with only one apt-get update), it would be a lot cleaner and faster.

pmkc: Yes. If it works (after changing the repo to http://)

bsidhom: Ack. I want to see if it still works with https without manually installing the package. In newer Debian releases, that appears to be the case. However, it looks like jessie uses apt < 1.5, so I'm not too hopeful.

bsidhom: @pmkc Sadly, it looks like `add-apt-repository` is not available by default. I assume that means I have to install `software-properties-common` separately, so we need 2 update/install cycles anyway. In that case, I think it's safer to just install all deps defensively.

bsidhom:
After banging my head on this for a while, I've found that Docker forces an HTTPS redirect on apt repos. So we need to install apt-transport-https after all. I've reintroduced the other (preinstalled) dependencies. Dropping add-apt-repository in favor of piping directly into sources.list.d/docker.list.

bsidhom:
Tested that it works after all the rewrites. I'll go ahead and merge.

PR #314, opened 2018-08-07
Add Portable Beam Support

axelmagn:
Add an init action that installs services from the beam portability framework, using flink as its execution engine.

### DIFF
--- a/beam/portable_beam.sh
+++ b/beam/portable_beam.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+readonly LOCAL_JAR_NAME='beam-runners-flink_2.11-job-server.jar'
+readonly RELEASE_SNAPSHOT_URL="http://repo1.maven.org/maven2/org/apache/beam/beam-runners-flink_2.11-job-server/2.6.0/beam-runners-flink_2.11-job-server-2.6.0.jar"
+readonly SERVICE_INSTALL_DIR='/usr/lib/beam-job-service'
+readonly SERVICE_WORKING_DIR='/var/lib/beam-job-service'
+readonly SERVICE_WORKING_USER='yarn'
+
+function is_master() {
+  local role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+  if [[ "$role" == 'Master' ]] ; then
+    true
+  else
+    false
+  fi
+}
+
+function install_job_service() {
+  local master_url="$(/usr/share/google/get_metadata_value attributes/dataproc-master)"
+  mkdir -p "${SERVICE_INSTALL_DIR}"
+  pushd "${SERVICE_INSTALL_DIR}"
+  curl -o "${LOCAL_JAR_NAME}" "${RELEASE_SNAPSHOT_URL}"
+  popd
+  mkdir -p "${SERVICE_WORKING_DIR}"
+  chown -R "${SERVICE_WORKING_USER}" "${SERVICE_WORKING_DIR}"
+  cat > "/etc/systemd/system/beam-job-service.service" <<EOF
+[Unit]
+Description=Beam Job Service
+After=default.target
+
+[Service]
+Type=simple
+User=${SERVICE_WORKING_USER}
+WorkingDirectory=${SERVICE_WORKING_DIR}
+ExecStart=/usr/bin/java -jar ${SERVICE_INSTALL_DIR}/${LOCAL_JAR_NAME} --job-host=${master_url}:8099 --artifacts-dir=hdfs:///tmp/beam-artifacts --flink-master-url=${master_url}:8081
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
+EOF
+systemctl enable beam-job-service
+}
+
+function run_job_service() {
+  systemctl restart beam-job-service
+}
+
+function main() {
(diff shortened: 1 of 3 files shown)


axelmagn:
Ready for review.

R: @bsidhom

beam/portable_beam.sh
```diff
@@ -0,0 +1,90 @@
...
+
+function run_job_service() {
+  systemctl restart beam-job-service
+}
+
+function create_flink_network() {
+  if docker network list --format '{{ .Name }}' | grep -q "^${DOCKER_FLINK_NETWORK_NAME}$"; then
+      echo "Docker network ${DOCKER_FLINK_NETWORK_NAME} already exists."
+  else
+    docker network create \
```
bsidhom: What's the purpose of this bridge network? The Flink runner only uses/supports host network mode as of now anyway.

flink/flink.sh
```diff
@@ -133,11 +133,30 @@ EOF
...
+}
+
 function main() {
 local role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
 if [[ "${role}" == 'Master' ]] ; then
   update_apt_get || err "Unable to update apt-get"
-  apt-get install -y flink || err "Unable to install flink"
+  # TODO: re-enable once flink 1.5 is available via apt
+  # apt-get install -y flink || err "Unable to install flink"
+  install_flink_snapshot || err "Unable to install flink"
```
bsidhom: I don't think we should install this by default for all users of Flink. Please make this configurable, e.g., by GCE metadata.

In addition, please add documentation of the new configuration option on the Flink README _and_ in the Beam README/init action (pointing out whatever version is required).

beam/portable_beam.sh
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+readonly BEAM_VERSION="2.6.0-nightly"
+readonly LOCAL_JAR_NAME='beam-runners-flink_2.11-job-server.jar'
+# TODO: find a way to automatically get latest nightly snapshot
+readonly NIGHTLY_SNAPSHOT_PATH="beam-runners-flink_2.11-job-server/2.6.0-SNAPSHOT/beam-runners-flink_2.11-job-server-2.6.0-20180719.181838-16.jar"
```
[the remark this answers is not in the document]
axelmagn: Actually, 2.6.0 just released so maybe now is the time to switch it over to the maven repo.  I believe that this snapshot can be expected to stick around in perpetuity, but maven is probably the more stable repo.

beam/portable_beam.sh
axelmagn: This must be out of date then.  It originally came from the `docker_flink_network.sh` script, which I thought we were still running.  I'll remove it.

flink/flink.sh
axelmagn: These are all good points.  Fixed.

axelmagn:
I've added all requested changes.  PTAL.

beam/portable_beam.sh
bsidhom: Yes, we started under the assumption that we would use a special network for this, but switched to using host networking (at least for now), so this is no longer needed.

beam/portable_beam.sh
```diff
@@ -0,0 +1,90 @@
...
+  local role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+  if [[ "$role" == 'Master' ]] ; then
+    true
+  else
+    false
+  fi
+}
+
+function snapshot_url() {
+  if [[ "${BEAM_VERSION}" == '2.6.0-nightly' ]] ; then
```
bsidhom: Since we're hard-coding the beam version here, I would just drop this method. We should only keep this if we allow it to be toggled by metadata. I'll leave it up to you as to whether that's worthwhile.

Note that if we decide _not_ to have this (as well as the snapshot repo) be toggled by metadata, we should clean up the file and remove unused references/variables.

beam/portable_beam.sh
axelmagn: removed.

beam/portable_beam.sh
axelmagn: removed and cleaned.

flink/flink.sh
```diff
@@ -133,11 +138,34 @@ EOF
 
 }
 
+function install_flink_snapshot() {
+  local FLINK_TOPLEVEL='flink-1.5.0'
```
bsidhom: Is the 1.5.2 toplevel dir really 1.5.0?

flink/flink.sh
```diff
@@ -133,11 +138,34 @@ EOF
...
+  local FLINK_URL='http://www-eu.apache.org/dist/flink/flink-1.5.2/flink-1.5.2-bin-hadoop28-scala_2.11.tgz'
+  local FLINK_LOCAL='/tmp/flink-1.5.0-bin-hadoop28-scala_2.11.tgz'
```
bsidhom: This should probably be named consistently with the upstream tarball.

flink/flink.sh
axelmagn: good catch.  I must have missed this during testing.  I will update and do another manual sanity check.

axelmagn:
Closing this since bsidhom is away and the branch needs to change significantly.

PR #321, opened 2018-08-13
Support init actions on debian 9

karth295:


datalab/datalab.sh
```diff
@@ -130,6 +130,8 @@ function run_datalab(){
 }
 
 function main(){
+  install_docker
```
pmkc: You could make this not happen if docker is installed so that users can provide their own docker init action.

openssl/README.md
```diff
@@ -1,7 +1,8 @@
 # OpenSSL Upgrade
 
 This [initialization action](https://cloud.google.com/dataproc/init-actions) installs OpenSSL 1.0.2
-from Jessie-backports for Dataproc clusters running Dataproc 1.0 through 1.2.
+from Jessie-backports for Dataproc clusters running Dataproc 1.0 through 1.2
+with debian 8. This init action is unnecessary on debian 9.
```
pmkc: nit: Debian

pmkc:
Can you also change: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/e3e5dda4f1436e92018b31ba705a7e4cf2ce9a42/hue/hue.sh#L49 ?

Rstudio also needs to be updated, but might be more involved.

Issue #320, opened 2018-08-13
Cloud-sql-proxy in HA mode fails on dataproc 1.0 with timeout.


szewi:
Cluster creation in HA mode (3 masters, 2 workers) is failing on dataproc 1.0 on a fresh MySQL instance.

Initialization action stucks on m-1 and m-2 in: 
>beeline -u jdbc:hive2://localhost:10000 -e 'SHOW TABLES;'

This only happens when creating a cluster on dataproc 1.0 `--image-version=1.0` on m-1 and m-2. 

Stuck process looks like this:
>kamil@test-cloud-sql-proxy-ha-1-0-1-m-1:~$ ps aux | grep beeline
root      6224  0.8  0.7 5637336 117000 ?      Sl   13:38   0:01 /usr/lib/jvm/java-8-openjdk-amd64/bin/java -Xmx256m -Xbootclasspath/p:/usr/local/share/google/alpn/alpn-boot-8.1.12.v20180117.jar -Dhadoop.log.dir=/usr/lib/hadoop/logs -Dhadoop.log.file=hadoop.log -Dhadoop.home.dir=/usr/lib/hadoop -Dhadoop.id.str= -Dhadoop.root.logger=INFO,console -Djava.library.path=/usr/lib/hadoop/lib/native -Dhadoop.policy.file=hadoop-policy.xml -Djava.net.preferIPv4Stack=true -Xmx3768m -Dlog4j.configuration=beeline-log4j.properties -Dhadoop.security.logger=INFO,NullAppender org.apache.hadoop.util.RunJar /usr/lib/hive/lib/hive-beeline-1.2.1.jar org.apache.hive.beeline.BeeLine -u jdbc:hive2://localhost:10000 -e SHOW TABLES;

If kill those processes manually on m-1 and m-2. The init action runs another validation attempt and whole cluster creation passes. 

Notes:
1. On other dataproc versions [run with retry](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/commit/a1956361d932303f2350ff799c8e8326f047daa8#diff-276614f956b43fcf03609259658f80feR255) is working fine. Beeline is not hanging and validation runs another attempt automatically.

Steps to reproduce:
`gcloud sql instances create db_nam --gce-zone zone`
[...]

PR #323, opened 2018-08-14
Fixes hanging beeline on dataproc version 1.0 #320.

szewi:
Using timeout command works fine on all dataproc versions. Fixes validation part that uses older beeline version available on dataproc 1.0.


szewi:
@karth295 @memoryoflife could you take a look?
Please see #320 for details.

PR #325, opened 2018-08-14
Fix for Hue initialization on HA cluster

### LINKED ISSUE #324: Hue does not initialize on HA clusters
Init action cannot be successfully completed on HA configuration (does not matter on which dataproc version).

I found that's because of missing mysql service on additional master nodes - I found it is installed only on m-0. 

I will submit PR with required changes. In PR I will fix also problem with failure on database operation using mysql for additional master nodes. Before script starts doing action on database, mysql restart operation should be performed.

```
+ mysql -u root -proot-password -e '     CREATE DATABASE hue;     CREATE USER '\''hue'\''@'\''localhost'\'' IDENTIFIED BY '\''hue-password'\'';     GRANT ALL PRIVILEGES ON hue.* TO '\''hue'\''@'\''localhost'\'';'
ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)
+ err 'Unable to create database'
```

It is already fixed in #325 .

adam-stasiak:
I added step for installation mysql service on additional master nodes if needed. I also moved mysql restart operation earlier to avoid problem with database creation on dataproc 1.3 image. 
This PR fixed #324

### DIFF
--- a/hue/hue.sh
+++ b/hue/hue.sh
@@ -166,15 +166,12 @@ EOF
 }
 
 function main() {
-  # Determine the role of this node
-  local role=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+  # Determine if script is launched on current master node (not additional)
+  local master_node=$(/usr/share/google/get_metadata_value attributes/dataproc-master)
+  local node_name=${HOSTNAME}
   # Only run on the master node of the cluster
-  if [[ "${role}" == 'Master' ]]; then
+  if [[ "${HOSTNAME}" == "${master_node}" ]]; then
     update_apt_get || err "Unable to update apt-get"
-    if [[ $(which mysql) == '' ]]; then
-      DEBIAN_FRONTEND=noninteractive apt-get -y install mysql-server \
-        || err "Unable to install mysql-server"
-    fi
     install_hue_and_configure || err "Hue install process failed"
   else
     return 0 || err "Hue can be installed only on master node - skipped for worker node"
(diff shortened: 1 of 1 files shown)


karth295:
Rather than running a separate Hue instance on every master node, a simpler fix is just to run it on `-m-0`. That's what I did in `jupyter2`, for example.

adam-stasiak:
@karth295 applied your suggestion -  now I check which node is a "main" master (not additional one) and then compare hostname to that one.

adam-stasiak:
Thx @karth295 for the response. I tested init action on those configs and all succeeded. Could you give me more details? Does init action fail for you?

Port is set based on   ```hdfs getconf -confKey dfs.namenode.http-address```:
```
local new_hdfs_url='webhdfs_url\=http:\/\/'"$(hdfs getconf -confKey dfs.namenode.http-address)" 
```
and documentation says it is ```

The address and the base port where the dfs namenode web ui will listen on.
```. I tried this running on various configurations (image ver.) and returns a value that seems to be ok – for HA 1.1 it is 50070 as you said but for 1.2 and 1.3 HA it is 9870.

The second issue - I will look on this now.

karth295:
The init action succeeded, I just saw those errors when I opened the web ui.

adam-stasiak:
Thanks! I will look on this.

medb:
This PR is obsolete, Hue init action muse have been fixed some time ago.

PR #331, opened 2018-08-23
Add retry for apt-get install and log the process holding the apt lock.

functicons:


functicons:
I've tested these changes by creating clusters with the init actions.

functicons:
Adding an example error message: 

+ apt-get install -y hbase
E: Could not get lock /var/lib/dpkg/lock - open (11: Resource temporarily unavailable)
E: Unable to lock the administration directory (/var/lib/dpkg/), is another process using it?
+ err 'Unable to install hbase.'

medb:
Root cause of the issue is apt upgrade that Debian 9 runs during start up:
https://github.com/QubesOS/qubes-issues/issues/2621#issuecomment-277578264

PR #342, opened 2018-09-05
Configure Docker to use GCR during initialization

axelmagn:
It is reasonable to expect that Docker users on Dataproc will want it
to automatically integrate with the Google Container Registry.  This
change configures Docker to use GCR as part of the initialization
action.

This change is also useful for Beam users who are building
their own container images and storing them on GCR.

### DIFF
--- a/docker/docker.sh
+++ b/docker/docker.sh
@@ -4,6 +4,8 @@ set -euxo pipefail
 
 # TODO: Allow this to be configured by metadata.
 readonly DOCKER_VERSION='18.06.0~ce~3-0~debian'
+readonly CREDENTIAL_HELPER_VERSION='1.5.0'
+
 
 function is_master() {
   local role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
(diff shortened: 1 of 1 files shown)


docker/docker.sh
```diff
@@ -37,9 +37,26 @@ function install_docker() {
...
+  OS=linux
+  ARCH=amd64
+
+  # this standalone method is recommended here:
+  # https://cloud.google.com/container-registry/docs/advanced-authentication#standalone_docker_credential_helper
+  curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v${VERSION}/docker-credential-gcr_${OS}_${ARCH}-${VERSION}.tar.gz" \
+    | tar xz --to-stdout ./docker-credential-gcr \
+    > /usr/bin/docker-credential-gcr && chmod +x /usr/bin/docker-credential-gcr
+  docker-credential-gcr configure-docker
+  su yarn --command "docker-credential-gcr configure-docker"
```
pmkc: I don't see why to run this as YARN. My (limited) understanding is this makes all Docker users able to read from GCR.

Does that not work?

docker/docker.sh
```diff
@@ -37,9 +37,26 @@ function install_docker() {
...
+function configure_gcr() {
+  VERSION=1.5.0
+  OS=linux
+  ARCH=amd64
+
+  # this standalone method is recommended here:
+  # https://cloud.google.com/container-registry/docs/advanced-authentication#standalone_docker_credential_helper
+  curl -fsSL "https://github.com/GoogleCloudPlatform/docker-credential-gcr/releases/download/v${VERSION}/docker-credential-gcr_${OS}_${ARCH}-${VERSION}.tar.gz" \
+    | tar xz --to-stdout ./docker-credential-gcr \
+    > /usr/bin/docker-credential-gcr && chmod +x /usr/bin/docker-credential-gcr
```
pmkc: Use /usr/local/bin because we are not installing from a debian package

docker/docker.sh
```diff
@@ -37,9 +37,26 @@ function install_docker() {
   apt-get install -y docker-ce="${DOCKER_VERSION}"
 }
 
+function configure_gcr() {
+  VERSION=1.5.0
+  OS=linux
+  ARCH=amd64
```
pmkc: Might as well hard code OS an ARCH inline. They are set in stone.

docker/docker.sh
```diff
@@ -37,9 +37,26 @@ function install_docker() {
   apt-get install -y docker-ce="${DOCKER_VERSION}"
 }
 
+function configure_gcr() {
+  VERSION=1.5.0
```
pmkc: Make this more descriptive at put it at the top of the file

docker/docker.sh
```diff
@@ -37,9 +37,26 @@ function install_docker() {
...
+    | tar xz --to-stdout ./docker-credential-gcr \
+    > /usr/bin/docker-credential-gcr && chmod +x /usr/bin/docker-credential-gcr
+  docker-credential-gcr configure-docker
+  su yarn --command "docker-credential-gcr configure-docker"
+}
+
 function configure_docker() {
   # The installation package should create `docker` group.
   usermod -aG docker yarn
+  # configure docker to use gcr
```
pmkc: nit: Google Container Registry. Be explicit in at least one comment.

docker/docker.sh
axelmagn: You would think so, right?  I actually went back to double check this when you asked, because that was my intuition as well.  However, this command configures docker on a per-user basis.  Unfortunately I have not been able to find any way to configure docker to use GCR globally.  Root and yarn are the two static users that I could see using docker, and I run as yarn since the configuration function seems to indicate that yarn is the only other user added to the docker group.  I will add a comment that makes this explicit.

axelmagn: A viable future change may be to let the user pass docker-enabled users via a metadata variable, and generalize beyond yarn this way.

PR #344, opened 2018-09-12
Add Beam Job Service init action

axelmagn:
This sets up a Beam Job Service on the master node.  It depends on Flink
and Docker being installed to function correctly.

The Beam team has advised us that Beam is not in a place where the portability
framework can be considered production ready, and as a result the stable
snapshots cannot be relied on.  They advise instead that for the time being,
users build their own Job Service and Worker Image artifacts.  This PR includes
both an init action for the job service, and an automated build script.
Build instructions are also available in the README.

### DIFF
--- a/beam/README.md
+++ b/beam/README.md
@@ -1,39 +1,31 @@
 # Beam Initialization Actions
 
-While Beam is primarily a SDK, jobs running under its portability framework
+While Apache Beam is primarily a SDK, jobs running under its portability framework
 require a job service to run correctly.  This directory contains a setup script
-to setup a properly configured beam job services.
+to properly configure beam job services.
 
 **WARNING:** The Beam portability framework is **under active development** and
 should not be used in a production context.
 
-Due to the current development status of Beam's portability framework, users are
-responsible for building and maintaining their own Beam artifacts manually.
-Instructions are included below.  The current development status of the Beam
-portability framework can be found
-[here](https://beam.apache.org/contribute/portability/#status).
+Due to the current development [status](https://beam.apache.org/contribute/portability/#status) of Beam's portability
+framework, users are responsible for building and maintaining their own Beam artifacts manually. Instructions are included below.
 
 ## Building Beam Artifacts
 
-There are two categories of artifacts that will need to be generated for this
-initialization action:
+You will generate two categories of artifacts for this initialization action:
 
-| Job Service | This microservice runs on the master node, accepting new beam jobs on port 8099.  It is configured to submit Beam jobs to Flink. |
-| Worker Container Images | These docker images are what run language-specific code on worker nodes. |
+| Job Service | This micro service runs on the master node, accepting new beam jobs on port 8099.  It is configured to submit Beam jobs to Apache Flink. |
+| Worker Container Images | These docker images run language-specific code on worker nodes. |
 
-Throughout the build process, we will refer to the following terms, which will
-need to be substituted into build commands if building manually.  In bash, one
-convenient way to use them is to set environment variables using `export
-<term>=<value>`.
+When building manually, substitute the following terms into build commands. In bash, set environment variables using `export <term>=<value>`.
 
-| `BEAM_JOB_SERVICE_DESTINATION` | A GCS directory path accessible by both the build machine and the cluster to be created. |
+| `BEAM_JOB_SERVICE_DESTINATION` | A Cloud Storage directory path accessible by both the build machine and the cluster to be created. |
 | `BEAM_CONTAINER_IMAGE_DESTINATION` | A Docker repository path prefix accessible by both the build machine and the cluster to be created. |
 | `BEAM_SOURCE_VERSION` | A tag, branch, or commit hash in the Beam source repositories to build artifacts from. (default: `master`) |
 
 ### Automated Build
 
-The `util` directory contains a script to assist in building Beam artifacts.
-From this directory on a build machine, you can invoke it with:
+You can invoke a helper script from `util` directory to build Beam artifacts.
 
 ```bash
 bash ./util/build-beam-artifacts.sh <BEAM_JOB_SERVICE_DESTINATION> <BEAM_CONTAINER_IMAGE_DESTINATION> [<BEAM_SOURCE_VERSION>]
@@ -41,8 +33,7 @@ bash ./util/build-beam-artifacts.sh <BEAM_JOB_SERVICE_DESTINATION> <BEAM_CONTAIN
 
 ### Manual Build
 
-Alternatively, developers can build the necessary artifacts manually.  To get
-started, first clone the beam source code into a working directory.
+To get started, clone the beam source code into a working directory.
 
 ```bash
 git clone https://github.com/apache/beam.git
@@ -52,15 +43,13 @@ git checkout ${BEAM_SOURCE_VERSION}
 
 #### Build the Job Service
 
-This step will build a standalone job service jar, and upload it to that path
-for use during cluster creation.
+Next, build a standalone job service jar.
 
 ```bash
 ./gradlew :beam-runners-flink_2.11-job-server:shadowJar
 ```
 
-After the jar has been built, it will need to be uploaded to a GCS path that
-clusters can access during initialization.
+Then, upload the jar to a Cloud Storage path that clusters can access during initialization.
 
 ```bash
 gsutil cp \
@@ -70,16 +59,17 @@ gsutil cp \
 
 #### Build the Worker Container Images
 
-This step will build the docker images used by Beam worker tasks.
+Build the docker images used by Beam worker tasks.
 
 ```bash
 ./gradlew docker
 ```
 
-This will build a number of docker images with names that look like
-`<USER>-docker-apache.bintray.io/beam/<LANGUAGE>`.  These need to be renamed and
-pushed to a docker repository path that clusters can access during
-initialization.  We recommend tagging these images with the
+Docker images names are generated in the following format:
+`<USER>-docker-apache.bintray.io/beam/<LANGUAGE>`. 
+
+Rename and push the images to a docker repository path that clusters can access during
+initialization.  As a best practice, tag the images with the
 `BEAM_SOURCE_VERSION` they were generated from.
 
 ```bash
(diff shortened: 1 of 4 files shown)


beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+}
+
+function pull_beam_images() {
+  local beam_image_version="$(/usr/share/google/get_metadata_value \
+    "attributes/${BEAM_IMAGE_VERSION_METADATA_KEY}" \
+    || echo "${BEAM_IMAGE_VERSION_DEFAULT}")"
+  local image_repo="$(/usr/share/google/get_metadata_value \
+    "attributes/${BEAM_IMAGE_REPOSITORY_KEY}" \
+    || echo "${BEAM_IMAGE_REPOSITORY_DEFAULT}")"
+  sudo -u yarn -i docker pull "${image_repo}/go:${beam_image_version}"
```
bsidhom: Please add a comment as to why the `-i` flag is necessary (presumably to set `HOME` etc. correctly). In most cases, this would _not_ be desired due to inheritance of `.profile` and other configuration files.

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+  sudo -u yarn -i docker pull "${image_repo}/python:${beam_image_version}"
+  sudo -u yarn -i docker pull "${image_repo}/java:${beam_image_version}"
+}
+
+function main() {
+  if [[ is_master ]]; then
+    install_job_service
+    run_job_service
```
bsidhom: Do we only want the job server to run on the primary master in HA clusters? We've been bitten by this before. The job server should technically not request any YARN resources until jobs are submitted, though this may cause problems in other ways.

Note that I think it's still worth _installing_ the job server on all masters in either case.

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+  local master_url="$(/usr/share/google/get_metadata_value attributes/dataproc-master)"
+  local artifacts_dir="$(get_artifacts_dir)"
+  local release_snapshot_url="$(/usr/share/google/get_metadata_value \
+    "attributes/${RELEASE_SNAPSHOT_URL_METADATA_KEY}" \
+    || echo "${RELEASE_SNAPSHOT_URL_DEFAULT}")"
+
+  echo "Retrieving Beam Job Service snapshot from ${release_snapshot_url}"
+
+  local flink_master="$(flink_master_url)"
+  echo "resolved flink master to: '${master_url}'"
```
bsidhom: Nit: capitalize

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+      | head -n1 \
+      | awk -F $'\t' '{print $9}' \
+      | cut -c8-
+  else
+    echo "localhost:8081"
+  fi
+}
+
+function install_job_service() {
+  local master_url="$(/usr/share/google/get_metadata_value attributes/dataproc-master)"
```
bsidhom: I believe this will resolve to the _primary_ master, but we should verify this.

@mikaylakonst Do you know if this is the primary master, "current" master, or _all_ masters?

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+  if [[ "$role" == 'Master' ]] ; then
+    true
+  else
+    false
+  fi
+}
+
+function get_artifacts_dir() {
+  /usr/share/google/get_metadata_value "attributes/${ARTIFACTS_DIR_METADATA_KEY}" \
+    || echo "gs://$(/usr/share/google/get_metadata_value "attributes/dataproc-bucket")/beam-artifacts"
```
bsidhom: If I'm not mistaken, `attributes/dataproc-bucket` points to the Dataproc _staging_ bucket for this cluster rather than the Dataproc init action bucket (which we should hard code if we want it).

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+readonly LOCAL_JAR_NAME='beam-runners-flink_2.11-job-server.jar'
+readonly SERVICE_INSTALL_DIR='/usr/lib/beam-job-service'
+readonly SERVICE_WORKING_DIR='/var/lib/beam-job-service'
+readonly SERVICE_WORKING_USER='yarn'
+
+readonly ARTIFACTS_DIR_METADATA_KEY='beam-artifacts-dir'
```
bsidhom: Please rename this variable (and related variables) to make it clear this should point to a GCS directory.

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+
+function flink_master_url() {
+  local start_flink_yarn_session="$(/usr/share/google/get_metadata_value \
+    "attributes/${START_FLINK_YARN_SESSION_METADATA_KEY}" \
+    || echo "${START_FLINK_YARN_SESSION_DEFAULT}")"
+  # TODO: delete this workaround when the beam job service is able to understand 
+  # flink in yarn mode.
+  if ${start_flink_yarn_session} ; then
+    # grab final field from the first yarn application that contains 'flink'
+    yarn application -list >&2 # Logging this here is useful for debugging potential failures
```
bsidhom: If you save this output into a variable, then `set -x` above will take care of logging the expansion.

Then you can use `echo $variable | grep ...` below to reuse the output.

Attempting to use the above log for debugging is less useful because the first list and filtering are not done atomically.

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+    yarn application -list \
+      | grep -i 'flink' \
+      | head -n1 \
+      | awk -F $'\t' '{print $9}' \
+      | cut -c8-
```
bsidhom: By the way, does this work for resolving the job server endpoint? I'm not sure how we get the job server out of the application list.

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+readonly LOCAL_JAR_NAME='beam-runners-flink_2.11-job-server.jar'
+readonly SERVICE_INSTALL_DIR='/usr/lib/beam-job-service'
+readonly SERVICE_WORKING_DIR='/var/lib/beam-job-service'
```
bsidhom: Are there any situations where the install and working dir are not the same? We should probably merge these, especially since creating/modifying them separately below without anticipating them being the same is error prone.

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+Description=Beam Job Service
+After=default.target
+
+[Service]
+Type=simple
+User=${SERVICE_WORKING_USER}
+WorkingDirectory=${SERVICE_WORKING_DIR}
+ExecStart=/usr/bin/java \
+  -jar ${SERVICE_INSTALL_DIR}/${LOCAL_JAR_NAME} \
+  --job-host=${master_url}\
```
bsidhom: Hmm. `master_url` here is the host that we want to listen on in the Beam job service rather than the Flink master. I assumed above that we were extracting the Flink master to listen on above. Does this work correctly?

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+  --artifacts-dir=${artifacts_dir} \
+  --flink-master-url=${flink_master}
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
+EOF
+systemctl enable beam-job-service
```
bsidhom: Note my comment below as to whether we want to run the job service on secondary masters in HA mode. We should disable this in those cases if we do not want the service to run.

beam/README.md
```diff
@@ -0,0 +1,169 @@
...
+| `BEAM_CONTAINER_IMAGE_DESTINATION` | A Docker repository path prefix accessible by both the build machine and the cluster to be created. |
+| `BEAM_SOURCE_VERSION` | A tag, branch, or commit hash in the Beam source repositories to build artifacts from. (default: `master`) |
+
+### Automated Build
+
+The `util` directory contains a script to assist in building Beam artifacts.
+From this directory on a build machine, you can invoke it with:
+
+```bash
+bash ./util/build-beam-artifacts.sh <BEAM_JOB_SERVICE_DESTINATION> <BEAM_CONTAINER_IMAGE_DESTINATION> [<BEAM_SOURCE_VERSION>]
```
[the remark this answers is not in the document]
axelmagn: My attempt to do so were on lines 29-30, which I believe document the constraints you mention.  Marking as resolved, but please re-open if you have a more specific change in mind.

beam/README.md
```diff
@@ -0,0 +1,169 @@
...
+| flink-start-yarn-session | `true` |
+| flink-snapshot-url | URL to a flink 1.5 snapshot such as [this one](https://archive.apache.org/dist/flink/flink-1.5.3/flink-1.5.3-bin-hadoop28-scala_2.11.tgz) |
+
+Beam 2.7 requires Flink 1.5, which is not the default Flink version in Dataproc
+1.2.  Therefore users must specify a custom flink snapshot.
+
+In practice, the best way to handle this many metadata values is through a
+script such as this one:
+
+TODO: anonymize this
```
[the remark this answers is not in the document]
axelmagn: Whoops.  will delete.

beam/README.md
```diff
@@ -0,0 +1,169 @@
...
+  --runner flink \
+  --endpoint localhost:8099 \
+  --experiments beam_fn_api \
+  --output=<out> \
+  --container_image <BEAM_CONTAINER_DESTINATION>/go:<BEAM_SOURCE_VERSION>
+```
+
+The Beam Job Service port will need to be opened if users wish to submit beam
+jobs from machines outside the cluster.  Please refer to the
+[guide](https://cloud.google.com/vpc/docs/using-firewalls) for instructions on
```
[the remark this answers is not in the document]
axelmagn: This is a really good point that I had overlooked.  I'll remove this for now, since we don't have a validated answer yet.

beam/README.md
```diff
@@ -0,0 +1,169 @@
...
+FLINK_SNAPSHOT="https://archive.apache.org/dist/flink/flink-1.5.3/flink-1.5.3-bin-hadoop28-scala_2.11.tgz"
+METADATA="beam-job-service-snapshot=<...>"
+METADATA+=",beam-image-enable-pull=true"
+METADATA+=",beam-image-repository=<...>"
+METADATA+=",beam-image-version=latest"
+METADATA+=",flink-start-yarn-session=true"
+METADATA+=",flink-snapshot-url=${FLINK_SNAPSHOT}"
+
+gcloud dataproc clusters create "${CLUSTER_NAME}" \
+  --initialization-actions="${INIT_ACTIONS}" \
```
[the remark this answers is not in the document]
axelmagn: I did this as a stylistic choice because expanding them out in the command line resulted in a very long line.  I

> Since the init actions are all usable as-is with metadata substitution, most people will actually use the init actions we host in our own bucket.

I'm not sure I quite follow the point you're making, but if I understand correctly then you're suggesting I do away with the `INIT_BUCKET` variable in favor of just substituting it into lines 121-123.  I'm fine with this, but please confirm that this is what you meant.

beam/beam_job_service.sh
axelmagn: They are not currently the same.  The install dir is in `/usr` while the working dir is in `/var`.

beam/beam_job_service.sh
axelmagn: Yes, this is correct.  My intention was to use the Dataproc staging bucket as the default artifacts bucket if none is supplied by the user.  The alternative is to make this metadata key required at creation time, so that we have somewhere user-created to store the artifacts.  I don't think we would want to store customer artifacts in the dataproc init action bucket.

It was sort of an independent design choice, but the staging bucket seemed like a good place to put them if no artifact directory was specified.  It also helps with the cleanup of integration tests.

beam/beam_job_service.sh
axelmagn: Total hack, but yes.  It turns out that the job server is the same port as the web UI provided in the tracking URL.  This can be deprecated one day when the job service learns to use flink's autodetect features.

beam/beam_job_service.sh
axelmagn: I'm not too concerned with HA mode for the moment.  I've been assuming that we only have one master.

beam/beam_job_service.sh
axelmagn: Correct.  `master_url` refers to the machine the beam job service can be reached on (localhost won't cut it).  `flink_master` refers to the flink job master.

beam/beam_job_service.sh
axelmagn: I wish I had documented the conversation (I think it was with you), but at one point we decided that HA clusters were not in scope for this first pass, since our flink init action does not set HA mode at the moment.  I view HA as something that could be added as a separate feature request when and if we need it, but I am open to arguments to the contrary.

beam/beam_job_service.sh
axelmagn: Ack.  See my response below.

beam/beam_job_service.sh
axelmagn: Ack. will do.  TL;DR: GCR authorization is weird, and happens on a per-user basis.

beam/beam_job_service.sh
```diff
@@ -0,0 +1,140 @@
...
+
+function main() {
+  if [[ is_master ]]; then
+    install_job_service
+    run_job_service
+  fi
+
+  local pull_images="$(/usr/share/google/get_metadata_value \
+    "attributes/${BEAM_IMAGE_ENABLE_PULL_METADATA_KEY}" \
+    || echo "${RELEASE_SNAPSHOT_URL_METADATA_KEY}")"
```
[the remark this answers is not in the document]
axelmagn: ATM, it's tough to test without image prefetch since all of my end-to-end tests require it (thus motivating this function in the first place).  Good catch though.  will fix.  Let me know if you have ideas on how to get it working without prefetch, since I was unable to.

beam/util/build-beam-artifacts.sh
```diff
@@ -0,0 +1,45 @@
...
+    local image_destination="${BEAM_CONTAINER_IMAGE_DESTINATION}/$(basename ${image}):${BEAM_SOURCE_VERSION}"
+    docker tag $image:latest ${image_destination}
+    docker push ${image_destination}
+  done
+}
+
+function main() {
+  local workdir=$(mktemp -d)
+  pushd ${workdir}
+  git clone https://github.com/apache/beam.git
```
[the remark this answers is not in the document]
axelmagn: Good point.  I'll add it as an option.

beam/README.md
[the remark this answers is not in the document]
bsidhom: I meant to fully expand it in-line because it's generally easier to copy-paste a single command rather than a mini-script that has to set up state/variables first. However, the command line is long enough that it's fine to leave the snippet as is.

beam/beam_job_service.sh
bsidhom: Sorry, not "Dataproc" init action bucket, but the bucket associated with this init action's artifacts. Seems reasonable to keep it like this for now, though I'm not sure how/when the staging buckets are cleaned up if created by Dataproc.

beam/beam_job_service.sh
bsidhom: Sounds fine for now. If that's the case, please add an explicit comment in the readme.

axelmagn:
All requested changes added.

beam/README.md
```diff
@@ -22,7 +24,7 @@ You will generate two categories of artifacts for this initialization action:
 When building manually, substitute the following terms into build commands. In
 bash, set environment variables using `export <term>=<value>`.
 
-| `BEAM_JOB_SERVICE_DESTINATION` | A Cloud Storage directory path accessible by both the build machine and the cluster to be created. |
+| `beam_DESTINATION` | A Cloud Storage directory path accessible by both the build machine and the cluster to be created. |
```
bsidhom: `beam_DESTINATION` -> `BEAM_DESTINATION`

Looks like this is used in several places; I assume this was just a bad substitution.

axelmagn: Yep.  Good catch.  Thanks.  Fixed.

axelmagn:
Good catch on the bad substitution. Fixed.

PR #350, opened 2018-09-21
Update secure/README.md

cyxxy:
Update secure/README.md to make it clearer cross-realm trust parameters are optional.

### DIFF
--- a/secure/README.md
+++ b/secure/README.md
@@ -19,14 +19,7 @@ gcloud dataproc clusters create <CLUSTER_NAME> \
     --metadata "root-password-uri=gs://<SECRET_BUCKET>/root-password.encrypted" \
     --metadata "keystore-password-uri=gs://<SECRET_BUCKET>/keystore-password.encrypted"
 ```
-Optionally, enable cross-realm trust: 
-```bash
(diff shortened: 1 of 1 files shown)


secure/README.md
```diff
@@ -17,13 +17,15 @@ gcloud dataproc clusters create <CLUSTER_NAME> \
...
+```
+Optionally, enable cross-realm trust: 
+```bash
     --metadata "cross-realm-trust-realm=<REMOTE.REALM>" \
     --metadata "cross-realm-trust-kdc=<REMOTE_KDC>" \
     --metadata "cross-realm-trust-admin-server=<REMOTE_ADMIN_SERVER>" \
     --metadata "cross-realm-trust-password-uri=gs://<SECRET_BUCKET>/cross-realm-trust-password.encrypted"
 ```
-
 All the metadata key-value pairs are required to properly secure the cluster.
```
cobb: This comment now seems to suggest that cross-realm trust is required to properly secure the cluster, which I don't think is true. Maybe just move this comment before the optional section begins?

cyxxy: Good catch. I did an reordering and split of the explanation of the parameters.

PR #352, opened 2018-09-25
Provide default values (and self-signed certificate) if user does not specify them in secure/secure.sh.

cyxxy:
Provide as many default values as possible.

### DIFF
--- a/secure/secure.sh
+++ b/secure/secure.sh
@@ -512,7 +512,8 @@ function copy_or_create_keystore_files() {
       # The validity of the cert is 1800 days.
       keytool -genkeypair -keystore keystore.jks -keyalg RSA -keysize 2048 \
         -storepass "${keystore_password}" -keypass "${keystore_password}" -validity 1800 -alias dataproc-cert \
-        -dname "CN=*.${DOMAIN}, OU=Google Cloud Platform, O=Google, L=Mountain View, S=CA, C=US"
+        -dname "CN=*.${DOMAIN}"
+        #-dname "CN=*.${DOMAIN}, OU=Google Cloud Platform, O=Google, L=Mountain View, S=CA, C=US"
       keytool -export -alias dataproc-cert -storepass "${keystore_password}" -file dataproc.cert -keystore keystore.jks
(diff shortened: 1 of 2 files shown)


secure/secure.sh
```diff
@@ -470,18 +492,50 @@ function config_mapred_site() {
...
+    gsutil cp "${keystore_uri}" "${HADOOP_CONF_DIR}"/ssl/keystore.jks
+    gsutil cp "${truststore_uri}" "${HADOOP_CONF_DIR}"/ssl/truststore.jks
+  else
+    local keystore_location="${staging_meta_info_gcs_prefix}/keystores"
+    if [[ "${FQDN}" == "${MASTER_FQDN}" ]]; then
+      echo "Generating self-signed certificate from ${FQDN}"
+      # The validity of the cert is 1800 days.
+      keytool -genkeypair -keystore keystore.jks -keyalg RSA -keysize 2048 \
+        -storepass "${keystore_password}" -keypass "${keystore_password}" -validity 1800 -alias dataproc-cert \
+        -dname "CN=*.${DOMAIN}, OU=Google Cloud Platform, O=Google, L=Mountain View, S=CA, C=US"
```
karth295: Let's avoid using Google, etc. If these are required fields, let's use fake values.

secure/secure.sh
```diff
@@ -470,18 +492,50 @@ function config_mapred_site() {
...
+          gsutil cp "gs://${keystore_location}/keystore.jks" "${HADOOP_CONF_DIR}"/ssl/keystore.jks
+          gsutil cp "gs://${keystore_location}/truststore.jks" "${HADOOP_CONF_DIR}"/ssl/truststore.jks
+          files_downloaded=1
+          break
+        fi
+        sleep 1
+      done
+
+      if [[ "${files_downloaded}" == 0 ]]; then
+        error 'Failed to download keystore and truststore files after 5 minutes. Something went wrong.'
```
karth295: Nit: Remove "Something went wrong."

secure/secure.sh
cyxxy: removed Google, GCP, etc. Only kept the wildcard CNAME.

secure/secure.sh
cyxxy: Done.

PR #355, opened 2018-09-27
Test scripts and prow config for init actions

ishanrc:
These are the Polidia authored init action tests and configuration to run them on the prow cluster at http://35.241.19.33/. 
Here's an example PR where prow has run the tests:
https://github.com/ishanrc/dataproc-initialization-actions/pull/32
And one from a user who is not an org member where prow is waiting for authorization:
https://github.com/ishanrc/dataproc-initialization-actions/pull/33

### DIFF
--- a/testing-scripts/bigtable/test_bigtable.py
+++ b/testing-scripts/bigtable/test_bigtable.py
@@ -1,0 +1,104 @@
+"""
+This module provides testing functionality of the BigTable Init Action.
+
+Test logic:
+1. Create test table and fill it with some data by injecting commands into hbase shell.
+2. Validate from local station that BigTable has test table created with right number of rows.
+
+Note:
+    Test REQUIRES cbt tool installed which provides CLI access to BigTable instances.
+    See: https://cloud.google.com/bigtable/docs/cbt-overview
+"""
+import unittest
+import random
+
+from parameterized import parameterized
+
+from dataproc_test_case import DataprocTestCase
+
+
+class BigTableTestCase(DataprocTestCase):
+    COMPONENT = 'bigtable'
+    INIT_ACTION = 'gs://dataproc-initialization-actions/bigtable/bigtable.sh'
+    TEST_SCRIPT_FILE_NAME = "run_hbase_commands.py"
+    METADATA = None
+    DB_NAME = None
+    ZONE = None
+
+    def setUp(self):
+        super().setUp()
+        self.DB_NAME = "test-{}-db".format(random.randint(1, 10000))
+        _, zone, _ = self.run_command("gcloud config get-value compute/zone")
+        self.ZONE = zone.strip()
+        _, project, _ = self.run_command("gcloud config get-value project")
+        project = project.strip()
+        self.METADATA = "bigtable-instance={},bigtable-project={}"\
+            .format(self.DB_NAME, project)
+
+        ret_code, stdout, stderr = self.run_command(
+            'gcloud beta bigtable instances create {} --cluster {} --cluster-zone {} '
+            '--display-name={} --instance-type=DEVELOPMENT'
+            .format(self.DB_NAME, self.DB_NAME, self.ZONE, self.DB_NAME)
+        )
+        self.assertEqual(ret_code, 0, "Failed to create bigtable instance {}. Last error: {}"
+                         .format(self.DB_NAME, stderr))
+
+    def tearDown(self):
+        super().tearDown()
+        ret_code, stdout, stderr = self.run_command(
+            'gcloud beta bigtable instances delete {}'.format(self.DB_NAME)
+        )
+        self.assertEqual(ret_code, 0, "Failed to delete bigtable instance {}. Last error: {}"
+                         .format(self.DB_NAME, stderr))
+
+    def _run_hbase_shell(self, name):
+        ret_code, stdout, stderr = self.ssh_cmd(name,
+        '"python {}"'.format(self.TEST_SCRIPT_FILE_NAME))
+        self.assertEqual(ret_code, 0, "Failed to validate cluster. Last error: {}".format(stderr))
+
+    def _validate_bigtable(self):
+        ret_code, stdout, stderr = self.run_command(
+            'cbt -instance {} read test-bigtable '.format(
+                self.DB_NAME
+            )
+        )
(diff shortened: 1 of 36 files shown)


testing-scripts/sqoop/test_sqoop.py
```diff
@@ -0,0 +1,102 @@
...
+    @parameterized.expand([
+        # ("SINGLE", "1.0"),
+        # ("SINGLE", "1.1"),
+        ("SINGLE", "1.2", ["m"]),
+        # ("STANDARD", "1.0"),
+        # ("STANDARD", "1.1"),
+        ("STANDARD", "1.2", ["m"]),
+        # ("HA", "1.0"),
+        # ("HA", "1.1"),
+        ("HA", "1.2", ["m"]),
```
karth295: Is this supposed to be -m-0?

testing-scripts/sqoop/test_sqoop.py
```diff
@@ -0,0 +1,102 @@
...
+
+import unittest
+import random
+
+from parameterized import parameterized
+
+from dataproc_test_case import DataprocTestCase
+
+
+class SqoopTestCase(DataprocTestCase):
```
karth295: Let's leave this one out, and instead add it to #269 .

testing-scripts/bigtable/test_bigtable.py
```diff
@@ -0,0 +1,104 @@
...
+import random
+
+from parameterized import parameterized
+
+from dataproc_test_case import DataprocTestCase
+
+
+class BigTableTestCase(DataprocTestCase):
+    COMPONENT = 'bigtable'
+    INIT_ACTION = 'gs://dataproc-initialization-actions/bigtable/bigtable.sh'
```
karth295: All of these tests will eventually need to run against PRs (code that hasn't been checked in and synced to the GCS bucket). I think it makes sense for the tests to copy the scripts from their checkout (`../bigtable/bigtable.sh`) into a staging GCS bucket and use that

testing-scripts/bigtable/test_bigtable.py
```diff
@@ -0,0 +1,104 @@
...
+        self.assertIn("value4", stdout, "Failed to validate cluster. value4 is missing in the table")
+
+    def verify_instance(self, name):
+        self.upload_test_file(name)
+        self._run_hbase_shell(name)
+        self._validate_bigtable()
+
+
+    @parameterized.expand([
+        # ("SINGLE", "1.0", ["m"]),
```
karth295: Nit: this (and other commented out parameterizations) should have a reason that they're commented out. E.g. the init action is not supported on 1.0.

testing-scripts/bigtable/test_bigtable.py
```diff
@@ -0,0 +1,104 @@
...
+    COMPONENT = 'bigtable'
+    INIT_ACTION = 'gs://dataproc-initialization-actions/bigtable/bigtable.sh'
+    TEST_SCRIPT_FILE_NAME = "run_hbase_commands.py"
+    METADATA = None
+    DB_NAME = None
+    ZONE = None
+
+    def setUp(self):
+        super().setUp()
+        self.DB_NAME = "test-{}-db".format(random.randint(1, 10000))
```
karth295: If you `Ctrl-C` a test, you will end up leaking Bigtable instances. It would be nice to use a consistent name, and call tearDown in setUp

testing-scripts/cloud-sql-proxy/test_cloud_sql.py
```diff
@@ -0,0 +1,83 @@
...
+from dataproc_test_case import DataprocTestCase
+
+
+class CloudSqlProxyTestCase(DataprocTestCase):
+    COMPONENT = 'cloud-sql-proxy'
+    INIT_ACTION = 'gs://polidea-dataproc-utils/cloud-sql-proxy/cloud-sql-proxy.sh'
+    TEST_SCRIPT_LOCATION = 'dataproc-initialization-actions/cloud-sql-proxy'
+    TEST_SCRIPT_FILE_NAME = 'pyspark_metastore_test.py'
+    SCOPES = 'sql-admin'
+    PROPERTIES = 'hive:hive.metastore.warehouse.dir=gs://polidea-hive-metastore/hive-warehouse'
```
karth295: GCS buckets like this should not be hardcoded. Either the test suite should create a bucket, or ask the user to specify one.

testing-scripts/bigtable/test_bigtable.py
ishanrc: They do. DataprocTestCase overwrites INIT_ACTION if the 'bucket' environment variable is set https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/355/files#diff-a2f3ab15b1953c7bdb4d44372493b727R47

We create a bucket and upload the pending tests in run_tests.sh
https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/355/files#diff-dadc96d79206c24169cd546cd80eece2R14

testing-scripts/bigtable/test_bigtable.py
ishanrc: Wouldn't that break multiple people running the tests concurrently? I'll see if I can add a TTL to the DB.

testing-scripts/sqoop/test_sqoop.py
ishanrc: ack. I wonder why these tests currently work?

testing-scripts/sqoop/test_sqoop.py
ishanrc: ack

testing-scripts/bigtable/test_bigtable.py
ishanrc: ack

testing-scripts/cloud-sql-proxy/test_cloud_sql.py
ishanrc: ack

testing-scripts/bigtable/test_bigtable.py
karth295: Can we remove this INIT_ACTION variable just to make it clear then?

testing-scripts/bigtable/test_bigtable.py
karth295: Ack, good point

PR #356, opened 2018-09-27
Add a run_with_retries nc call to sql proxy after starting it.

dennishuo:
The systemctl start may return asynchronously while cloud-sql-proxy
is not yet ready, such that subsequent attempts to initialize the
metastore will fail.

### DIFF
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -161,6 +161,7 @@ EOF
   systemctl enable cloud-sql-proxy
   systemctl start cloud-sql-proxy \
     || err 'Unable to start cloud-sql-proxy service'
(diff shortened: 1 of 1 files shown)


dennishuo:
Thanks @mikaylakonst for the suggested fix! Tested cleanly on 1.2 so far.

dennishuo:
For posterity: we narrowed down the change to an upstream release of cloud-sql-proxy updating the binary under dl.google.com earlier this week, (Sept 24, 7:30 PM pacific time); if the failures started around that time, then it's highly likely the root cause and you need to pull this fix into any custom forks of the init action.

We've already updated gs://dataproc-initialization-actions/cloud-sql-proxy/cloud-sql-proxy.sh with this fix.

martynbristow:
Thanks brilliant fix

PR #359, opened 2018-10-02
Prow setup

ishanrc:
These are the Polidia authored init action test for Jupyter and configuration to run them on the prow cluster at http://35.201.82.161/

Tests now run in the dataproc-github-tests project


ishanrc:
I'll send a new PR based on https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/360

PR #360, opened 2018-10-03
added integration tests framework and test for Hive-Hcatalog

DariuszAniszewski:


integration-tests/dataproc_test_case.py
```diff
@@ -0,0 +1,143 @@
...
+import os
+
+DEFAULT_TIMEOUT = 10  # minutes
+
+
+class DataprocTestCase(unittest.TestCase):
+    DEFAULT_ARGS = {
+        "SINGLE": [
+            "--single-node",
+            "--worker-machine-type n1-standard-4",
```
aniket486: I don't think master/worker-machine-type is necessary. We should use the default (which is n1-standard-4, but not pin to it).

integration-tests/dataproc_test_case.py
```diff
@@ -0,0 +1,143 @@
...
+        assert cls.COMPONENT
+        assert cls.INIT_ACTION
+
+    def createCluster(self, configuration, init_action, dataproc_version, metadata=None, scopes=None, properties=None,
+                      timeout_in_minutes=None):
+        self.name = "test-{}-{}-{}-{}".format(
+            self.COMPONENT,
+            configuration.lower(),
+            dataproc_version.replace(".", "-"),
+            random.randint(1, 10000),
```
aniket486: Use date instead of random here for better tracking.

integration-tests/dataproc_test_case.py
```diff
@@ -0,0 +1,143 @@
...
+        ret_val, stdout, stderr = self.run_command(cmd, timeout_in_minutes=timeout_in_minutes or DEFAULT_TIMEOUT)
+
+        self.assertEqual(ret_val, 0, "Failed to create Cluster {}. Error: {}".format(
+            self.name,
+            stderr
+        ))
+        self.cluster_version = json.loads(stdout).get("config", {}).get("softwareConfig", {}).get("imageVersion")
+
+    def tearDown(self):
+        cmd = "yes | gcloud dataproc clusters delete {}".format(self.name)
```
aniket486: gcloud dataproc clusters delete {} --quiet is better.

integration-tests/dataproc_test_case.py
DariuszAniszewski: @aniket486 good point - by now I used YYYYMMDDHHmmSS - it has advantage over YYYYMMDD that you can run several clusters on the same day. WDYT?

DariuszAniszewski:
@aniket486 Thanks for review - I made some changes, PTAL

integration-tests/dataproc_test_case.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+        timeout = timeout_in_minutes * 60
+        my_timer = Timer(timeout, kill, [p])
+        try:
+            my_timer.start()
+            stdout, stderr = p.communicate()
+            stdout, stderr = stdout.decode('utf-8'), stderr.decode('utf-8')
+        finally:
+            my_timer.cancel()
+        logging.info("Ran %s: retcode: %d, stdout: %s, stderr: %s", command, p.returncode, stdout, stderr)
```
karth295: Does this print a lot of spam when you're creating a cluster?

DariuszAniszewski: Not really, but depends on `a lot` definition ;)
I believe this log is important - might held interesting info and by default INFO isn't logged - I added small line that allows to specify which log_level should be used: ac301e3226d14f457112afcdafbbf25bbb3f0452.
PTAL

integration-tests/hive-hcatalog/test_hive.py
```diff
@@ -0,0 +1,86 @@
+import json
```
karth295: Can we put the test files in the same directory as the init actions? E.g. `hive-hcatalog/test_hive.py` rather than in the integration-tests directory

DariuszAniszewski: @karth295 Technically it can be moved, but not without implications. The way it's structured right now allows to:
- separate shell scripts of init actions and tests
- allow to run all tests at once or individual tests from the `integration-tests` directory directly

If I move `test_hive.py` to `hive-hcatalog` test code will be mixed with init-action shell script which may be confusing. Also, in order to be able to run the tests, I'd have to:
- add `__init__.py` file in `hive-hcatalog` (so add another python file, not directly related to init-action shell script)
- rename `integration-tests` into `integration_tests` to be able to import `DataprocTestCase` in `test_hive.py`

In my opinion, it's not worth to move. WDYT?

karth295: Thanks for the detailed analysis!

Just to clarify: can you run all the tests together if you split them out by directory?

If you can, then I still think it's worth it because I'd like to make it clear to users and maintainers that you should modify the tests before opening a PR. I think it's okay that we're adding an empty python file in every directory.

If not (or if it isn't straightforward), then let's just leave it as is.

DariuszAniszewski:
@karth295 I moved the test to the init-action directory as you suggested, PTAL.

karth295:
Awesome, thanks! Excited to get all the tests in.

DariuszAniszewski:
@karth295 is there something blocking you from merging this PR? :)

aniket486:
Merging this PR to unblock further progress. Lets make follow up changes in subsequent PRs.

PR #367, opened 2018-10-18
Only check 3306 if using cloud sql for metastore

### LINKED ISSUE #364: cloud_sql_proxy init action fails when hive metastore is not enabled
Hi! I'm running the cloud_sql_proxy init action with `enable-cloud-sql-hive-metastore":"false"`, specifying an additional instance instead. It looks like https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/356/commits broke the initialization action for this use case, as now it fails after a while of
```About to run 'nc -zv localhost 3306' with retries...
nc: connect to localhost port 3306 (tcp) failed: Connection refused
nc: connect to localhost port 3306 (tcp) failed: Connection refused
'nc -zv localhost 3306' attempt 1 failed! Sleeping 1.
nc: connect to localhost port 3306 (tcp) failed: Connection refused
nc: connect to localhost port 3306 (tcp) failed: Connection refused
'nc -zv localhost 3306' attempt 2 failed! Sleeping 1.```

mikaylakonst:
Fixes #364

### DIFF
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -161,7 +161,10 @@ EOF
   systemctl enable cloud-sql-proxy
   systemctl start cloud-sql-proxy \
     || err 'Unable to start cloud-sql-proxy service'
-  run_with_retries nc -zv localhost ${metastore_proxy_port}
+    
+  if [[ $enable_cloud_sql_metastore = "true" ]]; then
+    run_with_retries nc -zv localhost ${metastore_proxy_port}
+  fi
 
   echo 'Cloud SQL Proxy installation succeeded' >&2
 


mikaylakonst:
More info on the cloud sql proxy:
https://cloud.google.com/sql/docs/mysql/sql-proxy

It looks like when using the cloud sql proxy for the hive metastore, we use tcp port 3306 by default if you just specify an instance name:
https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/9dab3e18a7877224289b84447cb34378d1192f01/cloud-sql-proxy/cloud-sql-proxy.sh#L121

On the other hand, when specifying additional instances, we use unix sockets in /var/run/cloud_sql_proxy by default if you just specify an instance name with no tcp port.

If we really wanted to, we could loop through the additional instances and, if a tcp port is specified, do nc -zv localhost <that port> before confirming successful installation of cloud sql proxy

PR #380, opened 2018-11-06
Ganglia integration tests

adam-stasiak:
These scripts can be used for testing Ganglia init action on single, standard and ha configurations.

### DIFF
--- a/ganglia/test_ganglia.py
+++ b/ganglia/test_ganglia.py
@@ -1,8 +1,7 @@
 import unittest
 
 from parameterized import parameterized
-
-from dataproc_test_case import DataprocTestCase
+from integration_tests.dataproc_test_case import DataprocTestCase
 
 
 class GangliaTestCase(DataprocTestCase):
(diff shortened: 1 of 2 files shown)


ganglia/verify_ganglia_running.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, base, port):
+        self.path = 'http://{}:{}/ganglia/'.format(base, port)
+        self.host = socket.gethostname()
+        self.page = None
+        self.is_main_master = None
+
+    def validate_homepage(self):
+        if self.page:
+            if '-w-' in self.host:
+                raise Exception('NOK - Ganglia should not be running on worker node')
```
karth295: Since it's already an exception, you can remove `NOK -` (here any everywhere).

ganglia/verify_ganglia_running.py
```diff
@@ -0,0 +1,74 @@
...
+                    raise Exception('NOK - Ganglia UI is not found on master node')
+        else:
+            if '-w-' in self.host:
+                print('OK - Ganglia is not running on worker node')
+            if '-m-' in self.host and not (self.is_main_master):
+                print('OK - Ganglia is not designed to be running on additional master nodes')
+
+    def get_homepage(self):
+        try:
+            self.page = requests.get(self.path).text[:200]
```
karth295: A getter method should not mutate the state of an object -- this means that validate_homepage must come after get_homepage, which is confusing. Consider making GangliaApi purely an API client, and putting logic in separate functions.

Something like:

```
page = ganglia_api.get_page()
validate_homepage(page)
```

Same feedback with `detect_role()`.

karth295: You should use an HTML parser if this is HTML. It's scary to assume that the first 200 chars remain the same across versions.

ganglia/test_ganglia.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __run_test_file(self, name):
+        cmd = 'gcloud compute ssh {} -- "python3 {} "'.format(
+            name,
+            self.TEST_SCRIPT_FILE_NAME
+        )
+        ret_code, stdout, stderr = self.run_command(cmd)
+        self.assertEqual(ret_code, 0, "Failed to run test file. Error: {}".format(stderr))
+
+    def __run_command_on_cluster(self, name, command):
```
karth295: Nit: This seems useful enough to go in `DataprocTestCase`

adam-stasiak: yes but unfortunately DataprocTestCase is not uploaded to cluster so I can't apply re-usable solution for this. But I think the same about it and helper methods like this would be nice in the future.

karth295: Ah gotcha -- after all the tests are merged, we should think about refactoring these utilities into a common library. Pyspark jobs do let you bring additional libraries, fwiw.

But for now I'll go ahead and merge this PR.

PR #381, opened 2018-11-06
Hue integration tests

adam-stasiak:
These scripts can be used for testing Hue init action on single, standard and ha configurations.


hue/create.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
karth295: Is this file ever used? I think it needs to be deleted

hue/verify_hue_running.py
```diff
@@ -0,0 +1,88 @@
...
+
+import requests
+import socket
+import subprocess
+
+BASE = 'localhost'
+PORT = 8888
+
+
+class HueApi(object):
```
karth295: Similar feedback on the design of this class as in GangliaApi -- getter methods should not mutate the state of the object. Consider making these *Api classes stateless, and make assertions in other functions

adam-stasiak:
Review changes applied.

adam-stasiak:
@karth295 @functicons Could you take a look on this please?

adam-stasiak:
@karth295 All changes are applied. Could you check this once again?

PR #382, opened 2018-11-06
Oozie integration tests

adam-stasiak:
These scripts can be used for testing Oozie init action on single, standard and ha configurations.

### DIFF
--- a/oozie/validate.sh
+++ b/oozie/validate.sh
@@ -1,11 +1,21 @@
 #!/bin/bash
-echo "starting validation script"
+set -euxo pipefail
+
+function err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
+  return 1
+}
+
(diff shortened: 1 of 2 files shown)


oozie/validate.sh
```diff
@@ -0,0 +1,50 @@
...
+
+hdfs dfs -mkdir -p /user/$(whoami)/
+hadoop fs -put ~/examples/ /user/$(whoami)/
+
+echo "---------------------------------"
+echo "Starting validation on ${hostname}"
+oozie job -oozie http://localhost:11000/oozie -config /home/$(whoami)/examples/apps/map-reduce/job.properties -run
+oozie jobs -oozie http://localhost:11000/oozie -localtime -len 2 | grep "No Jobs match your criteria!"
+
+if [[ $? == 1 ]]; then
```
karth295: You could also use `set -euxo pipefail` at the top of this file

PR #383, opened 2018-11-06
Integration test for Bigtable init action.

szewi:
To run the test just use command:
  python -m unittest bigtable.test_bigtable

### DIFF
--- a/bigtable/run_hbase_commands.py
+++ b/bigtable/run_hbase_commands.py
@@ -4,23 +4,20 @@ import subprocess
 
 
 def create_commands_file():
-    file = open("commands", 'w+')
-    file.write("create 'test-bigtable', 'cf'\n"
-               "list 'test-bigtable'\n"
-               "put 'test-bigtable', 'row1', 'cf:a', 'value1'\n"
-               "put 'test-bigtable', 'row2', 'cf:b', 'value2'\n"
-               "put 'test-bigtable', 'row3', 'cf:c', 'value3'\n"
-               "put 'test-bigtable', 'row4', 'cf:d', 'value4'\n"
-               "exit"
-               )
-    file.close()
+    with open("commands.txt", "w+") as file:
+        file.write("create 'test-bigtable', 'cf'\n"
+                   "list 'test-bigtable'\n"
+                   "put 'test-bigtable', 'row1', 'cf:a', 'value1'\n"
+                   "put 'test-bigtable', 'row2', 'cf:b', 'value2'\n"
+                   "put 'test-bigtable', 'row3', 'cf:c', 'value3'\n"
+                   "put 'test-bigtable', 'row4', 'cf:d', 'value4'\n"
+                   "exit")
 
 
 def main():
     create_commands_file()
-    subprocess.check_output(['hbase', "shell", 'commands'])
+    subprocess.check_output(['hbase', "shell", 'commands.txt'])
 
 
 if __name__ == '__main__':
     main()
-
--- a/bigtable/run_hbase_commands.py
+++ b/bigtable/run_hbase_commands.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python
+
+import subprocess
+
+
+def create_commands_file():
+    file = open("commands", 'w+')
+    file.write("create 'test-bigtable', 'cf'\n"
+               "list 'test-bigtable'\n"
+               "put 'test-bigtable', 'row1', 'cf:a', 'value1'\n"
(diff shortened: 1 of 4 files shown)


szewi:
@karth295 @aniket486 could you take a look?
From all PRs with tests, I will suggest prioritizing this one. It changes the way we upload test scripts in integration_tests/dataproc_test_case.py file. Following PRs with tests will rely on this change.

bigtable/run_hbase_commands.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python
+
+import subprocess
+
+
+def create_commands_file():
+    file = open("commands", 'w+')
```
karth295: Nit: `with open(...) as file` (and then you don't need `file.close()`)

karth295: Nit: maybe `commands.txt` just to make it clear that it's a file. I was confused about `hbase shell commands` below

bigtable/test_bigtable.py
```diff
@@ -0,0 +1,111 @@
...
+
+from parameterized import parameterized
+from integration_tests.dataproc_test_case import DataprocTestCase
+
+
+class BigTableTestCase(DataprocTestCase):
+    COMPONENT = 'bigtable'
+    INIT_ACTION = 'gs://dataproc-initialization-actions/bigtable/bigtable.sh'
+    TEST_SCRIPT_FILE_NAME = "run_hbase_commands.py"
+    METADATA = None
```
karth295: Nit: if these aren't constants they should be initialized in `__init__` as `self.lowercasevariable = None`

bigtable/test_bigtable.py
```diff
@@ -0,0 +1,111 @@
...
+        self.assertEqual(ret_code, 0, "Failed to validate cluster. Last error: {}".format(stderr))
+
+    def _validate_bigtable(self):
+        ret_code, stdout, stderr = self.run_command(
+            'cbt -instance {} read test-bigtable '.format(
+                self.DB_NAME
+            )
+        )
+        self.assertEqual(ret_code, 0, "Failed to validate cluster. Last error: {}".format(stderr))
+        self.assertIn("value1", stdout, "Failed to validate cluster. value1 is missing in the table")
```
karth295: Can you assert the exact value of the output?

bigtable/test_bigtable.py
```diff
@@ -0,0 +1,111 @@
...
+            os.path.join(
+                os.path.dirname(os.path.abspath(__file__)),
+                self.TEST_SCRIPT_FILE_NAME
+            ),
+            name)
+        self._run_hbase_shell(name)
+        self._validate_bigtable()
+
+
+    @parameterized.expand([
```
karth295: Please add a comment on why 1.0 is not in the list

integration_tests/dataproc_test_case.py
```diff
@@ -82,13 +82,9 @@ def tearDown(self):
...
 
-    def upload_test_file(self, name):
+    def upload_test_file(self, testfile, name):
         cmd = 'gcloud compute scp {} {}:'.format(
-            os.path.join(
-                os.path.dirname(os.path.abspath(__file__)),
-                self.COMPONENT.lower(),
-                self.TEST_SCRIPT_FILE_NAME
-            ),
+            testfile,
```
karth295: If `TEST_SCRIPT_FILE_NAME` is no longer used in the base class, it should be removed from it

szewi: ok, I will also modify other methods that are using this variable.

bigtable/test_bigtable.py
szewi: 1.1 is also on the list but not working. I will remove 1.1 too adding proper comment.

bigtable/test_bigtable.py
szewi: The output contains current timestamps which will be problematic to assert. Instead, I will use `count` function and compare its output with the expected number of rows.

bigtable/test_bigtable.py
szewi: ok

bigtable/run_hbase_commands.py
szewi: ok

szewi: using `with open` make a lot of sense when dealing with files. Thanks.

szewi:
@karth295 this is ready for another round of review. Thanks.

PR #389, opened 2018-11-28
Init action with Apache HBase.

szewi:
Basic integration test for Apache HBase.
Test can be run using:
  python -m unittest hbase.test_hbase

### DIFF
--- a/hbase/test_hbase.py
+++ b/hbase/test_hbase.py
@@ -1,0 +1,52 @@
+import unittest
+from parameterized import parameterized
+
+from integration_tests.dataproc_test_case import DataprocTestCase
+
+
+class HBaseTestCase(DataprocTestCase):
+    COMPONENT = 'hbase'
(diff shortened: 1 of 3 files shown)


hbase/test_hbase.py
```diff
@@ -0,0 +1,52 @@
...
+        self.assertEqual(ret_code, 0, "Failed to validate cluster. Error: {}".format(stderr))
+
+    """
+    Single and standard mode on 1.0 and 1.1 are disabled because HBase doesn't work in those 
+    modes on older dataproc versions. HBase is intended to run on HA environment.
+    HA mode on 1.0 and 1.1 doesn't pass integration test.
+    """
+    @parameterized.expand([
+        # ("SINGLE", "1.0", ["m"]),
+        # ("STANDARD", "1.0", ["m"]),
```
aniket486: Let's remove these comments.

szewi: Thanks @aniket486 for a review. Not sure if I understood correctly, but I removed commented lines, let me know if it's ok now.

PR #396, opened 2018-12-11
Support connectors in Dataproc lib directory

medb:


functicons:
As discussed, we should still assume connector is in /usr/lib/hadoop/lib, but update symbolic link from Spark if necessary.

aniket486:
@medb can you make the suggested changes?

PR #402, opened 2018-12-21
ZooKeeper: support user-specified cluster properties

functicons:


zookeeper/zookeeper.sh
```diff
@@ -63,5 +64,12 @@ server.2=${CLUSTER_NAME}-w-0:2888:3888
 server.3=${CLUSTER_NAME}-w-1:2888:3888
 EOF
 
+# Merge user-specified zookeeper properties on cluster creation.
```
mikaylakonst: User-specified properties have already been merged into the zoo.cnf file by the time initialization actions are run.
We should change this initialization action so that it appends the above properties to the end of zoo.cnf instead of clobbering the contents of zoo.cnf.

mikaylakonst:
LGTM

PR #406, opened 2018-12-27
HBase can use GCS as storage backend.

szewi:
This commit: 
 - allows overwriting hbase.root.dir by passing hbase-root-dir as metadata
 - adds necessary gcs-connector to HBase classpath on dataproc 1.3
 - adds integration test that validate using GCS as data backend
 - updates README


hbase/hbase.sh
```diff
@@ -76,36 +127,34 @@ EOF
...
     --configuration_file "${HBASE_HOME}/conf/hbase-site.xml" \
     --source_configuration_file hbase-site.xml.tmp \
     --clobber
-  
-  # On single node clusters we must also start regionserver on it.
+
+  # On single node clusters we must also start regionserver on it and provide zookeeper-server.
   if [[ "${WORKER_COUNT}" -eq 0 ]]; then
-    "${HBASE_HOME}/bin/hbase-daemon.sh" start regionserver
+    install_apt_get zookeeper-server || err 'Unable to install Zookeeper on single node cluster.'
```
aniket486: Let's not special case zookeeper installation on single node cluster. Similar to standard cluster, we can use initiailization action to install zookeeper even on the single node cluster.

szewi: @aniket486 I agree this was just workaround, so then first we need to modify zookeeper init action, because in [README](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/zookeeper/README.MD#important-notes) of zookeeper init action it is mentioned that: 
> This script will not work on Single Node clusters, as it requires at least 3 nodes.

hbase/hbase.sh
```diff
@@ -54,20 +59,66 @@ function install_and_configure_hbase() {
...
   if [[ "${MASTER_ADDITIONAL}" != "" ]]; then
     # If true than init action is running on high availability dataproc cluster.
     # HBase will use zookeeper service pre-installed and running on master nodes.
     bdconfig set_property \
       --configuration_file 'hbase-site.xml.tmp' \
       --name 'hbase.zookeeper.quorum' --value "${zookeeper_nodes}-0,${MASTER_ADDITIONAL}" \
       --clobber
-    bdconfig set_property \
-      --configuration_file 'hbase-site.xml.tmp' \
-      --name 'hbase.rootdir' --value "hdfs://${CLUSTER_NAME}-m-0:8020/hbase" \
```
szewi: Removing this will cause HBase master being unable to start on standard and single node clusters. I will refactor this part of code to be more readable.

hbase/hbase.sh
```diff
@@ -76,36 +127,34 @@ EOF
       --configuration_file 'hbase-site.xml.tmp' \
       --name 'hbase.zookeeper.quorum' --value "${zookeeper_nodes}" \
       --clobber
-    bdconfig set_property \
-      --configuration_file 'hbase-site.xml.tmp' \
-      --name 'hbase.rootdir' --value "hdfs://${CLUSTER_NAME}-m:8020/hbase" \
```
szewi: Removing this will cause HBase master being unable to start on standard and single node clusters. I will refactor this part of the code to be more readable.

hbase/hbase.sh
szewi: I removed this line in latest commits. Now when creating standard or single cluster it is required to additionally pass zookeeper init action. The fix for running zookeeper on single clusters is here: https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/418.

szewi:
Hi, now the configuration part of Hbase is more readable. It is divided into two parts. One is for setting `hbase.root.dir` and second is for `zookeeper.quorum`. Instead of manual generating of zookeeper nodes, now I am reading it from the zookeeper configuration file, which is more elegant. 
Creating single and standard clusters require additionally using zookeeper init action, it is mentioned in readme. I've also extended zookeeper init action for support to single node clusters in PR #418 . Let me know what do you think @aniket486

PR #411, opened 2019-01-04
ZooKeeper: write config on all nodes whether it is part of the cluster

functicons:


medb:
Why is this necessary?

functicons:
> Why is this necessary?

Because other components (e.g., [Kafka](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/415a984/kafka/kafka.sh#L56)) might need to know where zookeeper is. It is okay to just know 1 master node (the default config), but it is better to know all for redundancy.

PR #418, opened 2019-01-14
Zookeeper init action can be run on single node clusters.

szewi:


szewi:
HBase in pseudo-distributed setup(running on single node) needs zookeeper service. Instead of installing zookeeper in init action, we agreed that would be better to pass additionally zookeeper init action to `--initialization-actions` during the cluster creation process(the same will be for standard clusters).  Can you take a look @aniket486 ? I am also mentioning @functicons who was maintaining/updating zookeeper recently. What do you think? In the meantime, I will prepare a fix for HBase, that will work with changes from this PR.

aniket486:
Thanks @szewi !

PR #419, opened 2019-01-15
Init action for Knox with automated tests

adam-stasiak:
This init actions installs and launch Knox gateway. You can easily test this with automated test attached to PR with command specified in readme or following steps described in official manual of Knox service.
Test is running only under 1.3 because for 1.2 and 1.1 knox package is missing in apt-get repo.

### DIFF
--- a/knox/knox.sh
+++ b/knox/knox.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+# Copyright 2016 Google, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This init script installs knox on each master node in the cluster
+
+set -xe
+
+export GATEWAY_HOME="/usr/lib/knox"
+
+function err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
+  return 1
+}
+
+function prepare_helper(){
+  cd ${GATEWAY_HOME}
+  cat << 'EOF' > bin/knox_helper.sh
+#!/usr/bin/expect -f
+
+set timeout 1800
+set cmd [lindex $argv 0]
+
+spawn {*}$cmd
+expect {
+    "Enter master secret:" {
+        exp_send "pass\r"
+        exp_continue
+     }
+     "Enter master secret again:" {
+        exp_send "pass\r"
+        exp_continue
+     }
+     eof
+}
+EOF
+}
+
+function create_master(){
+  /usr/bin/expect ${GATEWAY_HOME}/bin/knox_helper.sh "${GATEWAY_HOME}/bin/knoxcli.sh create-master --force"
+  sudo chmod -R u+w ${GATEWAY_HOME}/data/security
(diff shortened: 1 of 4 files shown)


knox/knox.sh
```diff
@@ -0,0 +1,127 @@
...
+  sudo su - knox -p knox -c "${GATEWAY_HOME}/bin/gateway.sh start"
+}
+
+function install_dependencies(){
+  yes | apt-get install knox || err "Failed to install Knox"
+  yes | apt-get install expect || err "Failed to install expect"
+}
+
+function add_user(){
+  useradd knox
```
aniket486: Let's keep this step optional so that it works well with Dataproc optional component.

adam-stasiak: I am not sure if I made this in the right way - I've just added if statement to determine if user knox already exists in case of the situation when user knox is already created by apt-get install knox operation. Please let me know if is it ok.

adam-stasiak: @aniket486 Just remind - I applied all suggestions :)

adam-stasiak:
@karth295 Could you take a look on this?

knox/knox.sh
```diff
@@ -0,0 +1,133 @@
...
+}
+
+function set_webhdfs_address(){
+  local host
+  local port
+  local master
+
+  master="$(/usr/share/google/get_metadata_value attributes/dataproc-master)"
+  host=$(hdfs getconf -confKey dfs.namenode.http-address)
+  port="$(echo ${host} | sed -e 's,^.*:,:,g' -e 's,.*:\([0-9]*\).*,\1,g' -e 's,[^0-9],,g')"
```
mik-laj:
```suggestion
  port="$(echo "${host}" | sed -e 's,^.*:,:,g' -e 's,.*:\([0-9]*\).*,\1,g' -e 's,[^0-9],,g')"
```
I think, you should use double quote to prevent globbing and word splitting.

knox/knox.sh
```diff
@@ -0,0 +1,133 @@
...
+function grant_knox_folder_permissions(){
+  chmod -R u+w ${GATEWAY_HOME}
+  chown -R knox:sudo ${GATEWAY_HOME}
+}
+
+function configure_ha(){
+  local master
+  master="$(/usr/share/google/get_metadata_value attributes/dataproc-master)"
+
+  if [[ "${HOSTNAME}" == ${master} ]];then
```
mik-laj:
```suggestion
  if [[ "${HOSTNAME}" == ${master} ]];then
```
Quote the right-hand side of == in [[ ]] to prevent glob matching.

knox/knox.sh
```diff
@@ -0,0 +1,133 @@
...
+    hdfs dfs -mkdir -p /user/$(whoami)/
```
mik-laj:
```suggestion
    hdfs dfs -mkdir -p "/user/$(whoami)/"
```
Quote this to prevent word splitting.

knox/knox.sh
```diff
@@ -0,0 +1,133 @@
...
+    hadoop fs -put ${GATEWAY_HOME}/data/security/keystores/ /user/$(whoami)/
```
mik-laj:
```suggestion
    hadoop fs -put ${GATEWAY_HOME}/data/security/keystores/ "/user/$(whoami)/"
```
Quote this to prevent word splitting.

knox/knox.sh
```diff
@@ -0,0 +1,133 @@
...
+  else
+    until [ "$(hdfs dfs -ls /user/$(whoami)/keystores || echo $?)" != "0" ]; do
+       hdfs dfs -get -f /user/$(whoami)/keystores ${GATEWAY_HOME}/data/security/
```
mik-laj:
```suggestion
       hdfs dfs -get -f "/user/$(whoami)/keystores" "${GATEWAY_HOME}/data/security/"
```
Quote this to prevent word splitting.

karth295:
Knox gets installed when you enable Component Gateway, so I'm inclined to close this PR for now.

PR #421, opened 2019-01-17
Fix TonY README.md

gogasca:
Correct wording
Remove inaccurate statement in regards GPU

### DIFF
--- a/tony/README.md
+++ b/tony/README.md
@@ -72,8 +72,8 @@ You can easily test TonY is running by using ```validate.sh ${cluster_prefix} ${
 * ```cluster_prefix``` argument sets a prefix in created cluster name
 * ```bucket_name``` argument sets a place where init action will be placed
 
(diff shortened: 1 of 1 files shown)


tony/README.md
```diff
@@ -72,13 +72,12 @@ You can easily test TonY is running by using ```validate.sh ${cluster_prefix} ${
 * ```cluster_prefix``` argument sets a prefix in created cluster name
 * ```bucket_name``` argument sets a place where init action will be placed
 
-This script is used for testing dr-elephant on 1.1,1.2,1.3 dataproc images with single,
-standard and HA configurations. After clusters are created, script submit spark jobs on them.
+This script is used for testing Tony using 1.3 Dataproc images with standard configurations. 
+After clusters are created, script submit Hadoop jobs on them.
```
medb: submits
?

PR #431, opened 2019-01-23
Add script to push an init action module to GCS.

functicons:


push-to-gcs.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# This script is intended to be used by the Dataproc team to push init action
+# modules (e.g., kafka) to the Dataproc init actions GCS bucket. It performs
+# necessary validations to make sure the repo is in good shape before pushing.
+#
+# Usage: ./push-to-gcs.sh <head> <module>
```
medb: `head` -> `git-commit` or `gir-ref`

push-to-gcs.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+# Example: ./push-to-gcs.sh 119952 kafka
+
+set -euo pipefail
+
+[[ $# -eq 2 ]] || {
+    echo "Usage: ./push-to-gcs.sh <head> <module>"
```
medb: ditto

push-to-gcs.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
```
medb: Should we run integration test before pushing changes to GCS as well?

functicons: Currently integration test are running against the published init actions. They cannot test local changes. We need to think about how to change this later.

PR #433, opened 2019-01-24
Adds deprecation note to stackdriver initialization action

jaketf:
@medb


medb:
Thank you for contribution!

PR #435, opened 2019-01-24
Kafka: de-duplicate ZooKeeper addresses

functicons:


kafka/kafka.sh
```diff
@@ -56,6 +56,7 @@ function install_and_configure_kafka_server() {
   zookeeper_list=$(grep '^server\.' /etc/zookeeper/conf/zoo.cfg \
     | cut -d '=' -f 2 \
     | cut -d ':' -f 1 \
+    | uniq \
```
medb: add `sort` before `uniq` to ensure de-duplication.

PR #439, opened 2019-01-29
 Corrected path to toree installer

czarstark:
Path pointed to an outdated, in-development version of toree. It now points to a released version.


czarstark:
I signed it!

czarstark:
I signed it (again)!

PR #446, opened 2019-02-07
Solr in HA clusters is now available in SolrCloud mode.

szewi:
This commit enables SolrCloud mode on HA dataproc clusters.
Solr download link was switched to apache archives to avoid any future installation failures.
SolrCloud configuration is stored in separate zookeeper znode which is recommended for production deployments.

### DIFF
--- a/solr/solr.sh
+++ b/solr/solr.sh
@@ -30,6 +30,11 @@ function err() {
 }
 
 function install_and_configure_solr() {
+  local solr_home_dir
+  local zookeeper_nodes
+  zookeeper_nodes="$(grep '^server\.' /etc/zookeeper/conf/zoo.cfg \
+    | uniq | cut -d '=' -f 2 | cut -d ':' -f 1 | xargs echo | sed "s/ /,/g")"
+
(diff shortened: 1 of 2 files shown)


solr/solr.sh
```diff
@@ -19,21 +19,48 @@
...
 
-systemctl start solr || err 'Unable to start solr service.'
+  sed -i "s/^#SOLR_HOST=\"192.168.1.1\"/SOLR_HOST=\"${NODE_NAME}\"/" \
+    /etc/default/solr.in.sh
+  mkdir -p /var/log/solr && chown solr:solr /var/log/solr
+  sed -i 's/^SOLR_LOGS_DIR="\/var\/solr\/logs"/SOLR_LOGS_DIR="\/var\/log\/solr"/' \
+    /etc/default/solr.in.sh
+  # Enable SolrCloud setup in HA mode.
+  if [[ "${MASTER_ADDITIONAL}" != "" ]]; then
+    sed -i "s/^#ZK_HOST=\"\"/ZK_HOST=\"${CLUSTER_NAME}-m-0,${MASTER_ADDITIONAL}\/solr\"/" \
```
aniket486: Can we get this by parsing zoo.cfg file (similar to how we've done in other places)?

szewi: Sure, this makes a lot of sense and would be consistent with other scripts.

PR #462, opened 2019-02-19
[flink]: support HA YARN session

### LINKED ISSUE #272: Flink initializing script uses only first master name when high availability cluster is created
The problem is with :
`/usr/share/google/get_metadata_value attributes/dataproc-master` 
It will always return the first master e.g - cluster-name-m-0 even when the script is currently running on the second master. The other masters are available only through the   `attributes/dataproc-master-additional ` option, but it returns all other masters and not the current one. Of course it could be used `hostname` as a variable, but this is not a clean way, cause if you try to setup something in the slaves, hostname will be resolved as the slave's name and will not point to the master.

szewi:
Fixes #272


flink/flink.sh
```diff
@@ -158,6 +163,18 @@ taskmanager.network.numberOfBuffers: ${FLINK_NETWORK_NUM_BUFFERS}
 fs.hdfs.hadoopconf: ${HADOOP_CONF_DIR}
 EOF
 
+  if [[ "${MASTER_ADDITIONAL}" != "" ]]; then
```
medb: Use `-n` check instead of `!= ""`

flink/README.md
```diff
@@ -44,3 +44,4 @@ You can find more information about using initialization actions with Dataproc i
 * By default, a detached Flink YARN session is started for you. To find its application id, run `yarn application -list`.
 * The default session is configured to consume all YARN resources. If you want to submit multiple jobs in parallel or use transient sessions, you'll need to disable this default session. You can either fork this init script and set START_FLINK_YARN_SESSION_DEFAULT to `false`, or set the cluster metadata key `flink-start-yarn-session` to `false` when you create your cluster.
 * By default, flink is installed via apt.  However, a custom flink snapshot can be installed by specifying the `flink-snapshot-url` metadata key to a URL that points to a valid flink tarball.
+* When creating **HA cluster** with Flink please add cluster property `--properties 'yarn:yarn.resourcemanager.am.max-attempts=4'` to enable HA YARN session. You can find more information in the Flink [documentation](https://ci.apache.org/projects/flink/flink-docs-release-1.7/ops/jobmanager_high_availability.html#yarn-cluster-high-availability).
```
medb: May you modify Flink test to set this property for HA clusters and add Dataproc 1.4 configuration?

flink/flink.sh
```diff
@@ -158,6 +163,18 @@ taskmanager.network.numberOfBuffers: ${FLINK_NETWORK_NUM_BUFFERS}
...
+    zookeeper_nodes="$(grep '^server\.' /etc/zookeeper/conf/zoo.cfg | uniq | cut -d '=' -f 2 \
+    | cut -d ':' -f 1 | xargs -L1 echo | sed "s/$/:2181/g" | xargs echo | sed "s/ /,/g")"
+    cat << EOF >> ${FLINK_INSTALL_DIR}/conf/flink-conf.yaml
+high-availability: zookeeper
+high-availability.zookeeper.quorum: ${zookeeper_nodes}
+high-availability.zookeeper.storageDir: hdfs:///flink/recovery
+high-availability.zookeeper.path.root: /flink
+yarn.application-attempts: 10
```
medb: Why is this value 10, but `yarn.resourcemanager.am.max-attempts` property is recommended to be set to 4?

PR #464, opened 2019-02-19
Update version-less connector link

medb:


connectors/connectors.sh
```diff
@@ -61,6 +61,11 @@ update_connector() {
       local path="gs://hadoop-lib/${name}/${name}-connector-${version}-hadoop2.jar"
     fi
     gsutil cp "$path" "${vm_connectors_dir}/"
+    
+    # Update version-less connector link
+    if [[ -f ${vm_connectors_dir}/${name}-connector.jar ]]; then
+      ln -s -f "${vm_connectors_dir}/${name}-connector-"* "${vm_connectors_dir}/${name}-connector.jar"
```
functicons: Use name instead of wildcard.

PR #465, opened 2019-02-21
Add Sqoop init action (#268).

szewi:


sprzedwojski:
It would be awesome if this PR could be merged in the near future, as we are working on an [Oozie-to-Airflow converter](https://github.com/GoogleCloudPlatform/oozie-to-airflow) and one of the Oozie actions that we will support is the [Sqoop action](https://oozie.apache.org/docs/5.1.0/DG_SqoopActionExtension.html).

In order to run an end-to-end test, where our converted Airflow DAG will communicate with Sqoop, we will need a Dataproc cluster with Sqoop installed.

potiuk:
Absolutely! We need it!

sqoop/test_sqoop.py
```diff
@@ -0,0 +1,284 @@
+import unittest
+import datetime
+from parameterized import parameterized
+
+from integration_tests.dataproc_test_case import DataprocTestCase
+
+
+class SqoopTestCase(DataprocTestCase):
```
medb: May you rebase this PR on master and fix tests?

szewi:
@medb I added support for fastunit. Hive metastore is no longer configured to use Cloud SQL, and cloud sql proxy is configured to use Cloud SQL instance on port 3307. This instance is reused by all test cases. Pleases also notice that tests which imports to hbase uses intentionally `machine_type="n1-standard-2"` instead of the default one n1-standard-1. Running HBase on slowest machines especially on Dataproc 1.3 in Single Node is problematic. Usually there is a shortage of the resources to run MapReduce jobs resulting in various errors like:
> 19/07/18 09:25:28 INFO mapreduce.Job:  map 0% reduce 0%
> 19/07/18 09:26:30 INFO retry.RetryInvocationHandler: java.io.IOException

szewi:
@medb One more comment I change the way we provision database, now it's a part of the setUpClass. I created a dump ot test database and uploaded it to publicly available gs bucket. So now:
Ran 24 tests in `2035.479`s --> Ran 24 tests in `922.432`s.
It turned out that previous implementation when 18 cluster were writing to CloudSQL at the same time results in significant slower upload of test db. Increase: 24s -> ~9minutes.

medb:
@aniket486 should we merge this PR as is or modify to install Sqoop from apt repo?

functicons:
Can we merge this PR first?

PR #469, opened 2019-02-25
Intel BigDL 0.7 and Analytics Zoo 0.4 support for dataproc init script

abdolence:
The current version of BigDL init script has a very old link to old version of BigZL and doesn't support BigDL with Zoo configurations.
This PR fixes this.

### DIFF
--- a/bigdl/README.md
+++ b/bigdl/README.md
@@ -1,9 +1,12 @@
-# Intel BigDL
+# Intel BigDL and Analytics Zoo
 
 This initialization action installs [BigDL](https://github.com/intel-analytics/BigDL)
 on a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster.
 BigDL is a distributed deep learning library for Apache Spark. More information can be found on the
 [project's website](https://bigdl-project.github.io/)
+
+This script also supports Intel Analytics Zoo which includes BigDL as well. 
+More information [project's website](https://analytics-zoo.github.io) 
 
 ## Using this initialization action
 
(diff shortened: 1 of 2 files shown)


abdolence:
I signed it!

bigdl/bigdl.sh
```diff
@@ -66,11 +66,31 @@ if [[ "${ROLE}" == "Master" ]]; then
...
+# Check if it BigDL conf or Zoo  
+
+SPARK_BIG_DL_CONF=
+
+if [ -f conf/spark-bigdl.conf ]; then
+  SPARK_BIG_DL_CONF=conf/spark-bigdl.conf
+elif [ -f conf/spark-analytics-zoo.conf ]; then
+  SPARK_BIG_DL_CONF=conf/spark-analytics-zoo.conf
+else	
+  echo "Can't find any suitable spark config for Intel BigDL/Zoo"
```
karth295: Nit: if you exit right here, you won't need the extra if statements below.

This is the error function we use throughout the init actions -- feel free to copy paste it at the top:

```bash
function err() {
  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
  return 1
}
```

And then use it like this:

```bash
err "Can't find any suitable spark config for Intel BigDL/Zoo"
```

bigdl/bigdl.sh
```diff
@@ -66,11 +66,31 @@ if [[ "${ROLE}" == "Master" ]]; then
   # will never come up.
   SPARK_NUM_EXECUTORS=$((SPARK_NUM_EXECUTORS_PER_NODE_MANAGER * CURRENTLY_RUNNING_NODEMANAGERS - 1))
 
-  cat conf/spark-bigdl.conf >> /etc/spark/conf/spark-defaults.conf
+# Check if it BigDL conf or Zoo  
+
+SPARK_BIG_DL_CONF=
```
karth295: Nit: Consider putting the `cat blah >> blah` in the if statements to avoid this extra variable

bigdl/README.md
```diff
@@ -27,6 +35,17 @@ gcloud dataproc clusters create <CLUSTER_NAME> \
     --metadata 'bigdl-download-url=https://s3-ap-southeast-1.amazonaws.com/bigdl-download/dist-spark-1.6.2-scala-2.10.5-all-0.4.0-dist.zip'
```
karth295: Is there a new version for Spark 1.6 as well?

bigdl/README.md
```diff
@@ -27,6 +35,17 @@ gcloud dataproc clusters create <CLUSTER_NAME> \
...
 ```
 
+Or, for example, to download Analytics Zoo 0.4.0 with BigDL v0.7.2 use this:
+
+```
+gcloud dataproc clusters create <CLUSTER_NAME> \
+    --image-version 1.0 \
```
karth295: This should be `--image-version=1.3` since you need spark 2.3.

abdolence:
> Out of curiosity, does the new BigDL version still work on Dataproc 1.2 (Spark 2.2.1?). Just asking for backwards compatibility reasons.

Never tried it myself, but BigDL v0.7 should work even with Spark 1.6. The related packages are available to download from here https://repo1.maven.org/maven2/com/intel/analytics/bigdl/.

abdolence:
> Also, consider adding an integration test, if you have time :) E.g. https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/391/files
:) It looks that I need more time (and being Python-minded :P) for this that I have had at the moment.

PR #472, opened 2019-02-28
Flink init action: Set HADOOP_PATH env var to start the yarn session.

jphalip:
Without this, an exception is raised:
>java.lang.NoClassDefFoundError: org/apache/hadoop/conf/Configuration


medb:
May you provide repro for this issue? We have tests and they are passing with default configuration.

PR #470, opened 2019-02-25
Installing Solr using bigtop debian packages

ramjammigumpula:
Installing Solr using Bigtop Debian packages

PR #473, opened 2019-03-01
Update ranger to work with Solr installed as debian package.

szewi:


szewi:
Once #470 will be approved this PR need to be merged too.

PR #483, opened 2019-03-23
Solr integration test.

szewi:
python -m unittest solr/test_solr.py

### DIFF
--- a/solr/test_solr.py
+++ b/solr/test_solr.py
@@ -1,0 +1,112 @@
+import unittest
+import json
+from parameterized import parameterized
+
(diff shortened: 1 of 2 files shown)


szewi:
SOLR_DIR must be changed when Solr will be installed using apt-get method. This change is introduced in PR https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/470

szewi:
Thanks, I can see that this will simplify sending JSONs over gcloud ssh command and reduces complexity in the main test class.

PR #485, opened 2019-03-28
Apache Ranger integration test.

szewi:
python -m unittest ranger/test_ranger.py

### DIFF
--- a/ranger/__init__.py
+++ b/ranger/__init__.py
@@ -1,0 +1,1 @@
+# This file is required for integration tests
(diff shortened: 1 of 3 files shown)


szewi:
@aniket486 @ramjammigumpula Here you can find integration tests for ranger init action. Detailed test description can be found in verify_ranger.py script. 
At this moment the test suite is running on Datatproc 1.2 in all configurations. On Dataproc 1.3 init action will fail because of the yarn plugin issue.

PR #490, opened 2019-04-03
Update tony.sh

gogasca:
Update tony.sh to:

- Latest TonY release 0.3.1
- Latest TensorFlow version 1.13.1
- Resize cluster from 4g to 2g Parameter Server

### DIFF
--- a/tony/tony.sh
+++ b/tony/tony.sh
@@ -19,17 +19,17 @@ set -x -e
 # TonY settings
 readonly TONY_INSTALL_FOLDER='/opt/tony/'
 readonly TONY_SAMPLES_FOLDER="${TONY_INSTALL_FOLDER}"'/TonY-samples'
-readonly TONY_DEFAULT_VERSION='cc8c4ea599143b775092d1b05af63398ad98a9de'
+readonly TONY_DEFAULT_VERSION='6273f16d9c0597b3715a04645445aeedc495baf8'
 
 # Tony configurations: https://github.com/linkedin/TonY/wiki/TonY-Configurations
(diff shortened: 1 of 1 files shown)


gogasca:
@functicons can you please take a look when you have a chance? Thanks

tony/tony.sh
```diff
@@ -19,17 +19,17 @@ set -x -e
 # TonY settings
 readonly TONY_INSTALL_FOLDER='/opt/tony/'
 readonly TONY_SAMPLES_FOLDER="${TONY_INSTALL_FOLDER}"'/TonY-samples'
-readonly TONY_DEFAULT_VERSION='cc8c4ea599143b775092d1b05af63398ad98a9de'
+readonly TONY_DEFAULT_VERSION='6273f16d9c0597b3715a04645445aeedc495baf8'
 
 # Tony configurations: https://github.com/linkedin/TonY/wiki/TonY-Configurations
 readonly PS_INSTANCES=1
-readonly PS_MEMORY='4g'
+readonly PS_MEMORY='2g'
```
aniket486: Why is it okay to change the defaults? Wouldn't it cause issues with users of this init action?

gogasca: I decided to reduce the memory to save on resources and fit more jobs. 
Currently each sample job requires 1 parameter server (2GB) + 2 workers (4GB * 2) + Application container (2GB) = 12 GB, which fits exactly in one `n1-standard-4 instance` worker node. (15 GB RAM) * 0.80 = 12 GB .

gogasca:
@aniket486 is it ok to merge? Thanks

PR #502, opened 2019-04-11
Make usage of private ip disablable

### LINKED ISSUE #501: Unable to disable use-cloud-sql-private-ip option
The check is 
```bash
readonly use_cloud_sql_private_ip="$(/usr/share/google/get_metadata_value attributes/use-cloud-sql-private-ip || true)
```
It can never be false

Obdolbacca:
Fixes #501


Obdolbacca:
I signed it!

medb:
Thank you for contribution!

PR #524, opened 2019-05-02
Do not install Docker in Datalab init action on worker nodes

medb:
Datalab runs only on master node(s), that's why there no need to install Docker on worker nodes.

### DIFF
--- a/datalab/datalab.sh
+++ b/datalab/datalab.sh
@@ -18,12 +18,6 @@
 
 set -exo pipefail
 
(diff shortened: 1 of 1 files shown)


datalab/datalab.sh
```diff
@@ -18,6 +18,12 @@
 
 set -exo pipefail
 
+# Exit immediatedly if this is not a master node, because
```
functicons: It is more readable to put this check in main().

medb: Done

PR #526, opened 2019-05-07
HBase Kerberos Configuration and RPC protection

ThomasStadje:
At Commerzbank we have developed a new version of HBase init action that supports Kerberos based authentication. This mode can be changed by passing enable-kerberos and keytab-bucket as cluster metadata during cluster creation process. The script automatically changes the necessary configurations and creates all keytabs necessary for HBase. Please have a look at the init action and the changed documentation.

The provision of HBase with Kerberos and HBase RPC protection enables to securely authenticate and encrypt data transfer from HBase clients (e.g. developed via the HBase Java API) to the cluster. 

Contributors:
@dkoernlein
@PhilippePoth
@kkuemmel
@ThomasStadje
@commerzbank/bdaa
@commerzbank

### DIFF
--- a/hbase/README.md
+++ b/hbase/README.md
@@ -41,8 +41,8 @@ On dataproc clusters HBase uses HDFS as storage backend by default. This mode ca
         --num-masters 3 --num-workers 2
     ```
 
-## Configuration of kerberos authentication and rpc encryption for HBase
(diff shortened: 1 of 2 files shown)


ThomasStadje:
I signed it!

aniket486:
Thanks for the contribution, @ThomasStadje ! I'm sure this will be quite useful for a lot of Dataproc users other than Commerzbank.

I have made minor changes to remove requirement of HA (3 masters) cluster.

We can possibly avoid use of keytab-bucket by using decrypted password and adding worker principals from workers directly. I will take a stab at it in follow up PRs.

PR #529, opened 2019-05-12
Initialization actions for RAPIDS libraries 

randerzander:
Creating initialization actions for [RAPIDS](http://rapids.ai/), a suite of GPU accelerated analytics and data science libraries.


randerzander:
I signed it!

rapids/README.md
```diff
@@ -0,0 +1,96 @@
...
+## Using this initialization action
+You can use this initialization action to create a new Dataproc cluster with RAPIDS installed:
+
+1. Using the `gcloud` command to create a new cluster with this initialization action. The following command will create a new cluster named `<CLUSTER_NAME>`.
+
+    ```bash
+    DATAPROC_BUCKET=dataproc-initialization-actions
+
+    gcloud beta dataproc clusters create <CLUSTER_NAME> \
+    --zone us-east1-c \
```
aniket486: Is this option necessary?

rapids/README.md
```diff
@@ -0,0 +1,96 @@
...
+    --master-accelerator type=nvidia-tesla-t4,count=4 \
```
aniket486: does this only work with `nvidia-tesla-t4` type of accelerator? In any case (yes or no), can you clarify that in the documentation below?

Also, note, nvidia-tesla-t4 is not documented as a supported accelerator at: https://cloud.google.com/dataproc/docs/concepts/compute/gpus

rapids/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+RAPIDS is a relatively young project with APIs evolving quickly. If you encounter unexpected errors or have feature requests, please file them at the relevant [RAPIDS repo](https://github.com/rapidsai).
+
+### Options
+
+By default, the master node also runs dask-cuda-workers. This is useful for smaller scale jobs- processes run on 4 GPUs in a single node will usually be more performant than when run on the same number of GPUs in separate server nodes (due to higher communication costs).
+
+If you want to save the master's GPU(s) for other purposes, this behavior is configurable via a metadata key using `--metadata`.
+
+* `master-worker=false` - whether to run dask-cuda-workers on the master node
```
aniket486: can you use more descriptive name here? Something like: `run-cuda-worker-on-master`?

rapids/README.md
```diff
@@ -0,0 +1,96 @@
...
+--worker-accelerator type=nvidia-tesla-t4,count=4 \
+--worker-machine-type n1-standard-32 \
+--metadata "gpu-driver-url=http://us.download.nvidia.com/tesla/410.104/NVIDIA-Linux-x86_64-410.104.run" \
+--initialization-actions gs://$DATAPROC_BUCKET/rapids/install-gpu-driver.sh,gs://$DATAPROC_BUCKET/rapids/install-rapids.sh,gs://$DATAPROC_BUCKET/rapids/launch-dask.sh
+```
+
+## Important notes
+* RAPIDS is only supported on Pascal or newer GPU architectures.
+* You must set a GPU accelerator type for both master and worker nodes, else the GPU driver install will fail and the cluster will report an error state
+* We recommend an n1-standard-32 worker machine type or better to ensure sufficient host-memory for buffering data to and from GPUs.
```
aniket486: During testing I noticed that if you use only 1 accelerator card, you get following error message if n1-standard-32 machine type is used.
`Invalid value for field 'resource.guestAccelerators[0].acceleratorCount': '1'. At most 24 vCPUs can be used along with 1 accelerator cards in an instance..`

rapids/README.md
randerzander: I don't think so, but it's the only region I was able to use that had any T4 accelerator types. Will remove it.

I'll add a link to the [list of GPUs by region](https://cloud.google.com/compute/docs/gpus/) at the bottom of the README.

rapids/README.md
randerzander: RAPIDS also works on Tesla P100 and V100 GPU types. It will not work on K80 types.

The user currently must specify the correct driver URL, as drivers are specific to GPU type. I checked but did not see an environment property init-scripts can access that tells what GPU type is attached. If there is one you can point me to, I can look at updating install-gpu-driver.sh to automatically pick the right one.

rapids/README.md
randerzander: Added comments on this behavior in "Important notes" section

randerzander:
I signed it!

randerzander:
I signed it!

rapids/test_rapids.py
```diff
@@ -0,0 +1,58 @@
...
+
+from parameterized import parameterized
+
+from integration_tests.dataproc_test_case import DataprocTestCase
+
+
+class RapidsTestCase(DataprocTestCase):
+    COMPONENT = 'rapids'
+    #INIT_ACTION = 'gs://dataproc-initialization-actions/rapids/rapids.sh'
+    #METADATA = 'INIT_ACTIONS_REPO=https://github.com/GoogleCloudPlatform/dataproc-initialization-actions.git,INIT_ACTIONS_BRANCH=master'
```
aniket486: Remove these 3 lines to make this ready for submission.

rapids/internal/install-gpu-driver.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+apt-get update
+apt-get install -y pciutils
+export DEBIAN_FRONTEND=noninteractive
+apt-get install -y "linux-headers-$(uname -r)"
+
+# TODO: Detect attached GPU type, autopick recent driver URL
```
aniket486: I'm not sure if there is an easy way of doing this in the VM. `lspci` might help.

rapids/internal/launch-dask.sh
```diff
@@ -0,0 +1,62 @@
...
+readonly DASK_LAUNCHER='/usr/local/bin/launch-dask.sh'
+readonly INIT_SCRIPT='/usr/lib/systemd/system/dask-cluster.service'
+# For Anaconda Dataproc component
+readonly PREFIX='/opt/conda/anaconda/envs/RAPIDS/bin'
+# For miniconda dataproc init-action
+#readonly PREFIX='/opt/conda/bin'
+
+cat << EOF > "${DASK_LAUNCHER}"
+#!/bin/bash
+if [[ "${ROLE}" == 'Master' ]]; then
```
aniket486: I would install 2 different files based on the role instead of if-else inside the script.

i.e
```
  cat << EOF > "${DASK_LAUNCHER}"
# code for master
EOF
else
  cat << EOF > "${DASK_LAUNCHER}"
# code for worker
EOF
fi

File: rapids/internal/launch-dask.sh [hunk 8]
```diff
@@ -0,0 +1,62 @@
...
+Description=Dask Cluster Service
+[Service]
+Type=simple
+Restart=on-failure
+ExecStart=/bin/bash -c 'exec ${DASK_LAUNCHER}'
+[Install]
+WantedBy=multi-user.target
+EOF
+
+chmod a+rw "${INIT_SCRIPT}"
```
aniket486: For my understanding, why do we need `a+rw` here?

rapids/internal/launch-dask.sh
```diff
@@ -0,0 +1,62 @@
...
+
+readonly ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+readonly MASTER=$(/usr/share/google/get_metadata_value attributes/dataproc-master)
+readonly RUN_CUDA_WORKER_ON_MASTER=$(/usr/share/google/get_metadata_value attributes/run-cuda-worker-on-master || echo -n 'true')
+readonly DASK_LAUNCHER='/usr/local/bin/launch-dask.sh'
+readonly INIT_SCRIPT='/usr/lib/systemd/system/dask-cluster.service'
+# For Anaconda Dataproc component
+readonly PREFIX='/opt/conda/anaconda/envs/RAPIDS/bin'
+# For miniconda dataproc init-action
+#readonly PREFIX='/opt/conda/bin'
```
aniket486: I would suggest to support this with a metadata flag or remove these comments. Not sure if those can be helpful for someone trying to use conda, jupyter init-actions instead. (more such places).

rapids/internal/launch-dask.sh
```diff
@@ -0,0 +1,62 @@
...
+chmod a+rw "${INIT_SCRIPT}"
+
+echo "Starting Dask cluster..."
+
+systemctl daemon-reload
+systemctl enable dask-cluster
+systemctl restart dask-cluster
+systemctl status dask-cluster
+
+echo "Dask cluster instantiation successful" >&2
```
aniket486: why `>&2` here? remove?

rapids/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+These initialization actions automate the process of setting up a Dask-cuDF cluster by running the following:
+
+On the Dataproc master node:
+- `dask-scheduler`, and `dask-cuda-worker`
+
+On the Dataproc worker nodes:
+- `dask-cuda-worker`
+
+Our initialization actions do the following:
```
aniket486: Consider rewording to: `The initialization action does the following` since there is only 1 init-action used below.

rapids/internal/install-gpu-driver.sh
randerzander: I confirmed that for all GPU types RAPIDS works with in Dataproc (everything but K80s), the same driver URL works. Removing this comment.

For future reference, re: lspci, I don't believe it gives you enough information to determine card type until the driver is installed.

rapids/internal/launch-dask.sh
randerzander: I followed that pattern from [the existing Jupyter scripts](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/jupyter/internal/launch-jupyter-kernel.sh#L36).

rapids/rapids.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+readonly ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+readonly INIT_ACTIONS_REPO="$(/usr/share/google/get_metadata_value attributes/INIT_ACTIONS_REPO \
+  || echo 'https://github.com/GoogleCloudPlatform/dataproc-initialization-actions.git')"
+readonly INIT_ACTIONS_BRANCH="$(/usr/share/google/get_metadata_value attributes/INIT_ACTIONS_BRANCH \
+  || echo 'master')"
+
+echo "Cloning fresh dataproc-initialization-actions from repo ${INIT_ACTIONS_REPO} and branch ${INIT_ACTIONS_BRANCH}..."
+git clone -b "${INIT_ACTIONS_BRANCH}" --single-branch "${INIT_ACTIONS_REPO}"
```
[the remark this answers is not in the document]
randerzander: I added comments about this in the options section.

rapids/README.md
randerzander: done

rapids/internal/launch-dask.sh
randerzander: good catch, removed, thanks

rapids/internal/launch-dask.sh
randerzander: Comments removed

rapids/internal/launch-dask.sh
randerzander: Done

rapids/test_rapids.py
randerzander: done

PR #541, opened 2019-05-29
Sets up Cloud Build to run our init actions tests on PRs

estebanlau:
Sets up Cloud Build infrastructure for triggering tests through GitHub App Triggers.

The tests triggered depending on the modified init actions.


Dockerfile
```diff
@@ -0,0 +1,20 @@
+
```
medb: Remove extra spaces?

bigtable/test_bigtable.py
```diff
@@ -108,5 +108,4 @@ def test_bigtable(self, configuration, dataproc_version, machine_suffixes):
 
 
 if __name__ == '__main__':
-    unittest.main()
```
medb: Revert this change?

cloudbuild.yaml
```diff
@@ -0,0 +1,111 @@
...
+# Create docker image from regular Dockerfile
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/google.com/hadoop-cloud-dev/init-actions-image', '.']
+# Run tests from the repo, based on the branch name
+- name: 'gcr.io/google.com/hadoop-cloud-dev/init-actions-image'
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    [[ "$BRANCH_NAME" == .*bigtable.* ]] && python bigtable/test_bigtable.py || exit 0
```
medb: Is it possible to make this generic instead of copy/pasting for each test/init-action?

cloudbuild.yaml
```diff
@@ -0,0 +1,111 @@
...
+  - |
+    [[ "$BRANCH_NAME" == .*tez.* ]] && python tez/test_tez.py || exit 0
+- name: 'gcr.io/google.com/hadoop-cloud-dev/init-actions-image'
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    [[ "$BRANCH_NAME" == .*tony.* ]] && python tony/test_tony.py || exit 0
+
+# images:
```
medb: Is this comment necessary?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY . /
+RUN ls
+
+RUN apt-get -y update
+RUN apt-get -y install python3-pip
+
+ENV PATH=$PATH:/builder/google-cloud-sdk/bin/
+ENV PYTHONPATH /
+
+RUN pip install -r requirements.txt
```
medb: Rename `requirements.txt` to `requirements.pip`?

cloudbuild.yaml
```diff
@@ -0,0 +1,111 @@
+steps:
+
+# Create docker image from regular Dockerfile
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/google.com/hadoop-cloud-dev/init-actions-image', '.']
+# Run tests from the repo, based on the branch name
```
medb: Instead of running all test lets run tests only for modified init actions. We may need to create shell or python script that does this and execute it here.

cloudbuild.yaml
```diff
@@ -0,0 +1,111 @@
...
+  args:
+  - '-c'
+  - |
+    [[ "$BRANCH_NAME" == .*solr.* ]] && python solr/test_solr.py || exit 0
+- name: 'gcr.io/google.com/hadoop-cloud-dev/init-actions-image'
+  entrypoint: 'bash'
+  args:
+  - '-c'
+  - |
+    [[ "$BRANCH_NAME" == .*starburst.* ]] && python starburst-presto/test_presto.py || exit 0
```
findepi: Are you intentionally ignoring the exit code from python?

Dockerfile
estebanlau: Done.

bigtable/test_bigtable.py
estebanlau: Done.

cloudbuild.yaml
estebanlau: Removed.

Dockerfile
estebanlau: Changed the extension.

cloudbuild.yaml
estebanlau: I was previously making sure that the steps were all executed in case one of the tests failed. I now changed the step to only fail if the python test fails. If the python test is simply not run if the branch name doesn't match, the step still succeeds.

estebanlau: The 'waitFor' attribute defines the dependencies of each step. In this case, the tests run after the container is built.

cloudbuild.yaml
estebanlau: In here, only certain init actions are run (based on the branch name).

I propose changing this approach to having several defined set of build steps (different yaml files), each representing a given feature to be tested and triggered only when their respective files are modified. The downside is that we'll have many yaml files (one for each feature).

estebanlau: This functionality is supported by GitHub App Triggers Alpha and we would end up with multiple triggers defined in GCP.

medb: I think that this is nor reliable, because this convention is not obvious and requires human attention to enforce branch names.

What we need to do is to have script that diffs PR branch against master branch and after that runs tests for all init actions that modified by PR. Is it feasible to implement this with Cloud Build?

estebanlau: Yes, I've been working on this approach. I can run a different set of steps based on files modified. CB GitHub App Triggers supports this by triggering builds depending on files included in a given PR. I'll have to define each trigger in GCP.

cloudbuild.yaml
estebanlau: Removed this file in place of multiple config files. Instead of deciding whether to run a given test based on the branch name as shown here, now we handle that logic at the trigger level in GCP.

cloudbuild.yaml
medb: Is CB GitHub App Triggers allow to configure this in generic way? i.e if `init-action-folder` is modified run `init-action-folder/test.py`? I'm asking, because I don't want to modify them each time we are adding new init action.

bigtable/bigtable_cloudbuild.yaml
```diff
@@ -0,0 +1,6 @@
+steps:
+- name: 'gcr.io/cloud-builders/docker'
```
medb: I don't think that this is a good approach to copy/paste these files in all init actions if instead we can create a single script that will execute relevant tests automatically.

cloudbuild.yaml
estebanlau: Yes, it allows us to define a generic path (e.g. when bigtable/** changes, we run bigtable/test_bigtable.py). If no changes are made to the files in the init action folder, its test is not triggered. For each init action, a generic rule like this will need to be defined.

bigtable/bigtable_cloudbuild.yaml
estebanlau: Are you suggesting using git diff to determine the modified files instead of using the built-in CB functionality that does this for us? If so, the challenges that come to mind are:

- Authenticating whoever is making the changes
- Determining how to git clone a version of the code that contains the changes we want to test. Git cloning from GoogleCloudPlatform/dataproc-initialization-actions would not contain our changes and depending on the user, the repo name would differ.

estebanlau: I also thought that having every config copying the code seemed inconvenient, but I assumed that usually not many init action changes are included in a given PR (most commonly only 1?) so the building would only happen once for a given PR.

medb: I think we talk past each other, let me describe what I mean with more details:

1. Create a single script that diffs PR changes and runs relevant tests based on them
2. Run this script as entry point of CB integration - I assume that this script will have access  to CB credentials to run tests and will run in git repo context inside CB docker container, so it will have access to git history to diff PR changes and determine what tests it needs to run.

Is this approach not feasible with CB?

cloudbuild.yaml
medb: I guess this is not generic enough comparing to what I thought, because it requires per-folder/init action pattern definition.

bigtable/bigtable_cloudbuild.yaml
estebanlau: It should be possible with CB. I'll look more into this to confirm and work on implementing this approach instead.

estebanlau: CB now uses a single shell script that runs our tests as the entry point.

presubmit.sh
```diff
@@ -1,92 +1,58 @@
...
-	return 1
-}
+declare -a DIRECTORIES_TO_TEST
+for dir in */
+do
+  # skip not init action changes
+  if [[ $dir =~ ^(integration_tests/|util/)$ ]]; then
+    continue
+  fi
+  if contains "$dir"; then
```
estebanlau: Nice shell script wizardry.

Does this account for the edge case where, for instance, starburst-presto is modified and presto is not? In this case, only starburst-presto/test_presto.py should be triggered.

medb: Good catch, modified patter to match prefix from the beginning of the string instead of contains.

medb:
LGTM after @functicons comment addressed.

estebanlau:
@functicons I'm working on moving the CB components within their own directory and including a README.md

estebanlau:
I'm testing CB with the new script. My forked repo is currently integrated with CB for testing the triggers; once I verify that CB runs as expected with the new script, I'll need the credentials to the GoogleCloudPlatform/dataproc-initialization-actions repo to configure it in the google.com:cloud-dataproc-ci project or someone with the necessary credentials could do it.

cloud-build/README.MD
```diff
@@ -0,0 +1,12 @@
+# Continuous Integration via Cloud Build
+
+Cloud Build is configured in our GCP project (google.com:cloud-dataproc-ci) to trigger whenever there are PRs from a fork to the main repository. We are using [GitHub App Triggers Alpha](http://go/gcb-github#github-app-triggers) to integrate with GitHub.
```
functicons: I don't think we want to expose the GCP project name publicly.

cloud-build/presubmit.sh
```diff
@@ -0,0 +1,62 @@
...
+  fi
+  mapfile -t tests_array < <(echo "${tests}")
+  ALL_TESTS+=("${tests_array[@]}")
+done
+echo "Tests: ${ALL_TESTS[*]}"
+
+# Run tests of the init actions that were modified
+python -m unittest "${ALL_TESTS[@]}"
+
+exit $?
```
functicons: Is this line necessary?

cloud-build/presubmit.sh
```diff
@@ -0,0 +1,62 @@
...
+    exit 1
+  fi
+  mapfile -t tests_array < <(echo "${tests}")
+  ALL_TESTS+=("${tests_array[@]}")
+done
+echo "Tests: ${ALL_TESTS[*]}"
+
+# Run tests of the init actions that were modified
+python -m unittest "${ALL_TESTS[@]}"
+
```
functicons: What if ALL_TESTS is empty?

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM gcr.io/cloud-builders/gcloud
```
functicons: Outside the context of this PR, it is not clear what this Dockerfile is for, consider adding comments. Also, why this file is not under cloud-build?

estebanlau: The docker build uses the Dockerfile found at the root of the repo. I tried to place it in cloud-build and use that moved Dockerfile instead by setting the 'dir' attribute of the step, but that made it impossible to access/copy the init actions from the parent directory.

estebanlau: I'm looking more into this.

estebanlau: The Dockerfile is inside now.

cloud-build/presubmit.sh
estebanlau: If ALL_TESTS is empty, this means that no init action files were modified (so no tests will run) and the python command will return success.

medb: Actually, if `ALL_TESTS` is empty all tests will be run. I will add if condition.

cloud-build/cloudbuild.yaml
```diff
@@ -0,0 +1,7 @@
+steps:
+# Create docker image from regular Dockerfile
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/google.com/hadoop-cloud-dev/init-actions-image', '-f', 'cloud-build/Dockerfile', '.']
```
medb: Is it possible to have the same path for `args` and `name` tags?

cloud-build/presubmit.sh
medb: Removed

cloud-build/cloudbuild.yaml
estebanlau: The name refers to the base image that is being used and that has an entrypoint. In this case, gcr.io/cloud-builders/docker contains docker functionality.

cloud-build/README.MD
estebanlau: Removed.

cloud-build/presubmit.sh
```diff
@@ -16,47 +16,45 @@ echo "Changed files: ${CHANGED_FILES[*]}"
...
 done
-echo "Tests: ${ALL_TESTS[*]}"
+echo "Tests: ${TESTS_TO_RUN[*]}"
 
 # Run tests of the init actions that were modified
-python -m unittest "${ALL_TESTS[@]}"
-
-exit $?
+if [[ $RUN_ALL_TESTS == true ]]; then
+  python -m unittest
```
estebanlau: Are you sure this runs all tests? Running this command in my machine results in no tests being run.

medb: Hm, I tested with `fastunit` and it runs all tests from the git repo root. Maybe there are difference between `unittest` and `fastunit`.

Anyway, we need to use `fastunit` because without it it will take hours to run all tests.

May you configure to use `fastunit` here?

cloud-build/cloudbuild.yaml
medb: Sorry for confusion, I meant `name` in a line below:
```
- name: 'gcr.io/google.com/cloud-dataproc-ci/init-actions-image'
```

Can we change `args` line to use `cloud-dataproc-ci` in the path:
```
  args: ['build', '--tag=gcr.io/google.com/cloud-dataproc-ci/init-actions-image', '-f', 'cloud-build/Dockerfile', '.']
```

cloud-build/presubmit.sh
estebanlau: Yes, I will work on it.

cloud-build/cloudbuild.yaml
```diff
@@ -0,0 +1,7 @@
...
+- name: 'gcr.io/google.com/cloud-dataproc-ci/init-actions-image'
```
functicons: In the previous step, the project is hadoop-cloud-dev, but here it is cloud-dataproc-ci, why?

estebanlau: Great catch! 

Wow, I spend hours yesterday debugging why the packages installed in the Dockerfile weren't accessible in the script. This seems to be it, thanks!

cloud-build/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# This Dockerfile spins up a container where presubmit tests are run.
+# Cloud Build orchestrates this process.
+
+FROM gcr.io/cloud-builders/gcloud
+
+# Copy everything into the container
+COPY . /
```
functicons: I guess you want to copy the current project to a non / directory, e.g.: `COPY . /dataproc-init-actions/`

cloud-build/README.MD
```diff
@@ -0,0 +1,12 @@
+# Continuous Integration via Cloud Build
+
+Cloud Build is configured in our GCP project (google.com:cloud-dataproc-ci) to trigger whenever there are PRs from a fork to the main repository. We are using GitHub App Triggers Alpha to integrate with GitHub.
```
functicons: I think we should remove "google.com:cloud-dataproc-ci", that is sensitive information.

cloud-build/Dockerfile
estebanlau: Done.

cloud-build/cloudbuild.yaml
```diff
@@ -0,0 +1,7 @@
+steps:
+# Create docker image from regular Dockerfile
+- name: 'gcr.io/cloud-builders/docker'
+  args: ['build', '--tag=gcr.io/google.com/cloud-dataproc-ci/init-actions-image', '-f', 'cloud-build/Dockerfile', '.']
```
functicons: Seem we can use variable $PROJECT_ID https://cloud.google.com/cloud-build/docs/configuring-builds/substitute-variable-values

estebanlau: $PROJECT_ID would be substituted for "google.com:cloud-dataproc-ci" but we need "google.com/cloud-dataproc-ci" as part of our GCR path. See https://cloud.google.com/container-registry/docs/overview#domain-scoped_projects.

medb: I have asked Andrey to create new project without `google.com:` prefix, maybe let's wait until it will be done.

fastunit/__init__.py
```diff
@@ -0,0 +1,78 @@
+"""
```
medb: May we pull fastunit dependency from GitHub repo, instead of copying it in init actions repo? Maybe we need to create Docker image with pre-installed fastunit?

cloud-build/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# This Dockerfile spins up a container where presubmit tests are run.
+# Cloud Build orchestrates this process.
+
+FROM gcr.io/cloud-dataproc-ci/init-actions-image
```
medb: How this image is created?

cloud-build/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Copy everything into the container
+COPY . init-actions/
+WORKDIR init-actions/
+RUN ls -a
+
+RUN apt-get -y update && \
+    apt-get -y install python3-pip
+
+RUN python3.5 -m pip install --upgrade pip
```
medb: Can we use just `python3` instead?

cloud-build/Dockerfile
estebanlau: I created it by manually running cloud build steps that pushed the image created from the Dockerfile.

cloud-build/Dockerfile
estebanlau: Yes.

cloud-build/Dockerfile
medb: Can we depend on default gcloud image instead? Or this image has some pre-installed software (fastunit, etc)? In general I think that we shouldn't depend on one-off images and instead put all boot strapping logic in this image.

estebanlau: This image has fastunit pre-installed. I fixed the Dockerfile to base itself off the gcloud image and also included the fastunit directory so that it can be readily used.

estebanlau: The downside is that the gcloud image doesn't have fastunit pre-installed so we have to copy the fastunit directory.

cloud-build/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# This Dockerfile spins up a container where presubmit tests are run.
+# Cloud Build orchestrates this process.
+
+FROM gcr.io/cloud-builders/gcloud
+
+# Copy everything into the container
+COPY . init-actions/
```
medb: Cloud Build mounts GitHub repo to `/workspace` directory (current directory here). You need to copy it to new directory only if you are running multiple containers in the same build, otherwise you can just run presubmit in the workspace.

You can safely delete `COPY` and `WORKDIR` commands

fastunit/__init__.py
medb: We can just clone fastunit from GitHub in Docker container instead (without creating an image) by adding this command to `Docker` file and use it in pre-submit:
```bash
WORKDIR /fastunit
RUN git clone <fastunit repo> .
```

This way we will not need to create Docker image with pre-backed fastunit and will not need to include it into the init action repo.

cloud-build/Dockerfile
medb: I was thinking about cloning fastunit repo in Docker file as an alternative - added separate comment for this.

fastunit/__init__.py
estebanlau: Done.

integration_tests/setup.py
```diff
@@ -0,0 +1,31 @@
+from setuptools import setup, find_packages
```
medb: Seems like we don't need this file anymore.

cloudbuild/presubmit.sh
```diff
@@ -0,0 +1,91 @@
...
+is_prefix() {
+  for file in "${CHANGED_FILES[@]}"; do
+    if [[ $file =~ ^$1 ]]; then
+      return 0
+    fi
+  done
+  return 1
+}
+
+determine_tests_to_run() {
```
functicons: Can we add comments about how we detect changes? seems you cloned a repo, then somehow calculated the diff, but I don't understand.

medb: Done.

oozie/test_oozie.py
```diff
@@ -29,10 +29,10 @@ def __run_test_file(self, name):
             ("SINGLE", "1.1", ["m"]),
             ("SINGLE", "1.2", ["m"]),
             ("SINGLE", "1.3", ["m"]),
-            ("STANDARD", "1.1", ["m"]),
+            # ("STANDARD", "1.1", ["m"]),
```
estebanlau: Remove comments.

medb: Done.

cloudbuild/README.MD
```diff
@@ -0,0 +1,15 @@
+# Continuous Integration via Cloud Build
```
functicons: Update this file with the K8S stuff.

medb: Done.

cloudbuild/README.MD
```diff
@@ -0,0 +1,21 @@
...
+[Cloud Build](https://cloud.google.com/cloud-build/) is configured to trigger on
+`/gcbrun` comment from maintainer in PR to the `master` branch.
+
+Cloud Build triggers a set of build steps, configured in a `cloudbuild.yaml`
+config file. These steps are:
+
+1.  Create a presubmit Docker container with initialization actions code and
+    necessary dependencies installed; this includes git, gcloud, fastunit and
+    parametrized.
+1.  Push a presubmit Docker container to
```
functicons: s/a presubmit Docker container/the presubmit Docker container

cloudbuild/README.MD
```diff
@@ -0,0 +1,21 @@
+# Continuous Integration via Cloud Build
```
functicons: Can you mention a bit about how we store/encode and use the SSH key?

cloudbuild/README.MD
medb: Done.

cloudbuild/README.MD
medb: Done.

cloudbuild/kubectl.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
functicons: nit: the filename kubectl.sh is a little bit confusing to me. I'd prefer something like run-presubmit-on-k8s.sh.

medb: Done.

PR #544, opened 2019-06-06
Make cloud sql proxy init action work on kerberized clusters.

mikaylakonst:
For kerberized clusters, the connect string for beeline needs to include the hive principal


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -82,6 +82,53 @@ readonly ADDITIONAL_INSTANCES_KEY='attributes/additional-cloud-sql-instances'
...
+  else
+    return 1
+  fi
+}
+
+readonly KERBEROS_ENABLED=$(is_component_selected 'kerberos' && echo 'true' || echo 'false')
+
+function get_hive_principal() {
+  local host=$(hostname)
+  local domain=$(dnsdomainname)
```
cyxxy: Shouldn't you convert to all upper case?

mikaylakonst: Done
Used fully qualified hostname, made realm upper case

PR #545, opened 2019-06-06
Fixing issues with presto

ChethanUK:
Updated the init script to fix the issues (https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/533)

### DIFF
--- a/starburst-presto/README.MD
+++ b/starburst-presto/README.MD
@@ -11,13 +11,13 @@ You can use this initialization action to create a new Dataproc cluster with Pre
     gcloud dataproc clusters create <CLUSTER_NAME> \
       --initialization-actions gs://dataproc-initialization-actions/starburst-presto/presto.sh
     ```
-1. Once the cluster has been created, Presto is configured to run on port `8080` (though you can change this in the script) on the master node in a Cloud Dataproc cluster. To connect to the Presto web interface, you will need to create an SSH tunnel and use a SOCKS 5 Proxy as described in the [dataproc web interfaces](https://cloud.google.com/dataproc/cluster-web-interfaces) documentation. You can also use the [Presto command line interface](https://docs.starburstdata.com/latest/installation/cli.html) using the `presto` command on the master node.
+1. Once the cluster has been created, Presto is configured to run on port `8060` (though you can change this in the script) on the master node in a Cloud Dataproc cluster. To connect to the Presto web interface, you will need to create an SSH tunnel and use a SOCKS 5 Proxy as described in the [dataproc web interfaces](https://cloud.google.com/dataproc/cluster-web-interfaces) documentation. You can also use the [Presto command line interface](https://docs.starburstdata.com/latest/installation/cli.html) using the `presto` command on the master node.
 
 You can find more information about using initialization actions with Dataproc in the [Dataproc documentation](https://cloud.google.com/dataproc/init-actions).
 
 ## Important notes
 * This script must be updated based on which Presto version you wish to install
 * You may need to adjust the memory settings in `jvm.config` based on your needs
-* Presto is set to use HTTP port `8080` by default
+* Presto is set to use HTTP port `8060` by default
 * Only the Hive connector is configured by default
 * High-Availability configuration is discouraged as coordinator is started only on `m-0` and other master nodes are idle
--- a/starburst-presto/presto.sh
+++ b/starburst-presto/presto.sh
@@ -23,7 +23,7 @@ readonly PRESTO_MASTER_FQDN="$(/usr/share/google/get_metadata_value attributes/d
 readonly WORKER_COUNT=$(/usr/share/google/get_metadata_value attributes/dataproc-worker-count)
 readonly PRESTO_MAJOR_VERSION="302"
 readonly STARBURST_PRESTO_VERSION="302-e.11"
-readonly HTTP_PORT="8080"
+readonly HTTP_PORT="8060"
 readonly INIT_SCRIPT="/usr/lib/systemd/system/presto.service"
 PRESTO_JVM_MB=0;
 PRESTO_QUERY_NODE_MB=0;
@@ -38,14 +38,13 @@ function err() {
 function wait_for_presto_cluster_ready() {
   # wait up to 120s for presto being able to run query
   for ((i = 0; i < 12; i++)); do
-    if presto --execute='select * from system.runtime.nodes;'; then
+    if presto --server="localhost:${HTTP_PORT}" --execute='select 1'; then
       return 0
     fi
     sleep 10
   done
   return 1
 }
-
 
 function get_presto(){
   # Download and unpack Presto server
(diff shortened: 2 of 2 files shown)


starburst-presto/presto.sh
```diff
@@ -23,7 +23,7 @@ readonly PRESTO_MASTER_FQDN="$(/usr/share/google/get_metadata_value attributes/d
 readonly WORKER_COUNT=$(/usr/share/google/get_metadata_value attributes/dataproc-worker-count)
 readonly PRESTO_MAJOR_VERSION="302"
 readonly STARBURST_PRESTO_VERSION="302-e.11"
-readonly HTTP_PORT="8080"
+readonly HTTP_PORT="8070"
```
mattsfuller: If we're going to change from Presto's default port, I would suggestion port 8060 to be consistent with: https://cloud.google.com/dataproc/docs/tutorials/presto-dataproc
> If using the optional component, the Presto server will run on port 8060.

starburst-presto/presto.sh
```diff
@@ -123,23 +128,35 @@ function configure_hive(){
...
 hive.metastore.uri=${metastore_uri}
 EOF
+
+# Add new connectors configs here 
+# For example: Adding jmx properties like
+  cat > presto-server-${PRESTO_VERSION}/etc/catalog/tpch.properties <<- "EOF"
+connector.name=tpch
+tpch.splits-per-node=4
+EOF
+
```
mattsfuller: Great idea. As an optional suggestion, could you add the tpcds connector as well?

starburst-presto/presto.sh
ChethanUK: Okay Updated :)

starburst-presto/presto.sh
```diff
@@ -35,18 +35,23 @@ function err() {
...
+        [ $(presto --server=localhost:${HTTP_PORT} --execute='select 1' | grep -q "1" && echo 0 || echo 1) -eq 0 ]
+}
+
 function wait_for_presto_cluster_ready() {
   # wait up to 120s for presto being able to run query
   for ((i = 0; i < 12; i++)); do
-    if presto --execute='select * from system.runtime.nodes;'; then
-      return 0
-    fi
+    # test_presto_with_query && echo "Presto is working properly." || echo "Presto is not ready, need to wait for presto cluster."
```
mattsfuller: Can you clean this up so there is not commented out code?

starburst-presto/presto.sh
ChethanUK: One will enough right? I added that so users can understand how users can connectors.

If you want I can add jmx and tpcds as well.

starburst-presto/presto.sh
```diff
@@ -123,23 +128,35 @@ function configure_hive(){
 connector.name=hive-hadoop2
 hive.metastore.uri=${metastore_uri}
 EOF
+
+# Add new connectors configs here 
+# For example: Adding jmx properties like
+  cat > presto-server-${PRESTO_VERSION}/etc/catalog/tpch.properties <<- "EOF"
+connector.name=tpch
+tpch.splits-per-node=4
```
findepi: @mattsfuller do you think we need `tpch.splits-per-node=4`? 
tpch connector defaults to number of CPUs which is a reasonable default.

starburst-presto/presto.sh
ChethanUK: Okay Done!.

starburst-presto/presto.sh
```diff
@@ -123,23 +128,35 @@ function configure_hive(){
...
--Xmn512m
--XX:+UseConcMarkSweepGC
--XX:+ExplicitGCInvokesConcurrent
--XX:ReservedCodeCacheSize=150M
+-XX:+UseG1GC
+-XX:-UseBiasedLocking
+-XX:G1HeapRegionSize=32M
+-XX:+UseGCOverheadLimit
 -XX:+ExplicitGCInvokesConcurrent
+-XX:+ExitOnOutOfMemoryError
```
findepi: @ChethanUK these are good changes, thanks for making them.
However, we need some more fine tuning here.
Please extract this to a separate PR and cc me. (Or I can take over the jvm.config changes if you prefer this way.)

starburst-presto/presto.sh
findepi: @ChethanUK what is the problem with 8080 port?
This is the default Presto port. Also, init action for Github Presto is also using 8080:
https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/presto/presto.sh#L31

starburst-presto/presto.sh
```diff
@@ -35,18 +35,21 @@ function err() {
...
+        [ $(presto --server=localhost:${HTTP_PORT} --execute='select 1' | grep -q "1" && echo 0 || echo 1) -eq 0 ]
+}
+
 function wait_for_presto_cluster_ready() {
   # wait up to 120s for presto being able to run query
   for ((i = 0; i < 12; i++)); do
-    if presto --execute='select * from system.runtime.nodes;'; then
-      return 0
-    fi
+    test_presto_with_query && return 0
```
findepi: I think the only real difference between the previous version and the current one is that now you provide the port.

So what about:

```suggestion
    if presto --server="localhost:${HTTP_PORT}" --execute="select 1; then
      return 0
    fi
```

starburst-presto/presto.sh
ChethanUK: No, 8080 as another service, if we use that we can't use other services like Apache Zepplin, etc.. So, It best to have 8060 or some other port other than 8080

starburst-presto/presto.sh
ChethanUK: Cool!, that works as well :) 
Made new commit https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/545/commits/5658bc26283670b3560235c0f67f8e0b2b535ea1

starburst-presto/presto.sh
```diff
@@ -123,23 +122,47 @@ function configure_hive(){
...
+  cat > presto-server/etc/catalog/tpch.properties <<- "EOF"
+connector.name=tpch
+tpch.splits-per-node=4
+EOF
+
+cat > presto-server/etc/catalog/tpcds.properties <<- "EOF"
+connector.name=tpcds
+EOF
+
+cat > presto-server/etc/catalog/jmx.properties <<- "EOF"
```
findepi: nit: please use `<<` rather then `<<-`. The addition of `-` causes stripping of leading whitespace, which is not wanted here.

starburst-presto/presto.sh
```diff
@@ -123,23 +122,47 @@ function configure_hive(){
 connector.name=hive-hadoop2
 hive.metastore.uri=${metastore_uri}
 EOF
+
+# Add new connectors configs here 
+# For example: Adding jmx properties like
```
findepi: It says about "jmx", but the next line is about "tpch". Please remove this line. 
```suggestion
```

starburst-presto/presto.sh
```diff
@@ -123,23 +122,47 @@ function configure_hive(){
 connector.name=hive-hadoop2
 hive.metastore.uri=${metastore_uri}
 EOF
+
+# Add new connectors configs here
```
findepi:
```suggestion
# Add connectors configs here 
```

starburst-presto/presto.sh
findepi: @ChethanUK please remove `tpch.splits-per-node`, let's use defaults.

```suggestion
```

starburst-presto/README.MD
```diff
@@ -18,6 +18,6 @@ You can find more information about using initialization actions with Dataproc i
 ## Important notes
 * This script must be updated based on which Presto version you wish to install
 * You may need to adjust the memory settings in `jvm.config` based on your needs
-* Presto is set to use HTTP port `8080` by default
+* Presto is set to use HTTP port `8070` by default
```
findepi: Default in code is 8060:

```suggestion
* Presto is set to use HTTP port `8060` by default
```

findepi: @mattsfuller are we OK changing the default?

@ChethanUK if 8080 is a wrong default, we should change this for both presto and starburst-presto the same way. This may be one commit for both.

starburst-presto/presto.sh
findepi: @ChethanUK as we spoke offline -- i extracted this to https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/546 and update to be in line with current recommended options. 
I also noticed you added `-Djdk.attach.allowAttachSelf=true`  which has not been recommended yet, so I created https://github.com/prestosql/presto/pull/937.

I suggest that you revert your changes here to avoid conflicts.

starburst-presto/README.MD
mattsfuller: I don't think there is a wrong default. It's conflicting with another optional component that is also using 8080. The same argument could be made that the other thing is using the wrong default.

I think the best solution is to make the port number configurable and passed in to override the default.

starburst-presto/presto.sh
mattsfuller: I don't think it's needed. I agree with the suggested change.

starburst-presto/README.MD
ChethanUK: > both presto and starburst-presto the same way.

Okay, Will update in presto as well :)

starburst-presto/presto.sh
ChethanUK: Okay!

starburst-presto/presto.sh
mattsfuller: one is fine. It was only a suggestion since TPCH and TPCDS seem natural to go together and didn't seem like a ton of work to add both.

medb:
This looks like a duplicate of existing 2 PRs:
https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/592
https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/597

PR #546, opened 2019-06-06
Update Presto jvm.config

findepi:
Update the `jvm.config` in accordance with the current recommended JVM
configuration for Presto.

cc @mattsfuller @ChethanUK


findepi:
@aniket486 could you please take a look?

PR #549, opened 2019-06-08
Fix lint warnings in Presto init action

medb:


presto/presto.sh
```diff
@@ -18,6 +18,8 @@ set -euxo pipefail
 # Use Python from /usr/bin instead of /opt/conda.
 export PATH=/usr/bin:$PATH
 
+readonly PRESTO_BASE_URL=https://repo1.maven.org/maven2/com/facebook/presto
```
findepi: I'd suggest move this just before `readonly PRESTO_VERSION='

medb: Done. Good point, thanks!

medb:
@findepi I was using newly introduced IntelliJ Shell linting support from recently released EAP:
https://blog.jetbrains.com/idea/2019/06/intellij-idea-2019-2-eap2-shell-script-support-improved-code-duplicates-detection-services-tool-window-and-more/

IntelliJ relies on [Shellcheck](https://github.com/koalaman/shellcheck) that we can configure for our presubmit when we have [CI](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/541):
https://github.com/koalaman/shellcheck#in-your-build-or-test-suites

At the moment we don't have automation around this.

PR #554, opened 2019-06-10
Fix Cloud SQL Proxy integration test

medb:
Fixes failure in test when only GCE zone is set:
```
======================================================================
FAIL: test_cloud_sql_proxy.mode: STANDARD.version: 1_3.random_prefix: nk85 (cloud-sql-proxy.test_cloud_sql.CloudSqlProxyTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/google/home/idv/dev/src/dataproc-initialization-actions/cloud-sql-proxy/test_cloud_sql.py", line 33, in setUp
    self.DB_NAME, stderr))
AssertionError: 1 != 0 : Failed to create sql instance test-cloud-sql-20190608-213716. Last error: ERROR: (gcloud.sql.instances.create) HTTPError 400: Invalid request: Invalid value for region: global.
```

### DIFF
--- a/cloud-sql-proxy/test_cloud_sql.py
+++ b/cloud-sql-proxy/test_cloud_sql.py
@@ -31,22 +31,24 @@ class CloudSqlProxyTestCase(DataprocTestCase):
             self.DB_NAME, self.REGION, "--async --format=json")
         ret_code, stdout, stderr = self.run_command(cmd)
         operation_id = json.loads(stdout.strip())['name']
+        self.wait_cloud_sql_operation(operation_id)
+
+    def tearDown(self):
+        super().tearDown()
+        cmd = 'gcloud sql instances delete {} --async --format=json'.format(
+            self.DB_NAME)
+        ret_code, stdout, stderr = self.run_command(cmd)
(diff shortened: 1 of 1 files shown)


cloud-sql-proxy/test_cloud_sql.py
```diff
@@ -16,21 +17,27 @@ def setUpClass(cls):
...
+        operation_id = json.loads(stdout.strip())['name']
+        cmd = 'gcloud sql operations wait {} --timeout=600'.format(
+            operation_id)
+        ret_code, stdout, stderr = self.run_command(cmd)
         self.assertEqual(
-            ret_code, 0,
-            "Failed to create sql instance {}. Last error: {}".format(
-                self.DB_NAME, stderr))
+            ret_code, 0, "Failed to create sql instance {}.".format(
+                self.DB_NAME,
```
functicons: This format() has 2 args, but only 1 placeholder {}, why?

medb: Fixed

PR #562, opened 2019-06-12
Download RAPIDS init action dependencies from GCS

medb:


rapids/rapids.sh
```diff
@@ -1,23 +1,27 @@
...
 
-echo "Cloning dataproc-initialization-actions from repo ${INIT_ACTIONS_REPO} and branch ${INIT_ACTIONS_BRANCH}..."
-git clone -b "${INIT_ACTIONS_BRANCH}" --single-branch "${INIT_ACTIONS_REPO}"
+readonly DEAFULT_INIT_ACTIONS_REPO=gs://dataproc-initialization-actions
+readonly INIT_ACTIONS_REPO="$(/usr/share/google/get_metadata_value attributes/INIT_ACTIONS_REPO ||
+  echo ${DEAFULT_INIT_ACTIONS_REPO})"
+readonly LOCAL_INIT_ACTIONS_REPO=/tmp/local-initialization-actions
 
-./dataproc-initialization-actions/rapids/internal/install-gpu-driver.sh
+echo "Cloning initialization actions from '${INIT_ACTIONS_REPO}' repo..."
```
aniket486: As discussed offline, can we download only the parts we need for running this init action.
In other words, set `INIT_ACTIONS_DIR` to `gs://dataproc-initialization-actions/rapids`.

Note, since you are removing a metadata flag, I would also suggest to rename `INIT_ACTIONS_REPO` to `INIT_ACTIONS_DIR`.

medb: Done.

Didn't rename `INIT_ACTIONS_REPO` because it simpler when user need to use one value to configure multiple init actions.

PR #564, opened 2019-06-12
Download Jupyter init action dependencies from GCS

medb:


jupyter/internal/setup-jupyter-kernel.sh
```diff
@@ -2,7 +2,7 @@
 
 # This script configures Jupyter
 
-set -e
+set -euo pipefail
```
functicons: Why don't enable -x?

jupyter/internal/setup-jupyter-kernel.sh
```diff
@@ -15,30 +15,34 @@ JUPYTER_PORT=$(/usr/share/google/get_metadata_value attributes/JUPYTER_PORT || t
...
-echo "c.NotebookApp.ip = '0.0.0.0'" >> ~/.jupyter/jupyter_notebook_config.py
-echo "c.NotebookApp.open_browser = False" >> ~/.jupyter/jupyter_notebook_config.py
-echo "c.NotebookApp.port = ${JUPYTER_PORT}" >> ~/.jupyter/jupyter_notebook_config.py
-echo "c.NotebookApp.contents_manager_class = 'jgscm.GoogleStorageContentManager'" >> ~/.jupyter/jupyter_notebook_config.py
-echo "c.GoogleStorageContentManager.default_path = '${NOTEBOOK_DIR}'" >> ~/.jupyter/jupyter_notebook_config.py
-echo "c.NotebookApp.token = u'${JUPYTER_AUTH_TOKEN}'" >> ~/.jupyter/jupyter_notebook_config.py
+
+cat <<EOF >>~/.jupyter/jupyter_notebook_config.py
+c.Application.log_level = 'DEBUG'
+c.NotebookApp.ip = '0.0.0.0'
```
functicons: Add an empty line before this line to handle the case that there is no newline character at the end of that file.

jupyter/README.MD
```diff
@@ -28,8 +28,7 @@ There are various options for customizing your Jupyter installation. These can b
 
 * `JUPYTER_PORT`=<integer>: Port on which the Jupyter server runs
 * `JUPYTER_CONDA_PACKAGES`=<colon-separated list of strings>: List of Conda packages to install.
-* `INIT_ACTIONS_REPO`=<https url>: Repo to clone to find other scripts to install/configure Conda and Jupyter.
-* `INIT_ACTIONS_BRANCH`=<string>: Branch in `INIT_ACTIONS_REPO` to use.
+* `INIT_ACTIONS_REPO`=<bucket>: GCS bucket to find other scripts to install/configure Conda and Jupyter.
```
functicons: I think we need to do it in a backwards compatible way. Instead of replace Github with GCS, we support GCS as another type of repo, and make it default.

jupyter/internal/setup-jupyter-kernel.sh
medb: Done.

jupyter/README.MD
medb: Preserved backward compatibility

jupyter/internal/setup-jupyter-kernel.sh
medb: Done.

PR #567, opened 2019-06-12
Ignore temp files

medb:


push-to-gcs.sh
```diff
@@ -43,14 +43,14 @@ if [[ ! -d "${MODULE}" ]]; then
 fi
 
 # Verify shell scripts have permission mode 75x.
-for file in ${MODULE}/*.sh; do
-  if [[ "$(stat -c '%a' ${file})" != 75* ]]; then
-    echo "The permission mode of script ${file} is $(stat -c '%a' ${file}), expected: 75x."
+for file in "${MODULE}/"*.sh; do
+  if [[ "$(stat -c '%a' "${file}")" != 75* ]]; then
```
functicons: Extract this as a variable.

medb: Done.

PR #585, opened 2019-07-08
Delete clusters asynchronously

medb:


integration_tests/dataproc_test_case.py
```diff
@@ -128,8 +129,12 @@ def stage_init_actions(self):
         return staging_dir
 
     def tearDown(self):
-        self.run_and_assert_command(
-            "gcloud dataproc clusters delete {} --quiet".format(self.name))
+        ret_code, _, stderr = self.run_command(
+            "gcloud dataproc clusters delete {} --quiet --async".format(
+                self.name))
+        if ret_code != 0:
```
functicons: What if the async operator fails?

medb: We leave it up to Dataproc, in general Dataproc guarantees that cluster will be deleted.

PR #586, opened 2019-07-08
Fail Drill validate script if Drill UI did not start

medb:
Prevents Drill test from hanging indefinitely if Drill UI failed to start.


drill/validate.sh
```diff
@@ -82,5 +90,5 @@ if ((ui_result == 200 && sql_result == 0)); then
   exit 0
 fi
 
-echo "Test failed"
+echo "  failed"
```
cyxxy: reason you want to remove the word "Test" and put two empty spaces?

medb: By mistake, reverted.

PR #591, opened 2019-07-10
Fix logic that detects changed files in presubmit.

medb:


cloudbuild/presubmit.sh
```diff
@@ -38,7 +38,7 @@ initialize_git_repo() {
 }
 
 # Determines whether a given string is a prefix string of any changed file name
-is_prefix() {
+is_changed_prefix() {
   for file in "${CHANGED_FILES[@]}"; do
     if [[ $file =~ ^$1 ]]; then
```
functicons: This condition is hard to understand, add comment. Also assign $1 to a meaningful local variable.

cloudbuild/presubmit.sh
```diff
@@ -38,7 +38,7 @@ initialize_git_repo() {
 }
 
 # Determines whether a given string is a prefix string of any changed file name
-is_prefix() {
+is_changed_prefix() {
```
functicons: Instead of enumerating all directories and see if they include changed files, can we find these directories directly from changed files? I think in this way, we can avoid recursion.

medb: Done.

cloudbuild/presubmit.sh
medb: Deleted this function.

PR #596, opened 2019-07-10
Declare TESTS_TO_RUN as a global variable

medb:
Fixes "TESTS_TO_RUN[@]: unbound variable" error

### DIFF
--- a/cloudbuild/presubmit.sh
+++ b/cloudbuild/presubmit.sh
@@ -67,7 +67,6 @@ determine_tests_to_run() {
   echo "Changed directories: ${changed_dirs[*]}"
 
   # Determines what tests in changed init action directories to run
-  declare -g -a TESTS_TO_RUN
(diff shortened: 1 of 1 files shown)


cloudbuild/presubmit.sh
```diff
@@ -67,7 +67,7 @@ determine_tests_to_run() {
   echo "Changed directories: ${changed_dirs[*]}"
 
   # Determines what tests in changed init action directories to run
-  declare -a TESTS_TO_RUN
+  declare -g -a TESTS_TO_RUN
```
functicons: Why put it inside of the function?

medb: This is how function passes result.

Do you think will be better to declare it as a global variable at the top of the script?

functicons: Yes.

medb: Done

PR #606, opened 2019-07-25
Replace dataproc-initialization-actions with <my-bucket> in README

functicons:


beam/README.md
```diff
@@ -111,9 +111,9 @@ shown later).
 
 ```bash
 CLUSTER_NAME="$1
-INIT_ACTIONS="gs://dataproc-initialization-actions/docker/docker.sh"
-INIT_ACTIONS+=",gs://dataproc-initialization-actions/flink/flink.sh"
-INIT_ACTIONS+=",gs://dataproc-initialization-actions/beam/beam.sh"
+INIT_ACTIONS="gs://<my-bucket>/docker/docker.sh"
```
medb: Replace with `$BUCKET` variable so it will be copy-pastable?

findepi:
@functicons in fact, why?

functicons:
> @functicons in fact, why?

We want to discourage users referencing gs://dataproc-initialization-actions directly, because it is changing, we have broken users in the past.

PR #613, opened 2019-08-14
Update presto version to 0.224

dennishuo:


dennishuo:
For posterity and thanks to @pmkc for analysis, updated versions of OpenJDK particularly in Ubuntu (and to be in upcoming Dataproc image versions) modify the java.vendor property in a way that breaks some things that rely on that Java property: https://bugs.launchpad.net/ubuntu/+source/openjdk-lts/+bug/1822839

Presto is one such affected component. It removes the java.vendor check in https://github.com/prestodb/presto/pull/11396/files which is present in 0.211 and newer: https://prestodb.github.io/docs/current/release/release-0.211.html

Anyone with private copies of the presto init action experiencing issues with newer Ubuntu-based dataproc images should update the presto version per this commit.

PR #623, opened 2019-09-11
Datalab: remove --allow-unauthenticated

Geander:
Removed --allow-unauthenticated (workaround for expired gpg key), because the expired gpg key problem is fixed.


Geander:
@googlebot I signed it!

PR #627, opened 2019-09-27
Add Spark XGBoost4J to RAPIDS init action

mengdong:
This pull request add dataproc init script for rapids spark xgboost4j. https://github.com/rapidsai/spark-examples. It installs NVIDIA driver, CUDA toolkit and NCCL library on dataproc GPU worker nodes. This is different from rapids script as GPU is not needed for master node and there is less software dependency.

The script is customizable for dataproc linux OS version, GPU driver version, CUDA toolkit version and NCCL version. But the prerequisite need to be met as indicated in README


mengdong:
I think debian is still broken with https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/issues/626

But ubuntu18 should work.

spark-gpu/rapids.sh
```diff
@@ -0,0 +1,35 @@
...
+readonly INIT_ACTIONS_REPO="$(/usr/share/google/get_metadata_value attributes/INIT_ACTIONS_REPO ||
+  echo ${DEAFULT_INIT_ACTIONS_REPO})"
+
+echo "Cloning RAPIDS initialization action from '${INIT_ACTIONS_REPO}' ..."
+RAPIDS_INIT_ACTION_DIR=$(mktemp -d -t rapids-init-action-XXXX)
+readonly RAPIDS_INIT_ACTION_DIR
+gsutil -m rsync -r "${INIT_ACTIONS_REPO}/spark-gpu" "${RAPIDS_INIT_ACTION_DIR}"
+
+if [[ "${LINUX_DIST}" == 'ubuntu' ]]; then
+  mv "${RAPIDS_INIT_ACTION_DIR}/internal/install-gpu-driver-ubuntu.sh" "${RAPIDS_INIT_ACTION_DIR}/internal/install-gpu-driver.sh"
```
medb: Is it possible to reuse [GPU init action](https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/tree/master/gpu) instead of custom GPU drivers install scripts?

spark-gpu/rapids.sh
```diff
@@ -0,0 +1,35 @@
...
+else
+  mv "${RAPIDS_INIT_ACTION_DIR}/internal/install-gpu-driver-debian.sh" "${RAPIDS_INIT_ACTION_DIR}/internal/install-gpu-driver.sh"
+fi
+find "${RAPIDS_INIT_ACTION_DIR}" -name '*.sh' -exec chmod +x {} \;
+
+if [[ "${ROLE}" != 'Master' ]]; then
+  # Ensure we have GPU drivers installed.
+  "${RAPIDS_INIT_ACTION_DIR}/internal/install-gpu-driver.sh"
+else
+  wget "https://rapidsai-data.s3.us-east-2.amazonaws.com/spark/mortgage.zip"
```
medb: Is there an official distribution in more authoritative place? Preferably versioned.

spark-gpu/rapids.sh
```diff
@@ -0,0 +1,35 @@
...
+  unzip "mortgage.zip"
+  hadoop fs -mkdir "/tmp/xgboost4j_spark"
+  hadoop fs -copyFromLocal mortgage "/tmp/xgboost4j_spark/mortgage"
```
medb: Please, add test case for this init action.

spark-gpu/rapids.sh
mengdong: There are a few additional dependencies required like NCCL and rapids spark libraries. Also updated install-gpu-driver-ubuntu.sh is NVIDIA's recommended route to install Tesla drivers.

spark-gpu/rapids.sh
mengdong: https://github.com/rapidsai/spark-examples/tree/master/getting-started-guides/csp/gcp contains the same script for dataproc initilization

spark-gpu/rapids.sh
mengdong: I have removed mortgage data related scripts.

mengdong:
@medb Sorry for the long delay. I have made additional modification and updates, please refer to https://github.com/rapidsai/spark-examples/tree/master/getting-started-guides/csp/gcp for the same script, we can point that to dataproc initiation script here once this checked in. Thanks!

mengdong:
Close this pull request to fix CLA

mengdong:
@medb  Hello Igor, any further update on this pull request? Thanks!

mengdong:
Hello, ping again, any action item on our side?

medb:
@mengdong I have fixed rebase on the master branch in this PR.

Please, add a test so it's possible to validate this init action.

Here is an [example test](https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/gpu/test_gpu.py) for GPU drivers installation initialization action.

You can run it with Bazel:
```bash
bazel test --test_arg=--image_version=1.4-debian9 gpu:test_gpu
```

mengdong:
Hi @medb, I have added the test cases, since the package contains a `-`, I have to modify the base BUILD as well. Please let me know if the test cases works. Thanks for your work and pointing me to the right example.

spark-gpu/test_spark_gpu.py
```diff
@@ -0,0 +1,27 @@
...
+    def test_spark_gpu(self, configuration, machine_suffixes, worker_accelerator):
+        init_actions = self.INIT_ACTIONS
+
+        self.createCluster(
+            configuration,
+            init_actions,
+            beta=True,
+            worker_accelerator=worker_accelerator)
+        for machine_suffix in machine_suffixes:
+            self.verify_instance("{}-{}".format(self.getClusterName(),
```
medb: You need to implement `verify_instance` method that doing actual verification of the installed software

mengdong:
BTW, this is the test command: `bazel test :test_spark_gpu --test_arg=--image_version=1.4-ubuntu18`

mengdong:
The internal folder inside spark-gpu need to be hosted in a GCS location, the default is dataproc-initialization-actions, in my case, I added metadata in test function (note that dongm-gcp-shared is my private GCS location, the script should be hosted in dataproc init GCS location, i.e the default one)
```
        self.createCluster(
            configuration,
            init_actions,
            metadata='INIT_ACTIONS_REPO="gs://dongm-gcp-shared",linux-dist="ubuntu",GCS_BUCKET="gs://dongm-gcp-shared"',
            beta=True,
            worker_accelerator=worker_accelerator)
```

spark-gpu/rapids.sh
medb: I have added support for NVIDIA-provided GPU drivers and CUDA to GPU init action. Please update this PR to use it.

medb: Done.

mengdong:
> @mengdong I have re-factored this PR to add NCCL and XGBoost4J installation into RAPIDS init action, instead of creating a new init action.
> 
> May you update RAPIDS [README.md](/rapids/README.md) to reflect this? And add Spark test that exercises XBoost4J to `test_rapids.py` (similar to [`verify_rapids.py`](/rapids/verify_rapids.py)), to make sure that it works?

Hello @medb, are we certain that this will work? Please note the Rapids spark XGBoost4J: 1, depends on spark but does not depends on Dask. 2, in a spark cluster, the master node doesn't require a GPU 3, the rapids init script relies on conda to install CUDA, whereas Rapids spark can be ran on Scala/Java and doesn't requires conda.

medb:
> Hello @medb, are we certain that this will work? Please note the Rapids spark XGBoost4J: 1, depends on spark but does not depends on Dask. 2, in a spark cluster, the master node doesn't require a GPU 3, the rapids init script relies on conda to install CUDA, whereas Rapids spark can be ran on Scala/Java and doesn't requires conda.

I'm not sure, but is it possible that some users would want to use RAPIDS in Python and Scala/Java on the same cluster? If not, then we can add a required `--rapids-impl=<CONDA|JAVA>` metadata property, that will force users to chose only one of them. If this is a valid use case that users could want both of them installed on the same cluster, then we can add a third value `ALL` for this property that will activate both of them. WDYT?

mengdong:
Hi @medb, sorry I am en route back to US, response is delayed. I think python/scala/java is less of a concern, user can use them all. The problem is coexisting of dask and spark. Both of them are excellent tools but if user provision a spark cluster, they won't need dask, right?

medb:
@mengdong on Dataproc, Spark is executed on YARN cluster. YARN cluster is automatically created by Dataproc and there are no way to disable it. But Spark executors are not running until user will submit Spark job or start Spark Shell - so all memory on cluster nodes are available for user workloads. This means that even with current RAPIDS init action implementation Dask cluster is always running alongside YARN cluster.

In theory, it could create problems, because YARN and Dask will step on each-other toes if user will start to run Spark and Dask workloads at the same time, but we usually assume that users will not do this.

Do Dask workers consume all VM memory after start even if no jobs are submitted? Or memory allocated only when workload on Dask is submitted?

But you are right, it seems that it make sense to add `--metada rapids-runtime=<DASK|SPARK>` flag to allow users to chose how do they want to run RAPIDS workloads.

medb:
This PR is almost ready, few items to address so we can submit it:

1. Add `--metada rapids-runtime=<DASK|SPARK|ALL>` flag to allow users to chose how they want to run RAPIDS workloads.
1. Add configuration and XGBost4J details to RAPIDS [README.md](/rapids/README.md).
1. Add Spark XBoost4J test to `test_rapids.py` (similar to [`verify_rapids.py`](/rapids/verify_rapids.py)).

mengdong:
@medb okay, let me spent some time this week to address those items. Thanks for the clean-up on https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/gpu/install_gpu_driver.sh, I will see if I can leverage it as well.

mengdong:
I have done the following:
> 1. Add `--metada rapids-runtime=<DASK|SPARK|ALL>` flag to allow users to chose how they want to run RAPIDS workloads.
I let the users choose to use `SPARK` or `DASK` with `--metadata`, since Dask cluster already had spark installed by default, I didn't include the `ALL` option.
> 2. Add configuration and XGBost4J details to RAPIDS [README.md](/rapids/README.md).
I have combined README and made some updates
> 3. Add Spark XBoost4J test to `test_rapids.py` (similar to [`verify_rapids.py`](/rapids/verify_rapids.py)). 
I have combined the test cases, I didn't add `verify_spark_rapids.py` because current spark-based rapids can only read data from csv files, I cannot manually create a dataframe and pass that to GPU memory. Add a csv file in initialization action seems too much and that need to used in a spark application. I will add `verify_spark_rapids.py` when spark 3.0 is available and we can move a dataframe to GPU memory.

mengdong:
@medb any feedback on the code change? Thanks

medb:
> @medb any feedback on the code change? Thanks

LGTM, I plan to merge it today/tomorrow.

mengdong:
Thanks, I will keep things updated once spark 3.0 becomes available on Dataproc.

PR #633, opened 2019-10-03
Restart Dataproc Agent with SIGKILL signal after connectors update.

medb:
Restarting Dataproc Agent using SIGTERM signal could cause cluster node loss.


connectors/connectors.sh
```diff
@@ -20,6 +20,16 @@ NEW_NAME_MIN_CONNECTOR_VERSIONS=(
 BIGQUERY_CONNECTOR_VERSION=$(/usr/share/google/get_metadata_value attributes/bigquery-connector-version || true)
 GCS_CONNECTOR_VERSION=$(/usr/share/google/get_metadata_value attributes/gcs-connector-version || true)
 
+is_worker() {
+  local role
+  role="$(/usr/share/google/get_metadata_value attributes/dataproc-role || true)"
+  if [[ $role != Master ]]; then
```
aniket486: shouldn't Master be quoted string here?

medb: No, in `[[` braces it's not necessary because it auto-quotes arguments.

PR #636, opened 2019-10-08
Update presubmit script to avoid failing on missing tests

aniket486:


cloudbuild/presubmit.sh
```diff
@@ -70,8 +70,8 @@ determine_tests_to_run() {
   for changed_dir in "${changed_dirs[@]}"; do
     local tests_in_dir
     if ! tests_in_dir=$(compgen -G "${changed_dir}test*.py"); then
-      echo "ERROR: presubmit failed - cannot find tests inside '${changed_dir}' directory"
-      exit 1
+      echo "WARNING: cannot find tests inside '${changed_dir}' directory"
```
medb: If there are no tests in PR why to run presubmit then?

medb: Discussed offline: the point is not to fail presubmit if there were upstream changes to the init actions without tests.

medb:
This PR superseded by https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/640

PR #637, opened 2019-10-08
Do not restart any services if GCS connector weren't updated

medb:


connectors/connectors.sh
```diff
@@ -20,6 +20,18 @@ NEW_NAME_MIN_CONNECTOR_VERSIONS=(
...
 GCS_CONNECTOR_VERSION=$(/usr/share/google/get_metadata_value attributes/gcs-connector-version || true)
 
+UPDATED_GCS_CONNECTOR=false
+
+is_worker() {
+  local role
+  role="$(/usr/share/google/get_metadata_value attributes/dataproc-role || true)"
+  if [[ $role != Master ]]; then
+    true
+  else
```
functicons: It is more common to "return 0", true doesn't return from the function, it is error-prone.

connectors/connectors.sh
```diff
@@ -20,6 +20,18 @@ NEW_NAME_MIN_CONNECTOR_VERSIONS=(
...
+    false
```
functicons: ditto.

connectors/connectors.sh
```diff
@@ -85,6 +101,16 @@ fi
 update_connector "bigquery" "$BIGQUERY_CONNECTOR_VERSION"
 update_connector "gcs" "$GCS_CONNECTOR_VERSION"
 
+if [[ $UPDATED_GCS_CONNECTOR != true ]]; then
```
functicons: It is better to quote it as "true", otherwise it might be considered as a command.

connectors/connectors.sh
```diff
@@ -85,6 +101,16 @@ fi
...
+  echo "GCS connector wasn't updated - no need to restart any services"
+  exit 0
+fi
+
+# Restart YARN NodeManager service on worker nodes so they can pick up updated GCS connector
+if is_worker; then
+  systemctl kill -s KILL hadoop-yarn-nodemanager
```
functicons: Why don't use `systemctl restart`?

connectors/connectors.sh
medb: Done.

connectors/connectors.sh
medb: Done.

connectors/connectors.sh
medb: `systemctl restart` will restart process with TERM signal that will cause graceful shutdown during which NM will tries to unregister - I want to avoid this.

connectors/connectors.sh
medb: It's not necessary in `[[` braces.

connectors/connectors.sh
```diff
@@ -85,6 +100,16 @@ fi
 update_connector "bigquery" "$BIGQUERY_CONNECTOR_VERSION"
 update_connector "gcs" "$GCS_CONNECTOR_VERSION"
 
+if [[ $UPDATED_GCS_CONNECTOR != true ]]; then
+  echo "GCS connector wasn't updated - no need to restart any services"
+  exit 0
```
pmkc: you might just want this to be a return in case the caller expects something after this function

medb: This is not a function,

PR #638, opened 2019-10-08
Make PIP init action work for Ubuntu

functicons:


python/pip-install.sh
```diff
@@ -4,13 +4,28 @@ set -exo pipefail
 
 readonly PACKAGES=$(/usr/share/google/get_metadata_value attributes/PIP_PACKAGES || true)
 
+function install_pip() {
+  if command -v pip >/dev/null; then
+    echo "pip is already installed."
```
medb: Return from here?

PR #642, opened 2019-10-11
Set up Bazel build system to outsource parameterization

estebanlau:
This change allows us to run tests from the command line with specific configurations that developers might want to test. Example: bazel test solr:test_solr --test_arg=--params="SINGLE 1.3 m" --test_arg=--params="STANDARD 1.2 m"

If no params are specified, then the default set of configurations is used.

More details here: https://docs.google.com/document/d/1-NXCdXnJ7RnORiZu2nN05xf3kUomFQruJKhqPNuL2n8/edit?usp=sharing


bigtable/test_bigtable.py
```diff
@@ -45,6 +50,29 @@ def tearDown(self):
         self.assert_command('gcloud bigtable instances delete {}'.format(
             self.db_name))
 
+    def buildParameters():
```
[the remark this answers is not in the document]
estebanlau: What should we use as the default? 1.3?

Also, should the test take in only one image_version and then use the params with that given version? I'm asking because we currently run sets of tests containing different image versions.

If the answer is to take in several image_versions for testing more than one image_version, then which combinations should we be testing since there is no consistent format? 

For example, test_rstudio.py tests:
                ("SINGLE", "1.0", "rstudio", "password"),
                ("SINGLE", "1.1", "rstudio", "password"),
                ("SINGLE", "1.2", "rstudio", "password"),
                ("SINGLE", "1.3", "rstudio", "password"),
                ("SINGLE", "1.3", "", "password"),  # default username
                ("SINGLE", "1.3", "rstudio", ""),  # no auth
                ("SINGLE", "1.3", "", ""),  # default username and no auth
                ("SINGLE", "1.4", "rstudio", "password")

If we provided image_version=[1.3, 1.4] with params=[("SINGLE", "rstudio", "password"), ("SINGLE", "", password)], is it ok to test all the combinations? That is, [("SINGLE", "1.3", "rstudio", "password"), ("SINGLE", "1.3","", password), ("SINGLE", "1.4", "rstudio", "password"), ("SINGLE", "1.4","", password)].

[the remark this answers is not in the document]
medb: Yes, by default we can set it to `1.3`.

I think that we don't need even to have support for multiple versions in `image_version` flag, it will be simpler just to run separate build for each version track when necessary instead.

absl/BUILD
```diff
@@ -0,0 +1,118 @@
+licenses(["notice"])  # Apache 2.0
```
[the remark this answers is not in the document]
medb: Seems that there are even simpler way to depend on `absl-py` in Bazel:
https://github.com/abseil/abseil-py/issues/37#issue-300799969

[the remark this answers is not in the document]
estebanlau: Ah nice, I implemented the Bazel PIP way but will change it to this one :)

bigtable/test_bigtable.py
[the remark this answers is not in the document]
estebanlau: Ok, cool.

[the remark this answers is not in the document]
estebanlau: Done.

absl/BUILD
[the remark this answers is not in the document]
estebanlau: Done.

bigtable/test_bigtable.py
```diff
@@ -65,25 +90,21 @@ def verify_instance(self, name):
...
-            ("SINGLE", "1.2", ["m"]),
-            ("STANDARD", "1.2", ["m"]),
-            ("HA", "1.2", ["m-0"]),
-            ("SINGLE", "1.3", ["m"]),
-            ("STANDARD", "1.3", ["m"]),
-            ("HA", "1.3", ["m-0"]),
-        ],
-        testcase_func_name=DataprocTestCase.generate_verbose_test_name)
-    def test_bigtable(self, configuration, dataproc_version, machine_suffixes):
+        buildParameters(),
```
medb: After removing `dataproc_version` from the parameters we may not need to override them, because they are fairly test specific and more or less static.

Let's remove `params` flag from the test and just use default parameters directly here and in all other tests.

estebanlau: Done. I also updated the design doc: https://docs.google.com/document/d/1-NXCdXnJ7RnORiZu2nN05xf3kUomFQruJKhqPNuL2n8/edit#

bigtable/test_bigtable.py
```diff
@@ -66,24 +70,24 @@ def verify_instance(self, name):
...
-            ("STANDARD", "1.3", ["m"]),
-            ("HA", "1.3", ["m-0"]),
+            ("SINGLE", ["m"]),
+            ("STANDARD", ["m"]),
+            ("HA", ["m-0"]),
         ],
-        testcase_func_name=DataprocTestCase.generate_verbose_test_name)
-    def test_bigtable(self, configuration, dataproc_version, machine_suffixes):
+        testcase_func_name=DataprocTestCase.generate_verbose_test_name,
+        skip_on_empty=True)
```
medb: Do we still need these option in the tests?

bigtable/test_bigtable.py
```diff
@@ -10,12 +10,16 @@
...
 import os
+import sys
 import unittest
 
+from absl import flags
 from parameterized import parameterized
-
 from integration_tests.dataproc_test_case import DataprocTestCase
 
+FLAGS = flags.FLAGS
```
medb: Can we move flag initialization in the super class `DataprocTestCase` so it will be shared between all the tests?

bigtable/test_bigtable.py
```diff
@@ -66,24 +70,24 @@ def verify_instance(self, name):
...
+            ("SINGLE", ["m"]),
+            ("STANDARD", ["m"]),
+            ("HA", ["m-0"]),
         ],
-        testcase_func_name=DataprocTestCase.generate_verbose_test_name)
-    def test_bigtable(self, configuration, dataproc_version, machine_suffixes):
+        testcase_func_name=DataprocTestCase.generate_verbose_test_name,
+        skip_on_empty=True)
+    def test_bigtable(self, configuration, machine_suffixes):
+        log = logging.getLogger( "BigTableTestCase.test_bigtable" )
```
medb: How this log is used?

cloud_sql_proxy/test_cloud_sql.py
```diff
@@ -1,11 +1,15 @@
 import json
```
medb: It it intentional rename for `cloud-sql-proxy` to `cloud_sql_proxy`? If so, it will be a braking change.

bigtable/test_bigtable.py
estebanlau: Removed the skip_on_empty argument.

bigtable/test_bigtable.py
estebanlau: Now that there is only the --image_version flag which is the same across all the tests this makes a lot of sense. Good catch.

bigtable/test_bigtable.py
estebanlau: Removed.

cloud_sql_proxy/test_cloud_sql.py
estebanlau: Bazel can't reference paths containing hyphens ("-") so the renaming is necessary. What changes must be made to make this change a non-breaking one? Also, with the changes we're making here, the GCB presubmit will break because "python3 -m unittest" will fail because when DataprocTestCase tries importing absl, it will fail.

bigtable/test_bigtable.py
medb: Thanks!

cloud_sql_proxy/test_cloud_sql.py
medb: I see, to make this change backward compatible we will need to change our deployment script to overwrite path when publishing to GCS.

But it seems that there are a [workaround](https://github.com/bazelbuild/bazel/issues/6841), may you try to use it instead of renaming?

medb: Also, we need to fix presubmit to use bazel, so we can test if this PR passes presubmit.

estebanlau: The problem is that whatever rule builds our test will have to mentioned the hyphened path that leads to it. The only workaround for the hyphen issue that occurs to me is:
- leaving only the init action script in its original folder
- moving the tests to another unhyphened folder from which we can run tests referencing our script

For example, having:

- cloud-sql-proxy/cloud-sql-proxy.sh
- cloud-sql-proxy/BUILD  # just needed to export the script

- cloud_sql_proxy/test_cloud_sql.py  # references the init action for testing
- cloud_sql_proxy/pyspark_metastore_test.py
- cloud_sql_proxy/BUILD

Thoughts?

estebanlau: I also tried this but it's not working for me: https://github.com/bazelbuild/bazel/issues/9171#issuecomment-532683449

medb: What about not having BUILD file in `cloud-sql-proxy` directory and configuring test target in the top-level BUILD file for it?

estebanlau: I tried that but it still doesn't work. When I try referencing, for example, "cloud-sql-proxy/test_cloud_sql.py" as a srcs, it complains:

`Label '//:cloud-sql-proxy/test_cloud_sql.py' is invalid because 'cloud-sql-proxy' is a subpackage; perhaps you meant to put the colon here: '//cloud-sql-proxy:test_cloud_sql.py'?`

It expects a rule. So instead, I referenced the srcs as "//cloud-sql-proxy:test_cloud_sql.py" and exported the test Python file by declaring exports_files(["test_cloud_sql.py"]) making it accessible. This output another error:

`/usr/local/google/home/estebanlau/Documents/InitActionsFork/dataproc-initialization-actions/BUILD:15:9: in srcs attribute of py_library rule //:pyspark_metastore_test: //cloud-sql-proxy:pyspark_metastore_test.py: paths to Python packages may not contain '-' `

estebanlau: Got it to work :) Apparently commenting out all the lines of the BUILD file in the cloud-sql-proxy folder is not sufficient. To be able to use cloud-sql-proxy files (e.g. test_cloud_sql) as a srcs, there can be no existing BUILD file so I deleted it.

medb: Cool, thanks for experimenting with this :)

conda/test_conda.py
```diff
@@ -2,22 +2,21 @@
...
 from parameterized import parameterized
-
 from integration_tests.dataproc_test_case import DataprocTestCase
 
 CONDA_BINARY = "/opt/conda/bin/conda"
 PIP_BINARY = "/opt/conda/bin/pip"
 PYTHON_VERSION_KEY = "python_version"
 
+# Test packages
+CONDA_PKGS = ["numpy", "pandas", "jupyter"]
```
medb: Revert this change?

drill/test_drill.py
```diff
@@ -1,10 +1,12 @@
 import os
+import sys
 import unittest
 
 from parameterized import parameterized
-
 from integration_tests.dataproc_test_case import DataprocTestCase
 
+FLAGS = flags.FLAGS
```
medb: Delete?

integration_tests/dataproc_test_case.py
```diff
@@ -72,7 +77,7 @@ def setUpClass(cls):
     def createCluster(self,
                       configuration,
                       init_actions,
-                      dataproc_version,
+                      # dataproc_version,
```
medb: delete?

third_party/enum34.BUILD
```diff
@@ -0,0 +1,13 @@
+# Description:
```
medb: Why do we need this files?

cloud-sql-proxy/test_cloud_sql_proxy.py
```diff
@@ -3,14 +3,13 @@
...
 
 from parameterized import parameterized
-
 from integration_tests.dataproc_test_case import DataprocTestCase
 
 
 class CloudSqlProxyTestCase(DataprocTestCase):
     COMPONENT = 'cloud-sql-proxy'
     INIT_ACTIONS = ['cloud-sql-proxy/cloud-sql-proxy.sh']
-    TEST_SCRIPT_FILE_NAME = 'cloud-sql-proxy/pyspark_metastore_test.py'
```
medb: revert?

estebanlau: Done.

third_party/enum34.BUILD
estebanlau: We needed the six.BUILD file but I'm now referencing the one packaged within io_abseil_py.

integration_tests/dataproc_test_case.py
estebanlau: Done

drill/test_drill.py
estebanlau: FLAGS is used within the test_drill function.

conda/test_conda.py
estebanlau: Done

cloud-sql-proxy/test_cloud_sql_proxy.py
```diff
@@ -45,30 +44,28 @@ def wait_cloud_sql_operation(self, operation_id):
     def verify_cluster(self, name):
         self.__submit_pyspark_job(name)
 
+    def __submit_pyspark_job(self, name):
```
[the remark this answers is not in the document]
estebanlau: Oops, deleted.

PR #650, opened 2019-10-15
Fix Datalab init actions tests

medb:


datalab/test_datalab.py
```diff
@@ -19,14 +19,11 @@ def verify_instance(self, name):
...
         [
-            ("STANDARD", "1.1", ["m"], "python2"),
-            ("STANDARD", "1.1", ["m"], "python3"),
             ("STANDARD", "1.2", ["m"], "python2"),
             ("STANDARD", "1.2", ["m"], "python3"),
             ("STANDARD", "1.3", ["m"], "python2"),
             ("STANDARD", "1.3", ["m"], "python3"),
-            ("STANDARD", "1.4", ["m"], "python2"),
-            ("STANDARD", "1.4", ["m"], "python3"),
+            ("STANDARD", "1.4", ["m"], "python3-default"),
```
functicons: What is python3-default? why do we want to use it?

datalab/test_datalab.py
```diff
@@ -35,7 +32,7 @@ def test_datalab(self, configuration, dataproc_version, machine_suffixes,
         metadata = 'INIT_ACTIONS_REPO={}'.format(self.INIT_ACTIONS_REPO)
         if python == "python3":
             init_actions = self.PYTHON_3_INIT_ACTIONS + init_actions
-            metadata += ',CONDA_PACKAGES="python==3.5"'
+            metadata += ',CONDA_PACKAGES="python=3"'
```
functicons: Is this the right syntax?

medb: Yes, it's correct:
```
$ python --version
Python 3.6.9 :: Anaconda, Inc.
```

datalab/test_datalab.py
medb: It means that we don't need to install Python 3 in 1.4 image because it's install by default (see test code below)

PR #663, opened 2019-10-30
Adding Jupyter Sparkmonitor init action

yhqs540:
Modifying the optional component of Jupyter to include Sparkmonitor which show Spark UI inside the notebook.


jupyter_sparkmonitor/README.MD
```diff
@@ -0,0 +1,59 @@
...
+    action. The following command will create a new cluster named
+    `<CLUSTER_NAME>`.
+
+    ```bash
+    # Simple one-liner; just use all default settings for your cluster.
+    # Jupyter will run on port 8123 of your master node.
+    CLUSTER=<CLUSTER_NAME>
+    gcloud dataproc clusters create $CLUSTER \
+        --optional-components ANACONDA,JUPYTER --enable-component-gateway \
+        --initialization-actions gs://$MY_BUCKET/jupyter-sparkmonitor/sparkmonitor.sh
```
karth295: Nit: your PR uses an underscore (`jupyter_sparkmonitor`) -- we should reconcile the two. Since the other init action directories use dashes, I'd suggest changing the file names to use dashes.

jupyter_sparkmonitor/README.MD
```diff
@@ -0,0 +1,59 @@
...
+
+-   Check to see if Conda and Jupyter Optional Components are installed. Fail if not.
+-   Installing Sparkmonitor using Python Pip
+-   Enable the sparkmonitor as Jupyter Extension and configure IPython to load it
+-   Configure Jupyter PySpark Kernel to use Sparkmonitor
+    -   Remove existing *PySpark* kernel under`jupyter/kernels/pyspark/kernel.json`
+    -   Generate a new *PySpark* kernel and add to the same location
+    -   Refresh Jupyter config and restart Jupyter service
+
+**NOTE**: to be run as an init action.
```
karth295: I think this note is unnecessary because there are no other files in this directory

jupyter_sparkmonitor/sparkmonitor.sh
```diff
@@ -0,0 +1,132 @@
...
+}
+
+function err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')]: $@" >&2
+  return 1
+}
+
+function install_sparkmonitor(){
+  # Install asyncio
+  retry_apt_command "${CONDA_DIRECTORY}/bin/pip install asyncio" ||
```
karth295: Nit: just change this to `retry_command` if it is used for more than `apt` commands

jupyter_sparkmonitor/sparkmonitor.sh
```diff
@@ -0,0 +1,132 @@
...
+    err "Failed to install asyncio"
+
+  # Remove existing PySpark kernel as it uses pyspark shell instead of ipython
+  retry_apt_command "${CONDA_DIRECTORY}/bin/jupyter kernelspec remove ${EXISTING_PYSPARK_KERNEL} -f" ||
```
karth295: Nit: does this command really need retries? (I assumed it was just modifying local files)

jupyter_sparkmonitor/sparkmonitor.sh
```diff
@@ -0,0 +1,132 @@
...
+  retry_apt_command "/${CONDA_DIRECTORY}/bin/pip install sparkmonitor"
+  if [ $? != 0 ]; then
+    err 'Failed to install sparkmonitor'
+  fi
+}
+
+function configure_sparkmonitor(){
+  local sparkmonitor_version;
+  sparkmonitor_version="$(${CONDA_DIRECTORY}/bin/pip list | grep -i sparkmonitor | awk 'END {print $2}')"
+  retry_apt_command "${CONDA_DIRECTORY}/bin/jupyter nbextension install sparkmonitor --py --system --symlink"
```
karth295: Same for all of these commands -- I'm not convinced they need retries since they're juts modifying local files

jupyter_sparkmonitor/sparkmonitor.sh
```diff
@@ -0,0 +1,132 @@
...
+
+  if [[ -f /etc/profile.d/conda.sh ]]; then
+    source /etc/profile.d/conda.sh
+  fi
+
+  if [[ -f /etc/profile.d/effective-python.sh ]]; then
+    source /etc/profile.d/effective-python.sh
+  fi
+
+  if [[ "${ROLE}" == 'Master' ]]; then
```
karth295: This init action only needs to run on the master, right? I think you can just `exit 0` from the top of `main` if it's not the master

jupyter_sparkmonitor/sparkmonitor.sh
```diff
@@ -0,0 +1,132 @@
...
+  sparkmonitor_version="$(${CONDA_DIRECTORY}/bin/pip list | grep -i sparkmonitor | awk 'END {print $2}')"
+  retry_apt_command "${CONDA_DIRECTORY}/bin/jupyter nbextension install sparkmonitor --py --system --symlink"
+  retry_apt_command "${CONDA_DIRECTORY}/bin/jupyter nbextension enable sparkmonitor --py --system"
+  retry_apt_command "${CONDA_DIRECTORY}/bin/jupyter serverextension enable --py --system sparkmonitor"
+  retry_apt_command "${CONDA_DIRECTORY}/bin/ipython profile create"
+  local ipython_profile_location="$(${CONDA_DIRECTORY}/bin/ipython profile locate default)"
+  echo "c.InteractiveShellApp.extensions.append('sparkmonitor.kernelextension')" >>  ${ipython_profile_location}/ipython_kernel_config.py
+}
+
+function install_pyspark_kernel() {
```
karth295: Please add a comment on why this new pyspark kernel works with sparkmonitor while the existing one does not

PR #664, opened 2019-10-30
Adding Jupyter Sparkmonitor init action

yhqs540:


yhqs540:
@googlebot I fixed it.

jupyter_sparkmonitor/README.MD
```diff
@@ -0,0 +1,59 @@
...
+    action. The following command will create a new cluster named
+    `<CLUSTER_NAME>`.
+
+    ```bash
+    # Simple one-liner; just use all default settings for your cluster.
+    # Jupyter will run on port 8123 of your master node.
+    CLUSTER=<CLUSTER_NAME>
+    gcloud dataproc clusters create $CLUSTER \
+        --optional-components ANACONDA,JUPYTER --enable-component-gateway \
+        --initialization-actions gs://$MY_BUCKET/jupyter-sparkmonitor/sparkmonitor.sh
```
medb: Git directory has an underscore in the name (`jupyter_sparkmonitor`), we should fix it to have dash as a separator to be consistent with rest of the init actions (`jupyter-sparkmonitor`).

yhqs540: this is already changed.

jupyter_sparkmonitor/sparkmonitor.sh
```diff
@@ -0,0 +1,136 @@
+#!/bin/bash
+# Copyright 2018 Google, Inc.
```
medb: "2019 Google LLC"

yhqs540:
@medb Tests were added.

jupyter_sparkmonitor/sparkmonitor.sh
```diff
@@ -0,0 +1,143 @@
...
+   }
+  }
+EOF
+
+  /opt/conda/default/bin/jupyter kernelspec install "${pyspark_kernel_tmp_dir}"
+  rm -rf $pyspark_kernel_tmp_dir
+  echo "PySpark kernel setup completed!"
+}
+function main() {
+  if [[ ${DATAPROC_VERSION} < '1.4' ]]; then
```
medb: Does this comparison always work for version strings?

jupyter_sparkmonitor/BUILD
```diff
@@ -0,0 +1,14 @@
+package(default_visibility = ["//visibility:public"])
+
+py_test(
+    name = "test_sparkmonitor",
+    size = "enormous",
+    srcs = ["test_sparkmonitor.py"],
+    data = ["sparkmonitor.sh"],
+    local = True,
+    shard_count = 3,
```
medb: You have only 2 tests, so you need only 2 shards

jupyter_sparkmonitor/test_sparkmonitor.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assert_instance_command(name, verify_cmd)
+
+
+    @parameterized.parameters(
+        ("SINGLE", ["m"]),
+        ("STANDARD", ["m"]),
+    )
+    def test_sparkmonitor(self, configuration, machine_suffixes):
+        # Use 1.4 version of Dataproc to test because it requires Python 3
+        dataproc_image_version = '1.4-debian9'
```
medb: Just return if image_version less than 1.4, don't override it.

Look into other tests, some of them already doing similar thing.

jupyter_sparkmonitor/sparkmonitor.sh
yhqs540: uses is_version_at_least

jupyter_sparkmonitor/test_sparkmonitor.py
yhqs540: Does the integration test pinned at 1.3? https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/blob/master/integration_tests/dataproc_test_case.py#L19

return it would result a successful test although it was not tested.

yhqs540: Changed it to return when image version lower than 1.4

PR #669, opened 2019-11-11
Fix and improve GCB presubmit

medb:
1. Exclude modified files in the root direcotry from presubmit.
1. Improve presubmit tests output.


functicons:
Why do we want "Exclude modified files in the root direcotry from presubmit"?

PR #675, opened 2019-11-25
Add an initialization action for fixing the YARN Timeline Server excessive logging

KoopaKing:
This commit creates a new "hotfix" top level directory for initialization actions that fix known issues in Dataproc that have not yet been fixed by a new release. This also specifically creates a new initialization action "disable-noisy-timeline-server-logs" that fixes a known issue where the YARN TImeline Server does an excessive amount of logging due to the YARN Resource Manager making bad requests to the Timeline Server REST API. Specifically this initialization action configures a Java util logging configuration file that disables the noisy logger then restarts the timeline server to use the new configuration.

### DIFF
--- a/hotfix/README.md
+++ b/hotfix/README.md
@@ -1,0 +1,4 @@
+# Cloud Dataproc Hotfix Initialization Actions
+
+This subdirectory contains initialization actions that can fix known Dataproc
+issues.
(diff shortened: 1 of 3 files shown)


KoopaKing:
Discarding to merge into new hotfixes branch.

PR #689, opened 2019-12-13
Fix Datalab init action on Debian 10

medb:


datalab/datalab.sh
```diff
@@ -111,20 +111,18 @@ function configure_master() {
...
-${ADD_BACKPORTS}
+
 ADD dataproc.list /etc/apt/sources.list.d/
 ADD trusted.gpg /tmp/vm_trusted.gpg
-
 RUN apt-key add /tmp/vm_trusted.gpg
+
+RUN apt-get update
+RUN apt-get install -y software-properties-common
+RUN add-apt-repository 'deb http://archive.ubuntu.com/ubuntu bionic main'
```
functicons: Add comment for why we need this.

datalab/datalab.sh
```diff
@@ -111,20 +111,18 @@ function configure_master() {
...
   cat <<EOF >Dockerfile
 FROM ${DOCKER_IMAGE}
-${ADD_BACKPORTS}
+
 ADD dataproc.list /etc/apt/sources.list.d/
 ADD trusted.gpg /tmp/vm_trusted.gpg
-
 RUN apt-key add /tmp/vm_trusted.gpg
+
+RUN apt-get update
```
functicons: Can we add if check to only apply this change to Debian 10? otherwise there is a risk to break users of other OSs.

medb: We have tests for all the OSes and they are passing, so we shouldn't break existing users.

datalab/datalab.sh
medb: Added.

PR #690, opened 2019-12-13
Fix RStudio compatibility with Debian 10

medb:
1. Update RStudio server to 1.2.5019 version
2. Use Debian 10 compatible package.

### DIFF
--- a/rstudio/rstudio.sh
+++ b/rstudio/rstudio.sh
@@ -17,6 +17,18 @@
 
 set -euxo pipefail
 
+# Only run on the master node
+ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+
+USER_NAME="$(/usr/share/google/get_metadata_value attributes/rstudio-user || echo rstudio)"
+USER_PASSWORD="$(/usr/share/google/get_metadata_value attributes/rstudio-password || true)"
+
+RSTUDIO_SERVER_VERSION=1.2.5019
+RSTUDIO_SERVER_PACKAGE=rstudio-server-${RSTUDIO_SERVER_VERSION}-amd64.deb
+
+OS_ID=$(lsb_release -is | tr '[:upper:]' '[:lower:]')
+OS_CODE=$(lsb_release -cs)
+
 function update_apt_get() {
   for ((i = 0; i < 10; i++)); do
     if apt-get update; then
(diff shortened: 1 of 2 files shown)


rstudio/rstudio.sh
```diff
@@ -80,29 +75,34 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+  run_with_retries apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys ${REPOSITORY_KEY}
+  apt-get install -y software-properties-common
+  add-apt-repository "deb http://cran.r-project.org/bin/linux/${OS_ID} ${OS_CODE}-cran35/"
   update_apt_get
-  apt install -y r-base r-base-dev gdebi-core
+  apt-get install -y r-base r-base-dev gdebi-core
 
+  # Download and install RStudio Server package:
+  # https://rstudio.com/products/rstudio/download-server/debian-ubuntu/
+  if [[ ${OS_CODE} == stretch ]]; then
```
functicons: Why there is no check for Debian 10?

rstudio/rstudio.sh
```diff
@@ -15,7 +15,19 @@
...
+set -euxo pipefail
+
+# Only run on the master node
+ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
+
+USER_NAME="$(/usr/share/google/get_metadata_value attributes/rstudio-user || echo rstudio)"
+USER_PASSWORD="$(/usr/share/google/get_metadata_value attributes/rstudio-password || true)"
+
+RSTUDIO_SERVER_VERSION=1.2.5019
+RSTUDIO_SERVER_PACKAGE=rstudio-server-${RSTUDIO_SERVER_VERSION}-amd64.deb
```
functicons: Seems the package name pattern is different between ubuntu and debian. This is why in the existing code, they use if/else.

medb: Maybe it was different in the past, but not now:
https://rstudio.com/products/rstudio/download-server/debian-ubuntu/

rstudio/rstudio.sh
medb: Because Debian 10 and Ubuntu 18 have the same packages:
https://rstudio.com/products/rstudio/download-server/debian-ubuntu/

PR #692, opened 2019-12-16
Add support for installing GPU drivers on Ubuntu

medb:


gpu/install_gpu_driver.sh
```diff
@@ -20,12 +20,20 @@ set -euxo pipefail
 readonly GPU_AGENT_REPO_URL='https://raw.githubusercontent.com/GoogleCloudPlatform/ml-on-gcp/master/dlvm/gcp-gpu-utilization-metrics'
 
 # Whether to install GPU monitoring agent that sends GPU metrics to StackDriver
-INSTALL_GPU_AGENT="$(/usr/share/google/get_metadata_value attributes/install_gpu_agent || true)"
+INSTALL_GPU_AGENT=$(get_metadata_attribute 'install_gpu_agent' 'false')
```
functicons: I assume this will fail as get_metadata_attribute is not yet defined.

gpu/install_gpu_driver.sh
```diff
@@ -37,48 +45,61 @@ function install_gpu_driver() {
...
+      for distro in ${OS_DIST} ${OS_DIST}-backports; do
+        echo "${type} http://deb.debian.org/debian ${distro} contrib non-free" \
           >>/etc/apt/sources.list.d/non-free.list
       done
     done
-  done
+    packages+=(nvidia-driver nvidia-kernel-common nvidia-smi)
+    modules+=(nvidia-current)
+    local nvblas_cpu_blas_lib=/usr/lib/libblas.so
+  else
```
functicons: It is better to check Ubuntu explicitly, we might have other OS distros in the future.

gpu/install_gpu_driver.sh
medb: Fixed

gpu/install_gpu_driver.sh
medb: Done.

gpu/install_gpu_driver.sh
```diff
@@ -37,48 +44,63 @@ function install_gpu_driver() {
...
           >>/etc/apt/sources.list.d/non-free.list
       done
     done
-  done
+    packages+=(nvidia-driver nvidia-kernel-common nvidia-smi)
+    modules+=(nvidia-current)
+    local nvblas_cpu_blas_lib=/usr/lib/libblas.so
+  fi
+
+  if [[ ${OS_NAME} == ubuntu ]]; then
```
functicons: If the OS is neither Ubuntu nor Debian, nvblas_cpu_blas_lib won't be initialized, it is better to return an error early.

medb: Done.

PR #699, opened 2019-12-27
Datalab: add metadata for server port

functicons:


datalab/datalab.sh
```diff
@@ -149,8 +150,9 @@ EOF
 }
 
 function run_datalab() {
-  if docker run -d --restart always --net=host \
-    -v "${DATALAB_DIR}:/content/datalab" ${VOLUME_FLAGS} datalab-pyspark; then
+  if docker run -d --restart always -p "${DATALAB_PORT}:8080" \
```
medb: Delete `:8080`

functicons: It is not a mistake, it is a port mapping.

PR #702, opened 2020-01-08
Updating init to work with Presto release 323

ChethanUK:
- Software Requirements for Presto 323 is Java JRE 11 (64-bit) - OpenJDK


ChethanUK:
@mattsfuller @findepi

starburst-presto/presto.sh
```diff
@@ -21,77 +21,105 @@ export PATH=/usr/bin:$PATH
...
-PRESTO_QUERY_NODE_MB=0
+PRESTO_JVM_MB=0;
+PRESTO_QUERY_NODE_MB=0;
 # Allocate some headroom for untracked memory usage (in the heap and to help GC).
-PRESTO_HEADROOM_NODE_MB=256
+PRESTO_HEADROOM_NODE_MB=256;
+
+# Add GCS Hive connector Jar
+if [[ -d /usr/local/share/google/dataproc/lib ]]; then
+  readonly CONNECTOR_JAR="$(find /usr/local/share/google/dataproc/lib -name 'gcs-connector-*.jar')"
```
findepi: gcs-connector jar comes builtin in Presto Hive connector

https://github.com/prestosql/presto/blob/7aaa0f5fac1454b23e4bd2ac3a511d749bd00fbe/presto-hive/pom.xml#L173-L174

@medb am i correct we don't need to update it and we can just use `hadoop2-2.0.0` version?

starburst-presto/presto.sh
```diff
@@ -21,77 +21,105 @@ export PATH=/usr/bin:$PATH
...
-readonly PRESTO_VERSION="${PRESTO_MAJOR_VERSION}-e.1"
-readonly HTTP_PORT="$(/usr/share/google/get_metadata_value attributes/presto-port || echo 8080)"
+readonly PRESTO_MAJOR_VERSION="$(/usr/share/google/get_metadata_value attributes/presto_major_version || echo '323')"
+readonly STARBURST_PRESTO_VERSION="$(/usr/share/google/get_metadata_value attributes/sbpresto_version || echo '323-e.3')" 
+readonly HTTP_PORT="$(/usr/share/google/get_metadata_value attributes/presto-port || echo 8060)"
 readonly INIT_SCRIPT="/usr/lib/systemd/system/presto.service"
-PRESTO_JVM_MB=0
-PRESTO_QUERY_NODE_MB=0
+PRESTO_JVM_MB=0;
+PRESTO_QUERY_NODE_MB=0;
```
findepi: please take a look at the diff on gh and revert redundant changes

starburst-presto/presto.sh
```diff
@@ -21,77 +21,105 @@ export PATH=/usr/bin:$PATH
 readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
 readonly PRESTO_MASTER_FQDN="$(/usr/share/google/get_metadata_value attributes/dataproc-master)"
 readonly WORKER_COUNT=$(/usr/share/google/get_metadata_value attributes/dataproc-worker-count)
-readonly PRESTO_BASE_URL=https://storage.googleapis.com/starburstdata/presto
-readonly PRESTO_MAJOR_VERSION="312"
-readonly PRESTO_VERSION="${PRESTO_MAJOR_VERSION}-e.1"
-readonly HTTP_PORT="$(/usr/share/google/get_metadata_value attributes/presto-port || echo 8080)"
+readonly PRESTO_MAJOR_VERSION="$(/usr/share/google/get_metadata_value attributes/presto_major_version || echo '323')"
+readonly STARBURST_PRESTO_VERSION="$(/usr/share/google/get_metadata_value attributes/sbpresto_version || echo '323-e.3')" 
+readonly HTTP_PORT="$(/usr/share/google/get_metadata_value attributes/presto-port || echo 8060)"
```
findepi: why did you change the port number from 8080 to 8060

findepi:
@ChethanUK 
many thanks for working on this!

Some general comment --
This does several things

- upgrades from 312-e.1 to 323-e.3
- makes the version configurable
- changes default port
- introduces formatting changes (like semicolons)
- introduces retrying

From these, I'd like to first see a pull request just doing upgrade to 323-e.3, separately from all other changes.

Configurable version -- i don't think it makes sense to do this.
if we want this we should limit variability to minor release patch level. (Otherwise we're lying that script works with any version even though it has version-dependent logic in it)

Changing default port -- we already discussed this and this is probably no-go.

Formatting changes -- I think Dataproc team is opinionated about this, so I'd defer to them on this. (a separate PR anyway)

Retrying -- let's make it a separate PR.

starburst-presto/presto.sh
medb: Yes, there no need to update it manually.

ChethanUK:
Okay, I will start with 
> * upgrades from 312-e.1 to 323-e.3

https://github.com/GoogleCloudPlatform/dataproc-initialization-actions/pull/703

PR #707, opened 2020-01-12
Create an individual Alluxio service per daemon

madanadit:
The Alluxio master and worker roles both start 2 daemons. Each daemon is split into an individual service so that the process lifetime is managed by systemctl independently.


madanadit:
@medb PTAL, thx again

PR #737, opened 2020-03-16
Add support for running integration tests with Docker

functicons:


integration_tests/docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
+# This Dockerfile creates a Docker image with the dependencies for running init
+# action integrations.
+
+FROM gcr.io/cloud-builders/gcloud
+
+# Install Bazel https://docs.bazel.build/versions/master/install-ubuntu.html
+RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
+RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
+RUN apt-get update && apt-get install -y openjdk-8-jdk python3-setuptools bazel
```
medb: I don't think that supporting separate Docker image for running tests worth it, taking into account that setup is just installing for 2 deb packages.

What is the problem with installing `python3-setuptools` and `bazel` locally?

functicons: Benefits of Docker: 1) isolation: there can be conflicting dependencies on the same machine; 2) convenience: users don't need to care about the dependencies added by the test framework, they just focus on the init actions.

I personally feel it is a lot easier to use Docker. I use Miniconda locally, I ran into missing Python packages (`absl`) issue when I was trying to run it without Docker.

medb: Docket is useful for complex environments, but I'm not sure that this is the case at hand. For example, we do not use Docker to run Maven builds in our connectors.

`absl` is installed by Bazel as a dependency. How did you run your test and what exactly error did you face?

functicons: I was reviewing Sqoop PR, which still uses the `python -m ...` to run tests. Anyway, I feel it is better to support Docker, but up to you.

medb: Yeah, this is not how it should be run.

I think it will be better to configure Bazel to pull-in `setuptools` as a dependency too, this way you will need to install only Bazel to run the tests which is simpler than configuring it in Docker environment, also maintenance cost is lower for us too:

```bash
bazel test :DataprocInitActionsTestSuite --jobs=25 --test_arg=--generate_ssh_key
```
vs

```bash
SERVICE_ACCOUNT_KEY_FILE=<path>
SERVICE_ACCOUNT=<email>
PROJECT=<project>
REGION=<region>
ZONE=<zone>
LOG_DIR=<dir>

# Running tests
docker run \
    -v ${PWD}:/src \
    -v ${LOG_DIR}:/root/.cache \
    -v ${SERVICE_ACCOUNT_KEY_FILE}:/keys/service-account.json \
    -e SERVICE_ACCOUNT_KEY_FILE=/keys/service-account.json \
    -e SERVICE_ACCOUNT=${SERVICE_ACCOUNT} \
    -e PROJECT=${PROJECT} \
    -e REGION=${REGION} \
    -e ZONE=${ZONE} \
    init-action-test-runner \
    bazel test :DataprocInitActionsTestSuite --jobs=25 --test_arg=--generate_ssh_key
```

PR #746, opened 2020-04-15
Updating presto to v332

ChethanUK:
NOTE: If we don't use Java 11
Then we get the following error:

```bash
INFO	main	stderr	ERROR: Future versions of Presto will require Java 11 after March 2020.

You may temporarily continue running on Java 8 by adding the following
JVM config option:

    -Dpresto-temporarily-allow-java8=true
```

### DIFF
--- a/starburst-presto/presto.sh
+++ b/starburst-presto/presto.sh
@@ -36,26 +36,6 @@ function err() {
   return 1
 }
 
-function retry_apt_command() {
-  cmd="$1"
-  for ((i = 0; i < 10; i++)); do
-    if eval "$cmd"; then
-      return 0
-    fi
-    sleep 5
-  done
-  return 1
-}
-
-function update_apt_get() {
-  retry_apt_command "apt-get update"
-}
-
-function install_apt_get() {
-  local pkgs="$*"
-  retry_apt_command "apt-get install -y $pkgs"
-}
-
 function wait_for_presto_cluster_ready() {
   # wait up to 120s for presto being able to run query
   for ((i = 0; i < 12; i++)); do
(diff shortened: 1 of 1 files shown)


ChethanUK:
@mattsfuller @findepi

starburst-presto/presto.sh
```diff
@@ -229,6 +249,9 @@ EOF
 function configure_and_start_presto() {
   mkdir -p /opt/presto-server/etc/catalog
 
+  update_apt_get
+  install_apt_get openjdk-11-jre
```
medb, line 253: May you instead set property to preserve compatibility w/ Java 8? Generally, on Dataproc we try to use the same JVM for all components to simplify deployment and maintenance.

ChethanUK: > property to preserve compatibility w/ Java 8?

Okay!.

findepi: @ChethanUK please add `-Dpresto-temporarily-allow-java8=true` to `jvm.config`

starburst-presto/presto.sh
```diff
@@ -36,6 +36,26 @@ function err() {
...
+}
+
+function update_apt_get() {
+  retry_apt_command "apt-get update"
+}
+
+function install_apt_get() {
+  local pkgs="$*"
+  retry_apt_command "apt-get install -y $pkgs"
+}
```
findepi, line 57: once you remove java 11 installation, you can also remove this

findepi:
@ChethanUK thanks for taking care of this.

The release artifacts are now live in google cloud storage too.
https://storage.googleapis.com/starburstdata/presto/332e/332-e.0/presto-server-332-e.0.tar.gz

starburst-presto/presto.sh
ChethanUK: Yes will give the option to choose Java8 as well.

starburst-presto/presto.sh
medb: Done.

starburst-presto/presto.sh
medb: JDK 11 removed

findepi:
Thanks @ChethanUK!

PR #751, opened 2020-05-05
Datalab: enable HTTPS for apt repos.

functicons:


datalab/datalab.sh
```diff
@@ -115,14 +115,16 @@ function configure_master() {
   cat <<EOF >Dockerfile
 FROM ${DOCKER_IMAGE}
 
+# Enabling APT to download from HTTPS repository.
+RUN apt-get update
+RUN apt-get install -y dirmngr apt-transport-https software-properties-common
```
medb, line 120: Why do we need `dirmngr`?

functicons: Removed

PR #752, opened 2020-05-05
Create workflow that runs all init actions tests and uploads its TestGrid-friendly formatted results into GCS 

estebanlau:
The purpose of this workflow is to periodically run all init actions tests, massage the Bazel output into a format consumable by TestGrid, and upload the results to GCS. From GCS, our TestGrid configuration can point at our buckets to show test results in TestGrid (at the target/init action level).


cloudbuild/Dockerfile
```diff
@@ -11,6 +11,9 @@ COPY --chown=ia-tests:ia-tests . /init-actions
 # https://docs.bazel.build/versions/master/install-ubuntu.html
 RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
 RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
-RUN apt-get update && apt-get install -y openjdk-8-jdk python3-setuptools bazel
+RUN apt-get update && apt-get install -y openjdk-8-jdk python3-setuptools bazel libxml2-utils
+
+# Install jq
+RUN apt-get -y install jq
```
medb, line 17: Split `apt-get update` and `apt-get install`:
```
RUN apt-get update
RUN apt-get install -y openjdk-8-jdk python3-setuptools bazel libxml2-utils jq
```

cloudbuild/Dockerfile
```diff
@@ -11,6 +11,9 @@ COPY --chown=ia-tests:ia-tests . /init-actions
 # https://docs.bazel.build/versions/master/install-ubuntu.html
 RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list
 RUN curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
-RUN apt-get update && apt-get install -y openjdk-8-jdk python3-setuptools bazel
+RUN apt-get update && apt-get install -y openjdk-8-jdk python3-setuptools bazel libxml2-utils
```
medb, line 14: Why do we need `libxml2-utils`?

cloudbuild/periodic/run-all-tests.sh
```diff
@@ -0,0 +1,108 @@
...
+  else
+    status="SUCCESS"
+  fi
+	echo $status
+}
+
+create_finished_json() {
+  component=$1
+  status=$(get_test_status $component)
+  jq -n \
```
medb, line 72: Why do you need `jq` for this? seems like you can create JSON string without it pretty easily.

cloudbuild/periodic/run-all-tests.sh
```diff
@@ -0,0 +1,108 @@
...
+
+set -uxo pipefail
+
+COMMIT=$(git rev-parse HEAD)
+TESTS_TO_RUN=(
+  ":DataprocInitActionsTestSuite"
+) 
+
+run_tests() {
+  bazel test \
```
medb, line 11: I think that we should reuse the same script and Cloud Build config for running presubmit and nightly tests. We just need to add a parameter that will force it to run all tests and upload results to GCS. After that we can have a separate workflow that will read test results from GCS and transform them in testgrid format.

cloudbuild/Dockerfile
estebanlau: Done.




> May you attach link with an example how this looks like?

A workflow run looks like this: https://pantheon.corp.google.com/cloud-build/builds/3212df70-c22f-4a0d-98f0-0d7d2dbc1287;step=5?project=cloud-dataproc-ci

TestGrid would look like this: https://screenshot.googleplex.com/jLhd89fyiVj.png

cloudbuild/Dockerfile
estebanlau: We need libxml2-utils to use xmlutils, which is used to parse XML test results created by running Bazel tests.

cloudbuild/periodic/run-all-tests.sh
estebanlau: I like having the GCB workflows separate because this way we can easily distinguish between presubmit and periodic runs when going through our GCB workflow history. This makes it easier for tracking and debugging. We also probably only want the periodic test results to show in TestGrid - we're not interested in having the presubmit results in TestGrid, correct? In this case, the transformation should only happen for the periodic tests' results.

What do you think?

cloudbuild/periodic/run-all-tests.sh
estebanlau: I like that jq ensures that the created file has a correct JSON format. Doing it by creating the string also works and is simple but with jq it's more reliable because whatever is uploaded to GCS will always be a valid JSON file that is readable. For example, doing it by string creation, a typo will not fail the file creation process causing a wrong JSON file to be uploaded. I'd rather have it fail when it's trying to create it.

cloudbuild/periodic/run-all-tests.sh
estebanlau: After discussion, I consolidated both workflows/triggers to be controlled by a single yaml config. The _RUN_ALL_TESTS controls if it's run as a presubmit or a periodic test.

Also, the build number is now extracted by reading GCS and finding the folder name (i.e. build number) with the highest number.

cloudbuild/cloudbuild.yaml
```diff
@@ -1,4 +1,13 @@
 steps:
+  - name: 'gcr.io/cloud-builders/gsutil'
+    id: 'get-build-num'
+    entrypoint: /bin/bash
+    args:
+      - -c
+      - |
+        chmod +x cloudbuild/periodic/get-and-update-build-num.sh &&
+        cloudbuild/periodic/get-and-update-build-num.sh
```
medb, line 9: Can we embed this script's logic in the `cloudbuild/Dockerfile` itself instead of heaving a separate step for it?

cloudbuild/run-tests.sh
```diff
@@ -93,6 +99,7 @@ run_tests() {
     --test_output=errors \
     --noshow_progress \
     --noshow_loading_progress \
+    --keep_going \
```
medb, line 102: Will this impact presubmit?

cloudbuild/run-tests.sh
```diff
@@ -101,9 +108,16 @@ main() {
   cd /init-actions
   configure_gcloud
   configure_gcloud_ssh_key
-  initialize_git_repo
-  determine_tests_to_run
-  run_tests
+  if [[ $RUN_ALL_TESTS=="true" ]]; then
+    # Run periodic
+    initialize_git_repo "false"
+    bash cloudbuild/periodic/run-all-tests.sh
```
medb, line 114: Seems like we can reuse `run_tests` function by modifying it to disable logic not needed for periodic tests and calling it with `true` to make it always succeed:
```bash
run_tests || true
```

cloudbuild/run-tests.sh
```diff
@@ -101,9 +108,16 @@ main() {
   cd /init-actions
   configure_gcloud
   configure_gcloud_ssh_key
-  initialize_git_repo
-  determine_tests_to_run
-  run_tests
+  if [[ $RUN_ALL_TESTS=="true" ]]; then
+    # Run periodic
+    initialize_git_repo "false"
```
medb, line 113: Do we even need to call this function here?

cloudbuild/periodic/get-and-update-build-num.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+shopt -s extglob;
+DIRS=$(gsutil ls gs://init-actions-github-tests/logs/init_actions_tests/hue-1.4-debian9)
```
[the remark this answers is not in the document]
estebanlau: I'm just going into any arbitrary init action and extracting the latest build number. The other option was using a file to keep track of the build number, like I was doing before.

cloudbuild/cloudbuild.yaml
```diff
@@ -19,6 +28,7 @@ steps:
       - 'COMMIT_SHA=$COMMIT_SHA'
       - 'CLOUDSDK_COMPUTE_ZONE=us-central1-f'
       - 'CLOUDSDK_CONTAINER_CLUSTER=init-actions-presubmit'
+      - 'RUN_ALL_TESTS=$_RUN_ALL_TESTS'
```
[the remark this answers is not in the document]
estebanlau: Yes, it's enforced.

cloudbuild/run-tests.sh
estebanlau: No, I think I could even remove this and it wouldn't change anything.

cloudbuild/run-tests.sh
estebanlau: Done.

cloudbuild/run-tests.sh
estebanlau: I'm calling it so that we can call "git rev-parse head" to get the commit number. Maybe, there is a better way?

cloudbuild/cloudbuild.yaml
estebanlau: If I put it in the Dockerfile, I have to authenticate or activate a service account. Doing it from the top-level GCB step allows me to use gsutil without problem.

cloudbuild/run-tests.sh
medb: Do you need `git rev-parse head` for TestGrid?

cloudbuild/cloudbuild.yaml
medb: I think that because Docker file executed it has access for the same credentials as this script. May you try to do this in Docker file?

cloudbuild/periodic/get-and-update-build-num.sh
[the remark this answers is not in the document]
medb: I see, I thought that image build number is a top-level folder. It seems that in this case multiple periodic concurrent runs can overlap with each other and produce invalid/inconsistent TestGrid config?

cloudbuild/run-tests.sh
estebanlau: Not really. I was getting it hoping that I could have it surface in TestGrid so that we can know which test run column corresponds to which code version. However, I'm already providing the GCS build number which surfaces the commit num. Example: https://screenshot.googleplex.com/ZRpNKg4CZv8.png

PR #753, opened 2020-05-08
Update install_gpu_driver.sh retry logic to double number of retries

cyxxy:
Increase retry number from 10 to 20 in "execute_with_retries()" method and increase "wget" retry number from 5 to 10.


cyxxy:
Ha, just realized 1.2 images have older version of wget so this won't work.

PR #758, opened 2020-05-15
Adding Machine Learning VM init action

bradmiro:
This init action pre-configures a Dataproc cluster with a set of machine learning packages and other tools frequently-used tools in the data science community, including:
- NVIDIA GPU drivers + CUDA
- RAPIDS on Spark
- Spark-BigQuery connector
- Anaconda and Jupyter via Optional Components
- Component Gateway integration


ml-vm/ml-vm.sh
```diff
@@ -0,0 +1,127 @@
...
+  local -r name="spark-nlp"
+  local -r repo_url="http://dl.bintray.com/spark-packages/maven/JohnSnowLabs/"
+  local -r version="2.4.3"
+  
+  download_spark_jar "${repo_url}/${name}/${version}/${name}-${version}.jar"
+}
+
+function install_r_packages() {
+  # Install R system dependencies
+  apt-get install -y libcurl4-openssl-dev libssl-dev libxml2-dev
```
roderickyao, line 96: add retry logic?

bradmiro:
Thanks for feedback and I'm working on an integration test. It appears a bit involved so I'd love to get feedback on my plan for this:

- Create a cluster with the IA installed.
- For every init. action utilized here, copy their unit test into this integration test.
- For the Python / R libraries, create a Python script and R script that imports and runs small snippets for each library.
- For anything else, create unit tests where appropriate.

medb:
@bradmiro plan LGTM, but I think that we need to focus on integration tests more than unit tests. Ideally we will have an integration test per each high-level framework that this init actions installs, to make sure that it actually works and compatible with new Datparoc images

bradmiro:
I added a first pass of the tests (still need to do Horovod). There's a lot of tests here and I'm actually unable to run this locally due to GPU quotas on my project (working on getting this adjusted). Can the test project handle this?

mlvm/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+You can use this initialization action to create a new Dataproc cluster with
+a set of preconfigured machine learning packages.:
+
+1.  Use the `gcloud` command to create a new cluster with this initialization action. The command shown below includes a curated list of packages that will be installed on the cluster:
+
+    ```bash
+    REGION=<region>
+    CLUSTER_NAME=<cluster_name>
+        gcloud beta dataproc clusters create ${CLUSTER_NAME} \
```
medb, line 28: Fix indentation.

mlvm/README.md
```diff
@@ -0,0 +1,48 @@
...
+        --region ${REGION} \
+        --master-machine-type n1-standard-8 \
+        --worker-machine-type n1-highmem-32 \
+        --worker-accelerator type=nvidia-tesla-t4,count=2 \
+        --image-version 1.5-ubuntu \
+        --metadata gcs-connector-version=2.1.1 \
+        --metadata bigquery-connector-version=1.1.1 \
```
medb, line 35: Delete these 2 metadata value? They should be not relevant for ML.

mlvm/README.md
bradmiro: ack

mlvm/README.md
bradmiro: ack, agreed

mlvm/test_mlvm.py
```diff
@@ -0,0 +1,333 @@
...
+                metadata = "{},{}".format(metadata, connector_metadata)
+        else:
+            metadata = connector_metadata
+        
+        config["metadata"] = metadata
+
+        super().createCluster(configuration, self.INIT_ACTIONS,
+                              **config)
+
+class ConnectorsTestCase(MLVMTestCase):
```
medb, line 33: Why do we need a separate connectors test case?

bradmiro: Since I'm using the connectors init action I just copied all the tests over from there. I suppose we don't need to keep all of them.

mlvm/README.md
```diff
@@ -0,0 +1,153 @@
...
+    ```bash
+    REGION=<region>
+    CLUSTER_NAME=<cluster_name>
+    gcloud dataproc clusters create ${CLUSTER_NAME} \
+        --region ${REGION} \
+        --master-machine-type n1-standard-16 \
+        --worker-machine-type n1-highmem-32 \
+        --worker-accelerator type=nvidia-tesla-t4,count=2 \
+        --image-version preview-ubuntu \
+        --metadata spark-bigquery-connector-version=0.17.0 \
```
medb, line 54: Update to 0.17.1

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+
+readonly JARS_DIR=/usr/lib/spark/jars
+readonly CONNECTORS_DIR=/usr/local/share/google/dataproc/lib
+
+# readonly DEFAULT_INIT_ACTIONS_BUCKET=gs://dataproc-initialization-actions
+# readonly INIT_ACTIONS_BUCKET="$(/usr/share/google/get_metadata_value attributes/INIT_ACTIONS_BUCKET ||
+#   echo ${DEFAULT_INIT_ACTIONS_BUCKET})"
+readonly DEFAULT_INIT_ACTIONS_REPO="https://github.com/GoogleCloudDataproc/initialization-actions.git"
+readonly INIT_ACTIONS_REPO="$(/usr/share/google/get_metadata_value attributes/INIT_ACTIONS_REP ||
+  echo ${DEFAULT_INIT_ACTIONS_REPO})"
```
medb, line 31: Let's delete support for pulling init actions from GitHub

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+readonly INIT_ACTIONS_BRANCH="$(/usr/share/google/get_metadata_value attributes/INIT_ACTIONS_BRANCH ||
+  echo 'master')"
+readonly INIT_ACTIONS_DIR=/opt/init-actions
+readonly include_gpus="$(/usr/share/google/get_metadata_value attributes/include-gpus || true)"
```
medb, line 35: Capitalize variable name

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+
+# Conda
+BASE_R_PACKAGES=(
```
medb, line 38: readonly

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+  git clone -b "${INIT_ACTIONS_BRANCH}" --single-branch "${INIT_ACTIONS_REPO}" "${INIT_ACTIONS_DIR}"
+  find "${INIT_ACTIONS_DIR}" -name '*.sh' -exec chmod +x {} \;
+}
+
+function install_connectors() {
+  "${INIT_ACTIONS_DIR}/connectors/connectors.sh"
+}
+
+function install_python_packages() {
+  readonly EXTRA_PYTHON_PACKAGES="$(/usr/share/google/get_metadata_value attributes/PYTHON_PACKAGES || true)"
```
medb, line 115: change name to lower case

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+
+  # Installing all at once is flaky. Added this here. 
+  for package in "${BASE_PYTHON_PACKAGES[@]}"; do
+    execute_with_retries "pip install $package"
```
medb, line 119: Pass all packages in a single command: `pip install "${BASE_PYTHON_PACKAGES[@]}"`

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+  done
+
+  if [[ -n "${EXTRA_PYTHON_PACKAGES}" ]]; then
+    for package in "${EXTRA_PYTHON_PACKAGES[@]}"; do
+      execute_with_retries "pip install $package"
```
medb, line 124: Pass all packages in a single command: pip install "${EXTRA_PYTHON_PACKAGES[@]}"

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+  local -r name="spark-nlp"
+  local -r repo_url="http://dl.bintray.com/spark-packages/maven/JohnSnowLabs/"
+  local -r version="2.5.4"
+  
+  pip install "spark-nlp==$version" 
+  download_spark_jar "${repo_url}/${name}/${version}/${name}-${version}.jar"
+}
+
+function install_r_packages() {  
+  readonly EXTRA_R_PACKAGES="$(/usr/share/google/get_metadata_value attributes/R_PACKAGES || true)"
```
medb, line 158: convert to lower case

mlvm/test_mlvm.py
```diff
@@ -0,0 +1,259 @@
...
```
medb, line 33: May we just that some BQ job on a public dataset works, instead of re-testing connectors init action?

mlvm/test_mlvm.py
```diff
@@ -0,0 +1,259 @@
+import os
+
+import pkg_resources
+from absl.testing import absltest
+from absl.testing import parameterized
+
+from integration_tests.dataproc_test_case import DataprocTestCase
+
+class MLVMTestCase(DataprocTestCase):
```
medb, line 9: There seems to be too many test cases here - 26

May we test all the functionality on the single cluster that has all features/packages instead? It will reduce test time and amount of test code.

mlvm/mlvm.sh
bradmiro: I originally had this but it would flake by occasionally leading to some closed connections from PyPi. I'll try again with the wrapper retry function.

mlvm/README.md
bradmiro: I think I made an error below, I don't believe 0.17.1 is released yet. I'm also going to move this functionality into the init. action as this is a repetitive step.

mlvm/test_mlvm.py
bradmiro: I was able to reasonably reduce this to 4. Verifying locally.

mlvm/mlvm.sh
bradmiro: Can we force an update to the public GCS bucket? Some of the tests here rely on the updates made to the RAPIDS init action last week.

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+  BASE_PYTHON_PACKAGES+=("tensorflow==2.2.0")
+fi
+
+if [ "$(echo "$DATAPROC_VERSION >= 2.0" | bc)" -eq 1 ]; then 
+  BASE_PYTHON_PACKAGES+=("spark-tensorflow-distributor==0.1.0")
+fi
+
+# Conda
+readonly BASE_R_PACKAGES=(
+  "r-essentials=3.6.0"
```
medb, line 72: Can we detect R version dynamically? We plan to upgrade Dataproc 2.0 to R 4.0, if we will do this then it we will brake init action.

mlvm/mlvm.sh
```diff
@@ -0,0 +1,197 @@
...
+function download_spark_jar() {
+  local -r url=$1
+
+  wget -nv --timeout=30 --tries=5 --retry-connrefused \
+  -P "${JARS_DIR}" "${url}" 
+}
+
+function download_init_actions() {
+  # Download initialization actions locally.
+  gsutil -m rsync -r "${INIT_ACTIONS_REPO}" "${INIT_ACTIONS_DIR}"
```
medb, line 101: Can we sync only required init actions?

mlvm/mlvm.sh
```diff
@@ -0,0 +1,217 @@
...
+  extra_r_packages="$(/usr/share/google/get_metadata_value attributes/R_PACKAGES || true)"
+  
+  # Conda R channel isn't currently as up-to-date with R4 as conda-forge
+  if [[ ${R_VERSION} == "4"* ]]; then
+    channel="conda-forge"
+  else
+    channel="r"
+  fi
+
+  conda install -y -c ${channel} "${BASE_R_PACKAGES[@]}"
```
medb, line 145: I think it's better to use default channel - if R 4.0 will be installed, then channel should be already configured that has it.

mlvm/mlvm.sh
```diff
@@ -0,0 +1,217 @@
...
+  
+  if [[ -n "${extra_r_packages}" ]]; then
+    conda install -y -c ${channel} "${extra_r_packages[@]}"
```
medb, line 148: The same applied to other packages?

mlvm/mlvm.sh
```diff
@@ -0,0 +1,217 @@
...
+  "sparksql-magic==0.0.3" 
+  "tensorflow-datasets==3.2.1"
+  "tensorflow-estimator==2.2.0"
+  "tensorflow-hub==0.8.0"
+  "tensorflow-io==0.14.0"
+  "tensorflow-probability==0.10.1" 
+  "torch==1.5.1" 
+  "torchvision==0.6.1" 
+  "xgboost==1.1.0"
+)
```
medb, line 67: What will happen if we will not pin versions? Or maybe pin just couple top-level packages? It could be toilsome if we need to update/maintain them manually?

mlvm/mlvm.sh
bradmiro: The R channel is a bit slower to update their packages to 4.0 than the community provides in conda-forge, in particular r-essentials does not have a 4.0 version available yet and I ran into versioning conflicts with the conda-forge 3.6 version as it's built on a newer R version than we have on Dataproc.

mlvm/mlvm.sh
bradmiro: Just a potential issue with unexpected version conflicts / breaking changes. I think it's good to provide the versioning transparently to the user as well. Going forward maybe we can use https://github.com/renovatebot/renovate.

mlvm/mlvm.sh
medb: I meant that we should install R 4.0 only if it already installed on the image, and if it installed on the image then it means that R 4.0 was either added to default r channel or image uses forge channel by default. In both cases we don't need to chose channel explicitly.

mlvm/mlvm.sh
medb: We don't do this even for base Dataproc image and rely on Ana/Miniconda version to resolve dependencies. The problem that it should work for future Dataproc versions but we do not version init actions, so we need to support this map per-image track or just relay on conda version to resolve dependencies.

mlvm/mlvm.sh
bradmiro: I see, that makes sense. Ack.

mlvm/mlvm.sh
bradmiro: Ack.

mlvm/mlvm.sh
bradmiro: Ok, in this case it probably makes sense to install these via conda where appropriate / possible. I'll look into.

PR #768, opened 2020-06-21
give yarn service account permission to gpu device

mengdong:
seems like 
chown :yarn -R /sys/fs/cgroup/cpu,cpuacct
chmod g+rwx -R /sys/fs/cgroup/cpu,cpuacct
chown :yarn -R /sys/fs/cgroup/devices
chmod g+rwx -R /sys/fs/cgroup/devices
is required to make gpu resource available to yarn.


mengdong:
I will organize and resubmit

PR #781, opened 2020-07-20
[rapids]: improve Spark 2.x compatibility and mix cleanups

bradmiro:
Recent change to init action produced backwards-compatible breaking change requiring Spark 2.x to be explicitly requested. Changes:

- init action can detect Spark version in the environment, no longer requiring passing it in
- default versions for CUDA, CUDF and RAPIDS configured for Spark 2.x and Spark 3.0 (previously passed in at runtime)
- moved spark 2.x property `spark:spark.submit.pyFiles` into init action
- Dask test skipped due to long test time and note of suggestion to use Dataproc 2.0 for this functionality

### DIFF
--- a/rapids/README.md
+++ b/rapids/README.md
@@ -131,9 +131,10 @@ To use XGBoost4j with Spark 2
     *   NVIDIA driver 410.48+
     *   CUDA v10.2/v10.1/v10.0/v9.2
     *   NCCL 2.4.7 and later
-*   `EXCLUSIVE_PROCESS` must be set for all GPUs in each NodeManager (this
-    initialization action sets this mode by default)
-*   `spark.dynamicAllocation.enabled` property must be set to `false` for Spark
+*   `EXCLUSIVE_PROCESS` must be set for all GPUs in each NodeManager (set by default in this
+    initialization action)
+*   `spark.dynamicAllocation.enabled` property must be set to `false` for Spark (set by default 
+    in this initialization action)
 
 ### Step 1. Download dataset for Spark RAPIDS XGBoost application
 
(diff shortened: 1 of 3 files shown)


bradmiro:
@mengdong FYI.

Additionally, I noticed in the initialization action that Spark jars for Scala 2.11 are installed for Dataproc 1.x. However, Dataproc 1.5 actually uses Scala 2.0 with Spark 2.4.5. The current tests for 1.5 pass, however I don't know if this is sufficient to confirm that the RAPIDS jars will work with Scala 2.11. Can please you confirm that this is either ok or I need to make another fix here?

mengdong:
Hi Brad, is there a typo `Dataproc 1.5 actually uses Scala 2.0 with Spark 2.4.5.`, do you mean scala 2.12?

the RAPIDS SQL plugin jar won't work with spark 2, we only download cuDF and XGBoost with spark 2, per https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/rapids/rapids.sh#L73-L82. I have tried 1.5 briefly it seems in this case mixed scala 2.12 and 2.11 is fine, but we probably shouldn't recommend it.

bradmiro:
Hey Dong,

Apologies, yes that's a typo and I just fixed it in the OP. Dataproc 1.5 uses Scala 2.12.10.

Ok, that sounds good. Are there instructions / supported functionality for building these with Spark 2.x and Scala 2.12?

mengdong:
User has to build from https://github.com/NVIDIA/spark-xgboost/tree/spark-2, I don't have the instruction yet, could we use existing 2.11 jar for now, let me ping the internal team to ask if they can provide another build.

bradmiro:
A Spark 2.4/Scala 2.12 release would be superb, for now we can leave as is. Thanks Dong!

bradmiro:
@medb I think this is good to merge unless you have any suggestions

PR #786, opened 2020-07-27
Added instructions on how to create metrics

### LINKED ISSUE #787: Add instructions on how to create Cloud Monitoring metrics for GPU init action
The GPU init action installs the Cloud monitoring agent to send metrics for GPU usage to Cloud Monitoring but it does not actually create the metrics for you. Therefore you need to run the following first:

```bash
git clone https://github.com/GoogleCloudPlatform/ml-on-gcp.git
cd ml-on-gcp/dlvm/gcp-gpu-utilization-metrics 
pip install -r ./requirements.txt
python create_gpu_metrics.py 
```

Taken from https://github.com/GoogleCloudPlatform/ml-on-gcp/tree/master/dlvm/gcp-gpu-utilization-metrics#report-metrics

I have proposed an update to the instructions in this pull request https://github.com/GoogleCloudDataproc/initialization-actions/pull/786 to show how to do this.

tfayyaz:
Fixes #787

### DIFF
--- a/gpu/README.md
+++ b/gpu/README.md
@@ -54,7 +54,7 @@ gpu-driver-provider=<OS|NVIDIA>` metadata value.
     and CUDA installed by initialization action. Additionally, it installs GPU
     monitoring service.
 
-    Prerequisite: Create metrics in Cloud Monitoring using Cloud Shell.
+    *Prerequisite:* Create metrics in [Cloud Monitoring](https://cloud.google.com/monitoring/docs/) using Cloud Shell.
 
     If you run this locally you will need to set up a service account.
 
@@ -78,7 +78,7 @@ gpu-driver-provider=<OS|NVIDIA>` metadata value.
         --worker-accelerator type=nvidia-tesla-v100,count=4 \
         --initialization-actions gs://goog-dataproc-initialization-actions-${REGION}/gpu/install_gpu_driver.sh \
         --metadata install-gpu-agent=true \
-        --scopes https://www.googleapis.com/auth/monitoring.write
+        --scopes https://www.googleapis.com/auth/monitoring.write,https://www.googleapis.com/auth/cloud-platform
     ```
 
 #### Supported metadata parameters:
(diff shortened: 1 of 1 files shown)


medb:
I think this already mentioned here: https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/gpu/README.md#generate-metrics

Should we make it more prominent instead or mention in cluster creation instructions?

tfayyaz:
I think it needs to be more prominent as currently option 3 states that it installs the GPU monitoring service but it is not clear that you need to actually generate the metrics in Cloud Monitoring first.

Also the instructions and code mentioned in https://github.com/GoogleCloudDataproc/initialization-actions/tree/master/gpu#generate-metrics do not actually work or related to anything in the init action code.

I am not sure what `create_metric_descriptor` relates to? Also it's not clear where you run the cmd mentioned?

```
pip3 install -r ./requirements.txt
python3 create_gpu_metrics.py
```

The 2 possible solutions I can think of is that.

1. There is an option in the init action that creates the metrics in Cloud Monitoring (not sure if this is possible due to auth and service account IAM issues)
2. Include the cmd to git clone the ml-on-gcp repo and run the python script to create the metrics (this is the option I included in my pull request)

There might be a better/easier solution also that I have missed.

medb:
@tfayyaz I think that your PR looks good, but may you delete [*generate-metrics* section](https://github.com/GoogleCloudDataproc/initialization-actions/tree/master/gpu#generate-metrics) or put your code in this section instead, so there will be only one place with these instructions?

tfayyaz:
Thanks @medb. I have updated the instructions and also made an update to fix cuda-url to cuda-version in the metadata parameters section.

I also added cloud-platform scope as just having monitoring.write scope blocks access to all other APIs. (eg. bigquery scope)

medb:
Thanks! I have removed cloud-platform scope because it maybe too broad by default. But feel free to send a follow-up PR to add a notice that if users would like to process BQ data on GPU they need broader scope.

PR #790, opened 2020-07-29
[rapids] Fix XGBoost with PySpark

bradmiro:
- Fixed a bug that prevented xgboost from being found in a PySpark environment
- Modified Spark test to include check for xgboost installation


bradmiro:
Test failed due to Dask build timeout. Increased the time.

PR #794, opened 2020-07-30
Wait to POD termination to fix presubmit flakiness

medb:


cloudbuild/run-presubmit-on-k8s.sh
```diff
@@ -23,6 +23,13 @@ kubectl wait --for=condition=Ready "pod/${POD_NAME}" --timeout=600s
 
 kubectl logs -f "${POD_NAME}"
 
+# Wait until POD will be terminated
+wait_secs=60
+while ((wait_secs > 0)) && ! kubectl describe "pod/${POD_NAME}" | grep -q Terminated; do
+  sleep 5
+  wait_secs-=5
```
functicons, line 30: I think this is the wrong syntax, you might want to use `((wait_secs-=5))`.

medb: Done.

PR #798, opened 2020-08-05
Enable use of GCS bucket as workflow file source for Oozie

Gaurangi94:
Implemented based on [SO answer](https://stackoverflow.com/questions/61233551/how-to-use-gcs-bucket-as-workflow-file-source-for-oozie-in-dataproc/61237392#61237392).

Fixes #798

### DIFF
--- a/oozie/oozie.sh
+++ b/oozie/oozie.sh
@@ -141,7 +141,7 @@ function install_oozie() {
     --name 'google.cloud.auth.service.account.enable' --value 'true' \
     --description 'Whether to use a service account for GCS authorization.
     Setting this property to `false` will disable use of service accounts for
-    authentication.'
+    authentication.' \
     --clobber
   
   if [ -e /usr/lib/hadoop/lib/gcs-connector.jar ]
--- a/oozie/oozie.sh
+++ b/oozie/oozie.sh
@@ -125,7 +125,22 @@ function install_oozie() {
     --configuration_file "/etc/hadoop/conf/core-site.xml" \
     --name 'hadoop.proxyuser.oozie.groups' --value '*' \
     --clobber
+    
+  bdconfig set_property \
+    --configuration_file "/etc/oozie/conf/oozie-site.xml" \
+    --name 'oozie.service.HadoopAccessorService.supported.filesystems' --value "hdfs,gs" \
+    --clobber
 
+  bdconfig set_property \
+    --configuration_file "/etc/hadoop/conf/core-site.xml" \
+    --name 'fs.AbstractFileSystem.gs.impl' --value 'com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS' \
+    --clobber
+    
+  bdconfig set_property \
+    --configuration_file "/etc/hadoop/conf/core-site.xml" \
+    --name 'google.cloud.auth.service.account.enable' --value 'false' \
+    --clobber
+  
   # Detect if current node configuration is HA and then set oozie servers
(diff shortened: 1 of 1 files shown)


oozie/oozie.sh
```diff
@@ -125,7 +125,30 @@ function install_oozie() {
...
+  bdconfig set_property \
+    --configuration_file "/etc/hadoop/conf/core-site.xml" \
+    --name 'google.cloud.auth.service.account.enable' --value 'true' \
+    --description 'Whether to use a service account for GCS authorization.
+    Setting this property to `false` will disable use of service accounts for
+    authentication.' \
+    --clobber
+  
+  if [ -e /usr/lib/hadoop/lib/gcs-connector.jar ]
+  then
```
Gaurangi94, line 148: @medb Can gcs-connector.jar be at any other location?

medb: Yes, it could be in the `/usr/local/share/google/dataproc/lib` on some image versions, you can add a check for this:
```bash
    local install_dir="/usr/local/share/google/dataproc/lib"
    if [[ ! -d \$install_dir ]]; then
      install_dir="/usr/lib/hadoop/lib"
    fi
```

oozie/oozie.sh
```diff
@@ -125,7 +125,30 @@ function install_oozie() {
...
+    --clobber
 
+  bdconfig set_property \
+    --configuration_file "/etc/hadoop/conf/core-site.xml" \
+    --name 'fs.AbstractFileSystem.gs.impl' --value 'com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS' \
+    --clobber
+    
+  bdconfig set_property \
+    --configuration_file "/etc/hadoop/conf/core-site.xml" \
+    --name 'google.cloud.auth.service.account.enable' --value 'true' \
```
medb, line 141: This is a default value, so we may just remove this property.

oozie/oozie.sh
```diff
@@ -125,7 +125,30 @@ function install_oozie() {
...
     --clobber
+    
+  bdconfig set_property \
+    --configuration_file "/etc/oozie/conf/oozie-site.xml" \
+    --name 'oozie.service.HadoopAccessorService.supported.filesystems' --value "hdfs,gs" \
+    --clobber
 
+  bdconfig set_property \
+    --configuration_file "/etc/hadoop/conf/core-site.xml" \
+    --name 'fs.AbstractFileSystem.gs.impl' --value 'com.google.cloud.hadoop.fs.gcs.GoogleHadoopFS' \
```
medb, line 136: Move `--value` parameter on a new line.

oozie/oozie.sh
```diff
@@ -125,7 +125,29 @@ function install_oozie() {
...
+  local gcs_connector_dir="/usr/local/share/google/dataproc/lib"
+  if [[ ! -d \$install_dir ]]; then
+    gcs_connector_dir="/usr/lib/hadoop/lib"
+  fi
+  
+  local gcs_connector_path="$gcs_connector_dir/gcs-connector.jar"
+  if [ -e $gcs_connector_path ]
+  then
+    cp $gcs_connector_path /usr/lib/oozie/lib/gcs-connector.jar
+  fi
```
medb, line 149: I think it's better to remove this check and just try to copy GCS connector - if it fails it means that configuration/script is wrong and we need to fix it.

Gaurangi94: done

Issue #791, opened 2020-07-29
GPU and Rapid error - Could not initialize class ml.dmlc.xgboost4j.java.XGBoostSparkJNI


tfayyaz:
When I train the model using GPUs from the notebook https://github.com/NVIDIA/spark-xgboost-examples/blob/spark-2/examples/notebooks/python/mortgage-gpu.ipynb it fails and the following error appears in the spark history server.

```
Job aborted due to stage failure: Task 0 in stage 2.0 failed 4 times, most recent failure: Lost task 0.3 in stage 2.0 (TID 5, jupyter-gpu-cluster-1-4-v4-w-1.us-central1-f.c.datalake-demo-admin.internal, executor 2): java.lang.NoClassDefFoundError: Could not initialize class ml.dmlc.xgboost4j.java.XGBoostSparkJNI
	at ml.dmlc.xgboost4j.scala.spark.XGBoost$.ml$dmlc$xgboost4j$scala$spark$XGBoost$$appendGpuIdToParameters(XGBoost.scala:585)
	at ml.dmlc.xgboost4j.scala.spark.XGBoost$$anonfun$ml$dmlc$xgboost4j$scala$spark$XGBoost$$trainForGpuDataset$1.apply(XGBoost.scala:614)
	at ml.dmlc.xgboost4j.scala.spark.XGBoost$$anonfun$ml$dmlc$xgboost4j$scala$spark$XGBoost$$trainForGpuDataset$1.apply(XGBoost.scala:613)
	at ml.dmlc.xgboost4j.scala.spark.rapids.GpuDataset$$anonfun$ml$dmlc$xgboost4j$scala$spark$rapids$GpuDataset$$getBatchMapper$1.apply(GpuDataset.scala:487)
	at ml.dmlc.xgboost4j.scala.spark.rapids.GpuDataset$$anonfun$ml$dmlc$xgboost4j$scala$spark$rapids$GpuDataset$$getBatchMapper$1.apply(GpuDataset.scala:486)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitions$1$$anonfun$apply$23.apply(RDD.scala:823)
	at org.apache.spark.rdd.RDD$$anonfun$mapPartitions$1$$anonfun$apply$23.apply(RDD.scala:823)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
[...]

PR #799, opened 2020-08-10
[rapids][gpu] fix RAPIDS README and GPU default Debian links to support CUDA 10.1

bradmiro:
fix documentation for RAPIDS to include cuda 10.1 with Dataproc 1.5, update tests
add cuda 10.1 debian link to GPU init action, update tests

fixes https://github.com/GoogleCloudDataproc/initialization-actions/issues/791

### DIFF
--- a/gpu/BUILD
+++ b/gpu/BUILD
@@ -8,7 +8,7 @@ py_test(
     srcs = ["test_gpu.py"],
     data = ["install_gpu_driver.sh"],
     local = True,
-    shard_count = 9,
+    shard_count = 12,
     deps = [
         "//integration_tests:dataproc_test_case",
         "@io_abseil_py//absl/testing:parameterized",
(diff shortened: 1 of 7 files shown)


bradmiro:
I think the test failure here is a flake?

medb:
> I think the test failure here is a flake?

No, seems like legit failure of the init action, maybe it flaked, but this is not test infrastructure flakiness.

bradmiro:
Ahh yes apologies, it looks like it may have actually timed out. I just tried running that specific build locally and it finished with more than enough time to spare. Can we try running it again?

medb:
> Ahh yes apologies, it looks like it may have actually timed out. I just tried running that specific build locally and it finished with more than enough time to spare. Can we try running it again?

Sure, let's check if it succeeds, re-triggered.

PR #800, opened 2020-08-19
feat: add Dask initialization action

bradmiro:
Adding Dask initialization action with Dask, Dask-Yarn.


medb:
It seems that this initialization action combines Dask and RAPIDS init actions into one. I think that we should split them the same as we did with GPU init action. I.e. we need to extract Dask cluster installation from RAPIDS init action and use RAPIDS init action only to install and configure RAPIDS with Spark or Dask.

bradmiro:
I see, so to make sure I understand correctly, you're suggesting we make this init action a prereq for the RAPIDS init action if Dask is selected? There are some potential coupling issues with doing that since the RAPIDS init action would then need to repack the conda environment, but that's probably ok?

medb:
Yes, this is exactly what I meant, Dask init action should be a pre-requisite for RAPIDS init action when RAPIDS configured to run on Dask.

bradmiro:
Awesome, thanks for the solid suggestion. I'll make the changes :)

dask/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+**:warning: NOTICE:** See
+[best practices](/README.md#how-initialization-actions-are-used) of using
+initialization actions in production.
+
+## Creating Dataproc Cluster with Dask and Dask-Yarn
+
+*Note:* Using the [Anaconda](https://cloud.google.com/dataproc/docs/concepts/components/anaconda) component does not add these new libraries into the Dask environment. You must manually include the libraries you want via the metadata flag `CONDA_PACKAGES` flag. 
+
+The following command will create a create a
```
medb, line 17: Remove duplicate "create a"

dask/dask.sh
```diff
@@ -0,0 +1,159 @@
...
+readonly ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+
+mkdir -p ${CONDA_ENV_DIR} ${DASK_CONFIG_DIR}
+
+readonly CONDA_BASE_CHANNELS=(
+    "conda-forge"
+)
+
+readonly CONDA_BASE_PACKAGES=(
+    "dask-yarn=0.8.1"
```
medb, line 42: Use `~=` to lock only `major.minor` versions:
```
dask-yarn~=0.8
```

dask/dask.sh
```diff
@@ -0,0 +1,159 @@
...
+    "dask>=2.6.0"
+    "dill"
+    "ipykernel"
+    "fastparquet"
+    "gcsfs"
+    "pyarrow"
+)
+
+function install_conda_kernel() {
+  conda install -y nb_conda_kernels
```
medb, line 52: Why do we need this?

dask/dask.sh
```diff
@@ -0,0 +1,159 @@
...
+  fi
+
+
+  local channels_cmd=()
+  for channel in "${conda_channels[@]}"; do
+    channels_cmd+=("-c $channel")
+  done
+
+  # Update conda
+  conda update -y conda
```
medb, line 126: Conda is pinned in Dataproc, I don't think that we should update it.

dask/dask.sh
```diff
@@ -0,0 +1,159 @@
...
+
+  # Install conda-pack to package a conda environment
+  conda install -y -c conda-forge conda-pack
+
+  # Create a conda environment
+  cmd="conda create -y ${channels_cmd[*]} -n ${CONDA_ENV_NAME} ${conda_packages[*]}"
```
medb, line 132: Just call command directly, without `cmd` variable

dask/README.md
```diff
@@ -0,0 +1,80 @@
+# Dask
+
+[Dask](https://dask.org/) is an open source set of tools for parallelizing Python analytics tasks. It is developed in coordination with popular open source Python libraries including Numpy, Pandas and Scikit-Learn. 
+
+This initialization action will create a Conda environment bundled with Dask and [Dask-Yarn](https://yarn.dask.org/), a tool for running Dask jobs on a YARN cluster such as [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster. You can then take advantage of Dataproc's features such as scaling out jobs with [Autoscaling](https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/autoscaling).
```
medb, line 5: I think that RAPIDS init action also supports Dask, but in standalone mode, not on YARN. May we extract creation of Dask cluster from RAPIDS init action to this init action and add an option that allows users to chose what type of Dask installation users want to use (YARN or Standalone)?

dask/dask.sh
bradmiro: This allows the created conda environment to show up as a kernel option in Jupyter.

dask/README.md
bradmiro: Ack. I can add standalone to this init action but was planning to modify the RAPIDS init action in a future PR. Should I just bundle it here?

medb: No, I agree that it's better to split them. Also we will need to let NVIDIA know about RAPIDS init action changes, so they can update their tutorials too.

bradmiro: Agreed, sounds good!

bradmiro:
@medb believe I fixed everything

dask/dask.sh
```diff
@@ -0,0 +1,242 @@
...
+    "conda-forge"
+)
+
+CONDA_BASE_PACKAGES=(
+    "dask>=2.6.0"
+    "dill"
+    "ipykernel"
+    "fastparquet"
+    "gcsfs"
+    "pyarrow"
```
medb, line 64: This seems like extra packages that are not strictly needed to run Dask, I would prefer if we limit this init action to bare minimum that required to run Dask.

dask/dask.sh
```diff
@@ -0,0 +1,242 @@
...
+)
+
+if [[ "${DASK_RUNTIME}" == "yarn" ]]; then
+  CONDA_BASE_PACKAGES+=("dask-yarn~=0.8")
```
medb, line 68: This version condition will install `dask-yarn` packages greater than 0.8 version but smaller than 1.0. Is this intentional?

dask/dask.sh
```diff
@@ -0,0 +1,242 @@
...
+# Config file for Dask Yarn.
+#
+# These values are joined on top of the default config, found at
+# https://yarn.dask.org/en/latest/configuration.html#default-configuration
+
+yarn:
+  environment: ${DASK_ENV_PACK}
+
+  worker: 
+    count: 2
```
medb, line 92: Why only 2 workers?

dask/dask.sh
```diff
@@ -0,0 +1,242 @@
...
+  # setup. Please see https://yarn.dask.org/en/latest/quickstart.html#usage
+  # for information on tuning Dask-Yarn environments.
+  cat <<EOF >"${DASK_CONFIG_FILE}"
+# Config file for Dask Yarn.
+#
+# These values are joined on top of the default config, found at
+# https://yarn.dask.org/en/latest/configuration.html#default-configuration
+
+yarn:
+  environment: ${DASK_ENV_PACK}
```
medb, line 89: Do we need packaged conda env to run on YARN? Can we just install everything required by Dask in default environment instead?

medb (changes requested):

dask/dask.sh
bradmiro: For dask-yarn specifically, [yes](https://yarn.dask.org/en/latest/quickstart.html#install-dask-yarn-on-an-edge-node).

dask/dask.sh
bradmiro: This is a baseline. In practice, the user will want to [customize](https://yarn.dask.org/en/latest/quickstart.html#usage) to their use case.

dask/dask.sh
bradmiro: I can remove `dill` here, but I think the rest are useful to include:
`ipykernel` - required to add the new environment as a Jupyter kernel
`gcsfs`- [required](https://yarn.dask.org/en/latest/quickstart.html#usage) for interacting with GCS
'pyarrow` / `fastparquet` - [required](https://yarn.dask.org/en/latest/quickstart.html#usage) for interacting with parquet data

dask/dask.sh
bradmiro: Some of the features of `dask-yarn` are experimental, so I think re-evaluating it upon a major version update is a good idea. What would you suggest for this use case, if not this?

dask/dask.sh
medb: I think that we should use default environment, because Dataproc ensures that conda env is the same on all nodes already:
https://yarn.dask.org/en/latest/environments.html#using-python-environments-local-to-each-node

dask/dask.sh
medb: I think it's better to add them to a default Dataproc conda environment and use it or let users use python init action to install them on all nodes - it's harder to support init action if it installs optional packages.

dask/dask.sh
medb: If we are pinning package then we should pin major.minor version, so we can pin it to 0.8 version for Dataproc 1.5: `dask-yarn=0.8`

For 2.0 preview version we can unpin it until it in preview

dask/test_dask.py
```diff
@@ -0,0 +1,58 @@
...
+        self.assert_instance_command(name, verify_cmd)
+        self.remove_test_script(script, name)        
+
+    @parameterized.parameters(
+        ("STANDARD", ["m"], None),
+        ("STANDARD", ["m"], "yarn"),
+        ("STANDARD", ["m"], "standalone"))
+    def test_dask(self, configuration, instances, runtime):
+        if self.getImageVersion() < pkg_resources.parse_version("1.5"):
+            return
```
medb, line 38: Use `self.skipTest("Not supported in pre 1.5 images")` instead of `return`

bradmiro: fixed

dask/dask.sh
bradmiro: Fixed

dask/dask.sh
bradmiro: Fixed

dask/dask.sh
bradmiro: Fixed this, only included a couple of these for Dataproc 1.5

bradmiro:
@medb changes acknowledged and implemented

dask/dask.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# Dask 'standalone' config
+readonly DASK_LAUNCHER='/usr/local/bin/dask-launcher.sh'
+readonly DASK_SERVICE='dask-cluster'
+
+CONDA_PACKAGES=("dask")
+
+if [ "$(echo "$DATAPROC_VERSION < 2.0" | bc)" -eq 1 ]; then
+  CONDA_PACKAGES+=("fastavro" "fastparquet" "gcsfs")
+fi
```
medb, line 42: I think this is redundant, it's nice to have, but let's let users use `python` init action to install extra packages, we can document it in the readme.

dask/dask.sh
```diff
@@ -0,0 +1,148 @@
...
+# limitations under the License.
+
+# This initialization action script will install Dask and other relevant
+# libraries on a Dataproc cluster. This is supported for either "yarn" or
+# "standalone" runtimes Please see dask.org and yarn.dask.org for more
+# information.
+
+set -euxo pipefail
+
+readonly DEFAULT_CONDA_ENV=/opt/conda/miniconda3
```
medb, line 24: Can we determine this path at runtime using `conda` command?

dask/dask.sh
```diff
@@ -0,0 +1,148 @@
...
+    systemctl start "${DASK_SERVICE}"
+  else
+    echo "Unsupported Dask Runtime: ${DASK_RUNTIME}"
+    exit 1
+  fi
+
+  echo "Dask for ${DASK_RUNTIME} successfully initialized."
+}
+
+main
```
medb, line 148: Add a new line at the end

dask/verify_dask_standalone.py
```diff
@@ -0,0 +1,9 @@
+from dask.distributed import Client
+import dask.array as da
+
+import numpy as np
+
+client = Client("localhost:8786")
+
+x = da.sum(np.ones(5))
+x.compute()
```
medb, line 9: Add a new line at the end

dask/verify_dask_yarn.py
```diff
@@ -0,0 +1,11 @@
...
+from dask.distributed import Client
+import dask.array as da
+
+import numpy as np
+
+cluster = YarnCluster()
+client = Client(cluster)
+
+x = da.sum(np.ones(5))
+x.compute()
```
medb, line 11: Add a new line at the end

medb (changes requested):

dask/verify_dask_yarn.py
bradmiro: done

dask/verify_dask_standalone.py
bradmiro: done

dask/dask.sh
bradmiro: done

dask/dask.sh
bradmiro: done

dask/dask.sh
bradmiro: done

medb (changes requested):

dask/dask.sh
```diff
@@ -0,0 +1,143 @@
...
+# limitations under the License.
+
+# This initialization action script will install Dask and other relevant
+# libraries on a Dataproc cluster. This is supported for either "yarn" or
+# "standalone" runtimes Please see dask.org and yarn.dask.org for more
+# information.
+
+set -euxo pipefail
+
+readonly DEFAULT_CONDA_ENV=$(conda info --envs | sed -n 's/base[[:blank:]*]*\(.*\)/\1/p')
```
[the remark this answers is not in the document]
bradmiro: Whoops, fixed.

dask/dask.sh
```diff
@@ -0,0 +1,143 @@
...
+
+yarn:
+  environment: python://${DEFAULT_CONDA_ENV}/bin/python
+
+  worker: 
+    count: 2
+EOF
+}
+
+function install_conda_packages() {
```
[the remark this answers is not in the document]
bradmiro: Fixed

medb (approved):

Issue #774, opened 2020-07-06
h2o init. action breaks with 1.5-ubuntu


bradmiro:
Integration test fails with 1.5-ubuntu, haven't fully figured out what's wrong. The testing script itself may just need an update:
```
Traceback (most recent call last):
  File "/tmp/43bb2a99b75d4097a2df49c8d504a63f/sample-script.py", line 6, in <module>
    hc = H2OContext.getOrCreate()
  File "/opt/conda/default/lib/python3.7/site-packages/ai/h2o/sparkling/H2OContext.py", line 89, in getOrCreate
    selected_conf = H2OConf()
  File "/opt/conda/default/lib/python3.7/site-packages/ai/h2o/sparkling/H2OConf.py", line 35, in __init__
    self._jconf = _jvm.org.apache.spark.h2o.H2OConf()
  File "/usr/lib/spark/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1525, in __call__
  File "/usr/lib/spark/python/lib/pyspark.zip/pyspark/sql/utils.py", line 63, in deco
  File "/usr/lib/spark/python/lib/py4j-0.10.7-src.zip/py4j/protocol.py", line 328, in get_return_value
py4j.protocol.Py4JJavaError: An error occurred while calling None.org.apache.spark.h2o.H2OConf.
: java.lang.NoClassDefFoundError: org/apache/spark/internal/Logging$class
	at ai.h2o.sparkling.utils.SparkSessionUtils$.<init>(SparkSessionUtils.scala:32)
	at ai.h2o.sparkling.utils.SparkSessionUtils$.<clinit>(SparkSessionUtils.scala)
	at org.apache.spark.h2o.H2OConf.<init>(H2OConf.scala:55)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at py4j.reflection.MethodInvoker.invoke(MethodInvoker.java:247)
	at py4j.reflection.ReflectionEngine.invoke(ReflectionEngine.java:357)
	at py4j.Gateway.invoke(Gateway.java:238)
[...]

PR #808, opened 2020-09-10
[h2o] update h2o version and deprecate Dataproc 1.5

bradmiro:
Removed support for 1.5 due to #774 . Now fails with an error message asking the user to use 1.3, 1.4 or 2.0+

Updated README to reflect this. 

Updated h2o version.

Fixed test script.

### DIFF
--- a/h2o/README.md
+++ b/h2o/README.md
@@ -25,7 +25,7 @@ Sparkling Water installed by:
     gcloud dataproc clusters create ${CLUSTER_NAME} \
         --image-version preview-ubuntu \
         --scopes "cloud-platform" \
-        --initialization-actions "gs://goog-dataproc-initialization-actions-${REGION}/conda/bootstrap-conda.sh,gs://goog-dataproc-initialization-actions-${REGION}/h2o/h2o.sh"
+        --initialization-actions "gs://goog-dataproc-initialization-actions-${REGION}/h2o/h2o.sh"
     ```
 
 1.  Submit sample job.
--- a/h2o/README.md
+++ b/h2o/README.md
@@ -4,8 +4,9 @@ This [initialization action](https://cloud.google.com/dataproc/init-actions)
 installs
 [H2O Sparkling Water](http://docs.h2o.ai/sparkling-water/2.4/latest-stable/doc/deployment/sw_google_cloud_dataproc.html)
 on all nodes of [Google Cloud Dataproc](https://cloud.google.com/dataproc)
-cluster. This initialization works with Dataproc image version `1.3` or newer,
-with an exception of version `1.5`. Please file an [issue](https://github.com/GoogleCloudDataproc/initialization-actions/issues/new) if support for `1.5` is desired.
+cluster. This initialization works with Dataproc image version `2.0` or newer.
+Please file an [issue](https://github.com/GoogleCloudDataproc/initialization-actions/issues/new)
+if support for older versions is desired.
 
 ## Using this initialization action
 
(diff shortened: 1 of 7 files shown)


h2o/h2o.sh
```diff
@@ -54,6 +54,11 @@ EOF
 }
 
 function main() {
+  if [[ "${DATAPROC_VERSION}" == "1.5" ]]; then
```
medb, line 57: Usually we put such conditions at very top of the init action: https://github.com/GoogleCloudDataproc/initialization-actions/blob/85e6216b9c0d9c6ea2f14b728b47862ab0e1bec1/conda/install-conda-env.sh#L4

h2o/h2o.sh
```diff
@@ -54,6 +54,11 @@ EOF
...
+    echo "Dataproc 1.5 not supported. Please use Dataproc 1.3, 1.4 or 2.0+"
```
medb, line 58: I think we need to add similar condition to skip test for 1.5 images, if we have it?

medb (changes requested):

h2o/h2o.sh
bradmiro: Fixed

h2o/h2o.sh
bradmiro: Fixed

medb:
Seems like you need to rebase (not merge) on top of master to fix presubmit.

bradmiro:
Think that should have fixed it.

bradmiro:
Skip logic was flawed; fixed.

medb:
Seems like tests are failing for older images

bradmiro:
I made a change to the test script that may have fixed the issue.

medb:
Plase, revert RAPIDS init action changes.

bradmiro:
Sorry about that, just rebased and removed the commit.

h2o/sample-script.py
```diff
@@ -3,7 +3,7 @@
 from pysparkling import *
 
 spark = SparkSession.builder.appName("SparklingWaterApp").getOrCreate()
-hc = H2OContext.getOrCreate(spark)
+hc = H2OContext.getOrCreate()
```
medb, line 6: Should we revert this?

medb (changes requested):

h2o/sample-script.py
bradmiro: I didn't think this was necessary but I'll add it back and see if that helps.

bradmiro:
These tests have been flaky for me locally with the same error that showed up here. I added the spark context back, can we try running again and see what happens?

h2o/sample-script.py
medb: It still failing, maybe try to increase cluster nodes size? By default they are configured to `e2-standard-2` machine type.

bradmiro: Done.

medb: Still does not work, maybe let's ignore all image versions except 2.0?

bradmiro: Work for me, I'll make the change

bradmiro: Just pushed

h2o/README.md
```diff
@@ -15,34 +17,17 @@ initialization actions in production.
...
 
     ```bash
     REGION=<region>
     CLUSTER_NAME=<cluster_name>
     gcloud dataproc clusters create ${CLUSTER_NAME} \
-        --image-version 1.3 \
-        --metadata 'H2O_SPARKLING_WATER_VERSION=3.28.0.1-1' \
+        --image-version preview-ubuntu \
         --scopes "cloud-platform" \
         --initialization-actions "gs://goog-dataproc-initialization-actions-${REGION}/conda/bootstrap-conda.sh,gs://goog-dataproc-initialization-actions-${REGION}/h2o/h2o.sh"
```
medb, line 28: Remove conda init action.

medb (changes requested):

h2o/sample-script.py
medb: Seems like it does not work even on 2.0

bradmiro: Seemed like it was a "dataproc internal error", maybe a flake? This has consistently passed for me locally on 2.0.

h2o/README.md
bradmiro: Done

h2o/sample-script.py
medb: Seems like presubmit passed for all image version except 1.5 after https://github.com/GoogleCloudDataproc/initialization-actions/pull/811

medb (approved):

PR #819, opened 2020-10-05
[gpu][rapids] Downgrade default CUDA version to 10.2

sameerz:
Signed-off-by: Sameer Raheja <sraheja@nvidia.com>

After making the default CUDA version 11.0 for the rapids.sh initialization scripts, we realized that one of our components, RAPIDS Spark XGBoost, is not yet working with 11.0.  We would like to change the default to 10.2 until this component is working with 11.0, which will happen by the end of the year.  

cc: @mengdong @garyshen2008


bradmiro:
Maybe we should revert the GPU init action to 10.2 as well?

medb (approved):

PR #823, opened 2020-10-23
Updating ubuntu key before running apt update

davidrabinowitz:


functicons (approved):

hue/hue.sh
```diff
@@ -41,6 +41,10 @@ function retry_apt_command() {
 }
 
 function update_apt_get() {
+  if grep ubuntu /etc/apt/sources.list > /dev/null 2>&1; then
```
medb, line 44: I do not think that we should do this because we do not do this in other init actions.

Also, we usually check OS like this:
```bash
if [[ "$(. /etc/os-release && echo "${ID}")" == 'ubuntu' ]]; then
```

medb (changes requested):

davidrabinowitz:
Closing as the issue this PR meant to solve is library compatibility, not missing key

PR #824, opened 2020-10-26
Support bigtable init action for CentOS

animeshnandanwar:


medb:
Bigtable init action does not work on 1.5+ clusters yet

PR #825, opened 2020-10-26
Add CentOS support for flink init action

animeshnandanwar:


medb:
Flink is supported via [Dataproc Component](https://cloud.google.com/dataproc/docs/concepts/components/flink) already available on CentOS.

PR #826, opened 2020-10-26
Support tez init action for CentOS

animeshnandanwar:


medb:
Tez is enabled by default in 1.3+ images and already available in all CentOS images.

Issue #830, opened 2020-11-05
pinned package seems to create conflict for rapids 0.16


mengdong:
launch cluster with following command:
```
export REGION=us-central1
export ZONE=us-central1-a
export GCS_BUCKET=[some bucket]
export CLUSTER_NAME=[some cluster name]
export DASK_RUNTIME=standalone
export NUM_GPUS=1
export NUM_WORKERS=2

gcloud dataproc clusters create ${CLUSTER_NAME} \
 --zone $ZONE \
 --region $REGION \
 --image-version preview-ubuntu \
 --master-machine-type a2-highgpu-1g \
 --num-workers $NUM_WORKERS \
 --worker-machine-type a2-highgpu-1g \
 --optional-components JUPYTER \
 --initialization-actions gs://goog-dataproc-initialization-actions-${REGION}/gpu/install_gpu_driver.sh,gs://goog-dataproc-initialization-actions-${REGION}/dask/dask.sh,gs://goog-dataproc-initialization-actions-${REGION}/rapids/rapids.sh \
 --initialization-action-timeout=60m \
 --metadata rapids-runtime=DASK,dask-runtime=${DASK_RUNTIME} \
 --metadata gpu-driver-provider=NVIDIA,cuda-version=11.1,rapids-version=0.16 \
 --bucket ${GCS_BUCKET} \
 --subnet default \
 --enable-component-gateway 
```

error observed:
```
+ sed -i '/pyarrow .*/d' /opt/conda/miniconda3/conda-meta/pinned
+ /opt/conda/miniconda3/envs/mamba/bin/mamba install -y -c rapidsai -c nvidia -c conda-forge -c defaults cudatoolkit=11.1 rapids=0.16 -p /opt/conda/miniconda3

                  __    __    __    __
                 /  \  /  \  /  \  /  \
                /    \/    \/    \/    \
███████████████/  /██/  /██/  /██/  /████████████████████████
              /  / \   / \   / \   / \  \____
             /  /   \_/   \_/   \_/   \    o \__,
            / _/                       \_____/  `
            |/
        ███╗   ███╗ █████╗ ███╗   ███╗██████╗  █████╗
        ████╗ ████║██╔══██╗████╗ ████║██╔══██╗██╔══██╗
        ██╔████╔██║███████║██╔████╔██║██████╔╝███████║
        ██║╚██╔╝██║██╔══██║██║╚██╔╝██║██╔══██╗██╔══██║
[...]

PR #831, opened 2020-11-05
[RAPIDS] remove  conda from pinned packages

mengdong:
see https://github.com/GoogleCloudDataproc/initialization-actions/issues/830#issue-737056533


medb:
This bug was fixed in base image.

PR #852, opened 2021-01-16
[rapids] fixing RAPIDS dependency issue

bradmiro:
There is a dependency conflict with "icu" that conda is unable to solve on it's own. Uninstall it and reinstall with RAPIDS to get correct version and fix the environment.

Also, as we no longer pin PyArrow in 2.0, I removed the logic that was concerned with that.

@mengdong @sameerz FYI

### DIFF
--- a/integration_tests/dataproc_test_case.py
+++ b/integration_tests/dataproc_test_case.py
@@ -94,6 +94,8 @@ class DataprocTestCase(parameterized.TestCase):
                       beta=False,
                       master_accelerator=None,
                       worker_accelerator=None,
+                      image=None,
+                      image_version=None,
                       optional_components=None,
                       machine_type="e2-standard-2",
                       boot_disk_size="50GB"):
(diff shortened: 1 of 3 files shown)


bradmiro:
@mengdong do the RAPIDS jars support 3.1?

sameerz:
Our 0.3 version does not yet support Spark 3.1.  The 0.4 release, planned for early March, will.  In the meantime we need to update our docs to ask users to install Dataproc `--image=2.0.0-RC22-ubuntu18`

bradmiro:
> Our 0.3 version does not yet support Spark 3.1. The 0.4 release, planned for early March, will. In the meantime we need to update our docs to ask users to install Dataproc `--image=2.0.0-RC22-ubuntu18`

Thanks @sameerz . @medb are we able to accommodate this in the integration test?

medb:
I guess we can skip Spark integration test for now, but I will merge this PR because it does not break anything additional.

medb (approved):

PR #861, opened 2021-01-29
Add 1.5 CentOS 8 image to presumbit

animeshnandanwar:


1
```diff
@@ -0,0 +1,12 @@
+Support kafka init action for CentOS
```
medb, line 1: Revert this file

medb (changes requested):

1
animeshnandanwar: Done

medb (approved):

PR #868, opened 2021-02-16
[cloud-sql-proxy] Clean up, fix and remove redundant functionality

medb:


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -313,68 +332,21 @@ function run_validation() {
   fi
 }
 
-function configure_hive_warehouse_dir() {
-  # Wait for master 0 to create the metastore db if necessary.
-  run_with_retries run_validation
```
mikaylakonst, line 318: It used to be that masters 1 and 2 in HA clusters would call configure_hive_warehouse_dir, which would call run_validation, which would call systemctl restart hive-metastore. Now the metastore will never be restarted on masters 1 and 2. We should pull restarting the metastore out of the run_validation function (since it didn't make sense there in the first place) and move it down into main, making sure we do it for all 3 masters

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -313,68 +332,21 @@ function run_validation() {
...
-
-  local hiveserver_uri
-  hiveserver_uri=$(get_hiveserver_uri)
-  HIVE_WAREHOURSE_URI=$(beeline -u "${hiveserver_uri}" -e "describe database default;" |
-    sed '4q;d' | cut -d "|" -f4 | tr -d '[:space:]')
-
-  echo "Hive warehouse uri: $HIVE_WAREHOURSE_URI"
-
-  bdconfig set_property \
-    --name 'hive.metastore.warehouse.dir' \
```
mikaylakonst, line 328: It seems that now this property never gets set for any of the 3 masters. Is that okay? It seems like it might be okay because it looks like previously, it was being set after we restarted the hive-metastore service anyways, so maybe it wasn't actually being used?

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -313,68 +332,21 @@ function run_validation() {
...
-    --value "${HIVE_WAREHOURSE_URI}" \
-    --configuration_file /etc/hive/conf/hive-site.xml \
-    --clobber
-  echo "Updated hive warehouse dir"
-}
-
-function reload_function() {
```
mikaylakonst, line 335: nit: maybe add to commit message that we intend to move this into internal Dataproc startup so it doesn't just seem like we're deleting it for no reason. In the meantime, since we haven't done that yet, would it make sense to keep this around but only run it if hive-server2 is running?

mikaylakonst (changes requested):

cloud-sql-proxy/cloud-sql-proxy.sh
mikaylakonst: If we do need to keep setting this property, we could extract the warehouse uri via mysql CLI like:
`
 HIVE_WAREHOURSE_URI=$(mysql -u "${db_hive_user}" "${db_hive_password_parameter}" -Nse \	    mysql -u "${DB_ADMIN_USER}" "${DB_ADMIN_PASSWORD_PARAMETER}" -e \
      "SELECT DB_LOCATION_URI FROM ${metastore_db}.DBS WHERE NAME = 'default';")
`

You can pass `--host` and `--port` to mysql command to make mysql point at cloud sql proxy on master 0, but will need to do this with retries on master 1 and 2 in case master 0 isn't ready yet

medb: Yes, I think this is how it should be - users needs to configure it explicitly or Hive will take it from DB automatically.

cloud-sql-proxy/cloud-sql-proxy.sh
medb: Added call for `run_with_retries run_validation` explicitly for 1 and 2 masters.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -253,32 +281,23 @@ EOF
...
 
 function configure_sql_client() {
-  # Configure mysql client to talk to metastore
+  # Configure MySQL client to talk to metastore
   cat <<EOF >/etc/mysql/conf.d/cloud-sql-proxy.cnf
 [client]
 protocol = tcp
-port = ${metastore_proxy_port}
+port = ${METASTORE_PROXY_PORT}
 EOF
```
medb, line 289: @mikaylakonst Why do we do this only on 0 master?

mikaylakonst: No reason - I guess we assumed the mysql client would only be used on master 0? 

But yeah we should do this for all masters, and we should explicitly set host = master 0 hostname (by default it is localhost). That way you won't have to pass --host and --port to mysql commands

I think we uninstall mysql client on workers (?) so that's probably why we don't do this on them

cloud-sql-proxy/cloud-sql-proxy.sh
medb: Kept it if hive-server2 is running for backward compatibility.

cloud-sql-proxy/cloud-sql-proxy.sh
medb: Ack, will create a follow up PR for this then.

mikaylakonst (approved):
LGTM - thanks!

Issue #871, opened 2021-02-17
Livy with Python 3.X


wsmolak:
By default Livy uses Python 2.7. Due to deprecation of that branch only Python3 should be used for DataProc 1.4+

Solution:

Minoconda:
```
export PYSPARK_PYTHON=/opt/conda/miniconda3/bin/python3
export PYSPARK_DRIVER_PYTHON=/opt/conda/miniconda3/bin/python3
```

Anaconda:
```
export PYSPARK_PYTHON=/opt/conda/anaconda/bin/python3
export PYSPARK_DRIVER_PYTHON=/opt/conda/anaconda/bin/python3
```

Paths from: https://cloud.google.com/dataproc/docs/tutorials/python-configuration#image_version_14

medb:
May send a PR for Livy init action with this fix?

PR #872, opened 2021-02-17
#871 | livy with Python3

wsmolak:


wsmolak:
@googlebot I signed it!

PR #860, opened 2021-01-28
[mlvm] Add cuDNN to documentation and update libraries

bradmiro:
Adding cuDNN to mlvm documentation, updating libraries.

Running into the same HDFS issue as in https://github.com/GoogleCloudDataproc/initialization-actions/pull/858#issuecomment-768583503

PR #877, opened 2021-02-26
[rapids] Updating RAPIDS to 0.18 and removing force uninstall of 'icu'

bradmiro:
Updating RAPIDs to 0.18. This update resolves a dependency conflict that previously required force uninstalling the `icu` dependency. 

Should also resolve https://github.com/GoogleCloudDataproc/initialization-actions/pull/860

### DIFF
--- a/rapids/rapids.sh
+++ b/rapids/rapids.sh
@@ -13,7 +13,7 @@ readonly DEFAULT_SPARK_RAPIDS_VERSION="0.3.0"
 
 if [[ "${SPARK_VERSION_ENV}" == "3"* ]]; then
   readonly DEFAULT_CUDA_VERSION="10.2"
-  readonly DEFAULT_CUDF_VERSION="0.17"
+  readonly DEFAULT_CUDF_VERSION="0.18"
   readonly DEFAULT_XGBOOST_VERSION="1.3.0"
   readonly DEFAULT_XGBOOST_GPU_SUB_VERSION="0.1.0"
   # TODO: uncomment when Spark 3.1 jars will be released.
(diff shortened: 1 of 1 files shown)


bradmiro:
Updated the version in the wrong spot, should work now...

Spark jars have not been released yet. @mengdong @sameerz  if you have any info on when 0.18 jars will be released that would be great.

sameerz:
We are testing the 0.18 jar now and expect to have it available by Monday.  Can we hold off on updating this until we can also update `readonly DEFAULT_SPARK_RAPIDS_VERSION="0.3.0"` to 0.4.0, which will be sometime around March 10?  

The Spark RAPIDS 0.3.0 version is not tested with 0.18, but 0.4.0 will be.

bradmiro:
Given that both Spark and Dask are currently broken anyway, I'm inclined to say we should just merge this once the Spark jars are available as this PR will at least fix Dask. We can update the Spark RAPIDS version as soon as it becomes available.

bradmiro:
@sameerz @mengdong wanted to see if we're able to proceed here?

mengdong:
Hi @bradmiro , we will update to 0.18 when v0.4 release of spark rapids is out. In meanwhile, dask rapids could work with cudf 0.18.

bradmiro:
@mengdong ok, I modified this such that RAPIDS `0.18` is used with Dask and RAPIDS `0.17` is used with Spark so we can merge this and fix Dask. When the jars are ready, we can bump Spark to 0.18.

bradmiro:
@medb the Spark tests will remain broken until Spark-RAPIDS v0.4 is released. Should I adjust the tests to use 2.0.0-RC22 (latest image with Spark 3.0.x) instead of latest, for the time being? Or should we just merge?

medb:
Let's just merge then.

medb (approved):

PR #879, opened 2021-02-27
[gpu] clean up and fix 2.0 clusters w/ master-only GPUs

medb:
Changes:
- configure YARN NodeManager to use GPUs only if GPUs are attached to the node (fixes 2.0 clusters creation with GPUs attached only to master nodes)
- install NCCL only if cuDNN installed
- code clean up
- run tests for single-node clusters

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -111,15 +111,21 @@ function install_nvidia_nccl() {
 }
 
 function install_nvidia_cudnn() {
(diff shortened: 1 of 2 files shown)


bradmiro:
@mengdong Does XGBoost not need nccl?

bradmiro:
> @mengdong Does XGBoost not need nccl?

Nevermind, I think this is fine. LGTM

animeshnandanwar (approved):
LGTM

PR #885, opened 2021-03-06
[rapids] Fix a scenario for single-node cluster where NM is also running on Master.

viadea:
Fix a scenario for single-node-cluster where NM is also running on Master.
So we need to restart NM as well if NM is detected on Master.


medb:
Thank you for contribution!

medb (approved):

PR #890, opened 2021-03-19
create home dir for livy user

pleonovitch:
Need to have home dir for livy as in some cases it will contain **.cache** folder for **Python-Eggs** 
Missing home will cause spark-submit to fail


pleonovitch:
@googlebot I signed it!

PR #893, opened 2021-03-19
[livy] Create home dir for livy user

pleonovitch:
Current instructions do not create home for **livy** user.
It can create problems under some circumstances; for examples when $HOME/.cache/Python-Eggs is being created when dealing with python eggs.


medb:
Thank you for contribution!

medb (approved):

PR #931, opened 2021-09-10
[GPU][RAPIDS] fix typo in GPU and support CUDA 11.2 in RAPIDS

bradmiro:
* Fixing a typo in the GPU init action
* If CUDA 11 is provided, always downloads RAPIDS SPARK jars for 11.0 (until future JARS are released) in the event that say CUDA 11.1 or 11.2 are supplied.

@mengdong please take a look and confirm that the latter is also safe to do. My understanding of CUDA's backwards compatibility is that this should be fine.


medb (approved):

bradmiro:
Discussed with @mengdong who confirmed this is ok.

PR #940, opened 2021-10-20
[livy] Add configurable Livy session timeout through metadata

dominikkedziak:
Problem:
While doing long running jobs on Dataproc cluster a session is being halted after one hour (default Livy session timeout)

Solution:
Add configurable Livy session timeout by adding timeout value by passing it in _metadata_ values. Default value stays the same (1h)


medb:
Thank you for the contribution!

PR #943, opened 2021-11-04
[RAPIDS] Update rapids.sh

mengdong:
add big query connector jars to read from big query storage API

### DIFF
--- a/rapids/rapids.sh
+++ b/rapids/rapids.sh
@@ -106,7 +106,7 @@ function install_spark_rapids() {
     wget -nv --timeout=30 --tries=5 --retry-connrefused \
       "${nvidia_repo_url}/rapids-4-spark_2.12/${SPARK_RAPIDS_VERSION}/rapids-4-spark_2.12-${SPARK_RAPIDS_VERSION}.jar" \
       -P /usr/lib/spark/jars/
-    wget -nv --timeout=30 --tries=5 --retry-connrefused \  
+    wget -nv --timeout=30 --tries=5 --retry-connrefused \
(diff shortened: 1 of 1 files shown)


rapids/rapids.sh
```diff
@@ -106,6 +106,12 @@ function install_spark_rapids() {
     wget -nv --timeout=30 --tries=5 --retry-connrefused \
       "${nvidia_repo_url}/rapids-4-spark_2.12/${SPARK_RAPIDS_VERSION}/rapids-4-spark_2.12-${SPARK_RAPIDS_VERSION}.jar" \
       -P /usr/lib/spark/jars/
+    wget -nv --timeout=30 --tries=5 --retry-connrefused \
+      "https://repo1.maven.org/maven2/com/google/cloud/spark/spark3support_2.12/0.22.2/spark3support_2.12-0.22.2.jar" \
+      -P /usr/lib/spark/jars/
+    wget -nv --timeout=30 --tries=5 --retry-connrefused \
+      "https://repo1.maven.org/maven2/com/google/cloud/spark/spark-bigquery-with-dependencies_2.12/0.22.2/spark-bigquery-with-dependencies_2.12-0.22.2.jar" \
+      -P /usr/lib/spark/jars/
```
medb, line 114: I think it will be better if we will modify instructions to use connectors init action to install Spark BQ connector: https://github.com/GoogleCloudDataproc/initialization-actions/tree/master/connectors#bigquery-connectors

medb (changes requested):

mengdong:
ah, good point, I missed the instruction. Thanks!

Issue #944, opened 2021-11-08
Add support for Oozie on a CentOS cluster


sshetty007:
The initialization action currently does not install the necessary packages on CentOS.

Installing packages via `dnf` still has the issue of not starting the Oozie EmbeddedServer properly.

Please add support for Oozie on a CentOS cluster.

medb:
What version of Dataproc CentOS image do you use?

sshetty007:
We have tried this with 2.0-centos and 1.5.48-centos

sshetty007:
I have submitted a PR to include CentOS support. Please review and let met me know if any changes are required.
https://github.com/GoogleCloudDataproc/initialization-actions/pull/948

suchchal:
> I have submitted a PR to include CentOS support. Please review and let met me know if any changes are required. #948

I am running updated oozie.sh which supports CentOS cluster. Although it is able to install Oozie and Oozie client, it is throwing below error at the end while running ooziedb.sh:-

+ sudo -u oozie /usr/lib/oozie/bin/ooziedb.sh create -run
  setting OOZIE_DATA=/var/lib/oozie
  setting JETTY_PID_FILE=/var/run/oozie/oozie.pid
  setting OOZIE_CONFIG=/etc/oozie/conf
  setting OOZIE_LOG=/var/log/oozie
  setting OOZIE_DATA=/var/lib/oozie
  setting JETTY_PID_FILE=/var/run/oozie/oozie.pid
  setting OOZIE_CONFIG=/etc/oozie/conf
  setting OOZIE_LOG=/var/log/oozie

Validate DB Connection
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/hadoop/util/ReflectionUtils
        at org.apache.oozie.service.Services.setServiceInternal(Services.java:378)
        at org.apache.oozie.service.Services.<init>(Services.java:111)
        at org.apache.oozie.tools.OozieDBCLI.getJdbcConf(OozieDBCLI.java:173)
        at org.apache.oozie.tools.OozieDBCLI.createConnection(OozieDBCLI.java:915)
        at org.apache.oozie.tools.OozieDBCLI.validateConnection(OozieDBCLI.java:923)
        at org.apache.oozie.tools.OozieDBCLI.createDB(OozieDBCLI.java:192)
        at org.apache.oozie.tools.OozieDBCLI.run(OozieDBCLI.java:135)
[...]

PR #948, opened 2021-11-24
Issue 944: Added support for CentOS. Copy jars to hdfs to run example…

sshetty007:
PR for issue #944 (add CentOS support to init action). Changes include adding support for CentOS, copying dependent jars to run the Spark example in cluster mode and using the MySQL database.

### DIFF
--- a/oozie/oozie.sh
+++ b/oozie/oozie.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!bin/bash
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -27,14 +27,15 @@ set -euxo pipefail
 
 # Use Python from /usr/bin instead of /opt/conda.
 export PATH=/usr/bin:$PATH
+export DATAPROC_IMAGE=$(/usr/share/google/get_metadata_value image)
 
-function retry_apt_command() {
+function retry_command() {
   local cmd="$1"
   for ((i = 0; i < 10; i++)); do
     if eval "$cmd"; then
       return 0
     fi
-    sleep 5
+    sleep $((i * 5 ))
   done
   return 1
 }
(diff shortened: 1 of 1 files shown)


oozie/oozie.sh
```diff
@@ -27,14 +27,15 @@ set -euxo pipefail
...
 
-function retry_apt_command() {
+function retry_command() {
   local cmd="$1"
   for ((i = 0; i < 10; i++)); do
     if eval "$cmd"; then
       return 0
     fi
-    sleep 5
+    sleep $((i * 5 ))
```
medb, line 38: Revert, this will make backoff exponential, which we probably do not want to do.

oozie/oozie.sh
```diff
@@ -48,8 +49,30 @@ function install_oozie() {
...
+    ln -sf /usr/lib/hadoop-yarn/hadoop-yarn-common.jar /usr/lib/oozie/lib/
+    ln -sf /usr/lib/hadoop-yarn/hadoop-yarn-client.jar /usr/lib/oozie/lib/
+    ln -sf /usr/lib/hadoop-yarn/hadoop-yarn-server-common.jar /usr/lib/oozie/lib/
+    ln -sf /usr/lib/hadoop-yarn/hadoop-yarn-api.jar /usr/lib/oozie/lib/
+
+    ln -sf /usr/lib/hadoop-mapreduce/hadoop-mapreduce-client-jobclient.jar /usr/lib/oozie/lib/
+    ln -sf /usr/lib/hadoop-mapreduce/hadoop-mapreduce-client-app.jar /usr/lib/oozie/lib/
+    ln -sf /usr/lib/hadoop-mapreduce/hadoop-mapreduce-client-common.jar /usr/lib/oozie/lib/
+    ln -sf /usr/lib/hadoop-mapreduce/hadoop-mapreduce-client-core.jar /usr/lib/oozie/lib/
+    ln -sf /usr/lib/hadoop-mapreduce/hadoop-mapreduce-client-shuffle.jar /usr/lib/oozie/lib/
```
medb, line 71: Why is this necessary for CentOS, but other OSes?

oozie/oozie.sh
```diff
@@ -48,8 +49,30 @@ function install_oozie() {
   master_node=$(/usr/share/google/get_metadata_value attributes/dataproc-master)
 
   # Upgrade the repository and install Oozie
-  retry_apt_command "apt-get update"
-  retry_apt_command "apt-get install -q -y oozie oozie-client"
+  if [[ "$DATAPROC_IMAGE" == *"cen"* ]]; then
```
medb, line 52: Use `if command -v apt-get >/dev/null; then` check for consistency with other scripts.

medb (changes requested):
May you separate changes that are required to run this init action on CentOS as is from other improvements in different PRs?

PR #959, opened 2021-12-29
[rapids] Update cuDF jar version to 21.12.2

viadea:
Update cudf jar version to 21.12.2


viadea:
@mengdong @bradmiro this is to update the cudf jar version to 21.12.2 which has a fix.

medb (approved):

PR #963, opened 2022-01-28
Added Trino init actions for Dataproc. 

sametkaradag:


rikturr:
Hi @sametkaradag thank you for putting this together! It's exactly what I've been looking for getting trino running on dataproc. My org would be more comfortable with using the init script once its merged into this repo. Is there anything I can do to help get this PR merged?

samkaradag (approved):
/gcbrun

samkaradag (approved):

samkaradag (approved):

samkaradag (approved):

trino/test_trino.py
```diff
@@ -0,0 +1,126 @@
...
+
+    @parameterized.parameters(("SINGLE", ["m"], 1, 0))
+    def test_trino_custom_port(self, configuration, machine_suffixes,
+                                coordinators, workers):
+        if self.getImageOs() == 'centos':
+            self.skipTest("Not supported in CentOS-based images")
+
+        # Skip on 2.0+ version of Dataproc because it's not supported
+        if self.getImageVersion() >= pkg_resources.parse_version("2.0"):
+            self.skipTest("Not supported in 2.0+ images")
```
bradmiro, line 110: What are the limitations? Dataproc 1.X will no longer be supported by the end of 2022.

samkaradag: Ah, this is an obsolete code piece, there is no limitation and works fine with 2+ images.

samkaradag: Can you check?

trino/trino.sh
```diff
@@ -0,0 +1,279 @@
...
+  cat >/opt/trino-server/etc/catalog/jmx.properties <<EOF
+connector.name=jmx
+EOF
+
+  cat >/opt/trino-server/etc/catalog/memory.properties <<EOF
+connector.name=memory
+EOF
+  cat >/opt/trino-server/etc/catalog/bigquery.properties <<EOF
+connector.name=bigquery
+bigquery.project-id=set-your-project-id
```
bradmiro, line 162: I don't think we should rely on the user editing this file directly - what if it's a cli arg such as `bigquery-project-id` and read with `/usr/share/google/get_metadata_value attributes/bigquery-project-id`?

samkaradag: great point, I have added the project id of Data proc cluster and pushed the code:
] BQ_PROJECT_ID="$(/usr/share/google/get_metadata_value ../project/project-id)"

jayadeep-jayaraman:
Trino is already supported as an optional component starting 2.1 in dataproc. It is also available as prestosql in Dataproc 2.0. Is this change still needed?

PR #975, opened 2022-02-24
The install_gpu_driver script should restart services for yarn config…

abellina:
…urations to take effect

The following restart step is needed if the `install_gpu_driver.sh` script is used in isolation, otherwise Yarn isn't finding the gpu configurations (isolation) that we are setting up here.

@mengdong FYI

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -378,9 +378,15 @@ function main() {
     fi
 
     configure_gpu_exclusive_mode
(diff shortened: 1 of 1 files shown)


mengdong:
@bradmiro @medb @abellina is from our spark team and helping us to refactor init script, could you please help review? Thanks

abellina:
It has been a while now, I am not entirely sure how to proceed here. Can one of the admins please check and suggest a plan of action? I should be already in the CLA.

PR #988, opened 2022-05-02
[gpu] Fix bug where attached GPU resources not available after cluster creation

flacode:
Fixes issue [#987](https://github.com/GoogleCloudDataproc/initialization-actions/issues/987)

PR #994, opened 2022-05-06
[gpu] Avoid restarting services if not started yet

vinayakumarb:
In case of 2.0+ dataproc images services are always started after the custom init actions.

So, Trying to restart services in init stage results in failure as 'ready' marker file will not be present for services to successfully start.
And exit code of init script can be non-zero due to this.

Following changes are on top of #988.


medb:
Thanks for the PR, I have incorporated this fix with some changes in the original https://github.com/GoogleCloudDataproc/initialization-actions/pull/988 PR

vinayakumarb:
> Thanks for the PR, I have incorporated this fix with some changes in the original #988 PR

Thank you.

PR #997, opened 2022-05-13
Remove deprecated (and in some cases broken) init actions

ojarjur:
This change removes the obsolete init actions for Tez, Datalab, Jupyter, Jupyter2, and Jupyter SparkMonitor


cloudbuild/run-presubmit-on-k8s.sh
```diff
@@ -14,6 +14,7 @@ kubectl run "${POD_NAME}" \
   --image="${IMAGE}" \
   --pod-running-timeout=15m \
   --restart=Never \
+  --limits="cpu=500m,ephemeral-storage=2Gi,memory=2Gi" \
```
medb, line 17: This is not supported in new k8s version - we need use future-proof config, or remove it - otherwise it will break in the future.

cloudbuild/presubmit.sh
```diff
@@ -47,9 +47,18 @@ initialize_git_repo() {
...
 determine_tests_to_run() {
   # Infer the files that changed
+  mapfile -t DELETED_BUILD_FILES < <(git diff origin/master --name-only --diff-filter=D | grep BUILD)
   mapfile -t CHANGED_FILES < <(git diff origin/master --name-only)
+  echo "Deleted BUILD files: ${DELETED_BUILD_FILES[*]}"
   echo "Changed files: ${CHANGED_FILES[*]}"
 
+  # Run all tests if common directories modified by deleting files
+  if [[ -n "${DELETED_BUILD_FILES}" ]]; then
+    echo "All tests will be run: the following BUILD files '${DELETED_BUILD_FILES}' were removed"
```
medb, line 57: nit: use `${DELETED_BUILD_FILES[*]}` to print all files.

medb (changes requested):

cloudbuild/presubmit.sh
```diff
@@ -47,9 +47,18 @@ initialize_git_repo() {
 # to determine all changed files and looks for tests in directories with changed files.
 determine_tests_to_run() {
   # Infer the files that changed
+  mapfile -t DELETED_BUILD_FILES < <(git diff origin/master --name-only --diff-filter=D | grep BUILD)
   mapfile -t CHANGED_FILES < <(git diff origin/master --name-only)
+  echo "Deleted BUILD files: ${DELETED_BUILD_FILES[*]}"
   echo "Changed files: ${CHANGED_FILES[*]}"
 
+  # Run all tests if common directories modified by deleting files
+  if [[ -n "${DELETED_BUILD_FILES}" ]]; then
```
medb, line 56: Use array length check for non-emptiness:
```
if [[ ${#CHANGED_FILES[@]} -gt 0 ]]; then
  . . .
fi
```

medb (changes requested):

cloudbuild/presubmit.sh
ojarjur: Done

cloudbuild/run-presubmit-on-k8s.sh
ojarjur: Removed

cloudbuild/presubmit.sh
ojarjur: Done

medb (approved):

PR #998, opened 2022-06-02
[ranger] Example script to block GCE VM metadata server for Ranger GCS plugin

aalexx-S:


ranger-gcs-plugin/blackhole_vm_metadata_server.sh
```diff
@@ -0,0 +1,10 @@
+set -xe
+
+# Blackhole the VM metadata server except for root.
+iptables -A OUTPUT -d 169.254.169.254 -p tcp -m owner --uid-owner root -j ACCEPT
+iptables -A OUTPUT -d 169.254.169.254 -p tcp -j REJECT
+
+# Save current iptable setting to preserve it after reboot.
+mkdir -p /etc/iptables
+iptables-save >/etc/iptables/rules.v4
```
cyxxy, line 9: Does this make sure the iptables rules are automatically restored?

aalexx-S: done

ranger-gcs-plugin/README.md
```diff
@@ -0,0 +1,8 @@
+# Ranger GCS Plugin
```
cyxxy, line 1: No need to mention the plugin, just describe what this script is: blocking non-root user access to the metadata server.

ranger-gcs-plugin/blackhole_vm_metadata_server.sh
```diff
@@ -0,0 +1,17 @@
...
+# Blackhole the VM metadata server except for root.
+iptables -A OUTPUT -d 169.254.169.254 -p tcp -m owner --uid-owner root -j ACCEPT
+iptables -A OUTPUT -d 169.254.169.254 -p tcp -j REJECT
+
+# Install iptables-persistent to preserve the setting after reboot.
+# This section only works on Debian and Ubuntu (Debian based systems).
+echo iptables-persistent iptables-persistent/autosave_v4 boolean true | debconf-set-selections
+echo iptables-persistent iptables-persistent/autosave_v6 boolean true | debconf-set-selections
+apt-get update
+yes | apt-get install iptables-persistent
```
cyxxy, line 12: you can do "apt-get -y"

aalexx-S: I can only do it for L12. The other lines are required since it has a different prompt.

ranger-gcs-plugin/README.md
aalexx-S: This is about this folder. There may be other scripts here in the future

cyxxy: I don't think we should have an "umbrella" folder in the init action repo. The script should just be about what it does, in this case, blocking non-root user access to metadata server.

If we have another init action script that performs a different job which might be used for the plugin, just create a new folder for that job.

aalexx-S: The repo is sturctured into components, this may not fit the structure well

cyxxy (approved):

ranger/blackhole_vm_metadata_server.sh
```diff
@@ -0,0 +1,17 @@
+set -xe
+
+# Blackhole the VM metadata server except for root.
```
medb, line 3: May we use something more obvious then "blackhole"?

ranger/blackhole_vm_metadata_server.sh
```diff
@@ -0,0 +1,17 @@
...
+iptables -A OUTPUT -d 169.254.169.254 -p tcp -m owner --uid-owner root -j ACCEPT
+iptables -A OUTPUT -d 169.254.169.254 -p tcp -j REJECT
+
+# Install iptables-persistent to preserve the setting after reboot.
+# This section only works on Debian and Ubuntu (Debian based systems).
+echo iptables-persistent iptables-persistent/autosave_v4 boolean true | debconf-set-selections
+echo iptables-persistent iptables-persistent/autosave_v6 boolean true | debconf-set-selections
+apt-get update
+apt-get -y install iptables-persistent
```
medb, line 12: Aren't this already installed in all Dataproc images?

medb (changes requested):

ranger/blackhole_vm_metadata_server.sh
aalexx-S: We have iptabes save, but using that solely requires users to use iptables restore to restore the setting. iptables persistent basically do that for you when it is loaded into the OS on boot.

medb: Interesting, probably we need to use it in our own images then too...

ranger/blackhole_vm_metadata_server.sh
aalexx-S: `Block` maybe?

aalexx-S: done

medb (approved):

PR #1000, opened 2022-06-24
MySQL Integration testing for backward compatibility

dilipgodhia:


hue/hue.sh
```diff
@@ -190,14 +190,19 @@ EOF
   # Disable logging and set random secret key
   set +x
   sed -i "s/secret_key=.*/secret_key=$(random_string)/" /etc/hue/conf/hue.ini
-  set -x
+
+  db_mysqlroot_password="$(grep 'password=' /etc/mysql/my.cnf | sed 's/^.*=//')"
+  if [[ -z ${db_mysqlroot_password} ]]; then
```
functicons, line 195: Add quote "".

hue/hue.sh
```diff
@@ -190,14 +190,19 @@ EOF
...
+    db_mysqlroot_password="root-password"
+  fi
 
   # Create a database, give 'hue' user permissions
-  mysql -u root -proot-password -e "
+  mysql -u root -p${db_mysqlroot_password} -e "
```
functicons, line 200: Prefer `--password=<password>`.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -66,8 +66,9 @@ if [[ -n "${DB_ADMIN_PASSWORD_URI}" ]]; then
 else
   readonly DB_ADMIN_PASSWORD=''
 fi
+
 if [[ -z ${DB_ADMIN_PASSWORD} ]]; then
-  readonly DB_ADMIN_PASSWORD_PARAMETER=''
+  readonly DB_ADMIN_PASSWORD_PARAMETER='-p'
```
functicons, line 71: I'm a little confused about this change. I check MySQL CLI doc [1], if the flag is not set, the default is to send no password.

Also can you replace `-p<password>` with `--password=<password>` in this script, the latter is less error-prone.

https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_password

functicons (changes requested):

hue/hue.sh
dilipgodhia: quotes added

cloud-sql-proxy/cloud-sql-proxy.sh
dilipgodhia: Quotes added and changed -p to a safer --password=

functicons (approved):
LGTM

PR #1004, opened 2022-07-16
Fix presubmit flakiness

medb:


cloudbuild/run-presubmit-on-k8s.sh
```diff
@@ -13,23 +13,24 @@ gcloud container clusters get-credentials "${CLOUDSDK_CONTAINER_CLUSTER}"
 kubectl run "${POD_NAME}" \
   --image="${IMAGE}" \
   --pod-running-timeout=15m \
+  --requests='cpu=750m,memory=2Gi,ephemeral-storage=2Gi' \
```
bradmiro, line 16: I believe you would also need to update the limits as well with the same params `--limits='cpu=750m,memory=2Gi,ephemeral-storage=2Gi'`

medb: Limits are inherited from requests, I think - at least GKE updates them too as it.

PR #1006, opened 2022-07-23
[cloud-sql-proxy] Add support for Rocky and CentOS

functicons:
Debian/Ubuntu and Rocky/CentOS have different MySQL config dir and systemd service names.

### DIFF
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -332,19 +332,19 @@ EOF
 function initialize_metastore_db() {
   echo 'Initialzing DB for Hive metastore ...' >&2
   # Check if metastore is initialized.
-  if ! mysql -u "${DB_HIVE_USER}" "${DB_HIVE_PASSWORD_PARAMETER}" -e ''; then
-    mysql -u "${DB_ADMIN_USER}" "${DB_ADMIN_PASSWORD_PARAMETER}" -e \
+  if ! mysql -h 127.0.0.1 -P "${METASTORE_PROXY_PORT}" -u "${DB_HIVE_USER}" "${DB_HIVE_PASSWORD_PARAMETER}" -e ''; then
+    mysql -h 127.0.0.1 -P "${METASTORE_PROXY_PORT}" -u "${DB_ADMIN_USER}" "${DB_ADMIN_PASSWORD_PARAMETER}" -e \
       "CREATE USER '${DB_HIVE_USER}' IDENTIFIED BY '${DB_HIVE_PASSWORD}';"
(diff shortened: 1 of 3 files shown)


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -314,8 +314,16 @@ EOF
...
+  if [[ -d /etc/mysql/conf.d ]]; then
+    mysql_conf_dir=/etc/mysql/conf.d
+  elif [[ -d /etc/my.cnf.d ]]; then
+    mysql_conf_dir=/etc/my.cnf.d
+  else
+    err "Neither /etc/mysql/conf.d nor /etc/my.cnf.d exists"
+  fi
   # Configure MySQL client to talk to metastore
-  cat <<EOF >/etc/mysql/conf.d/cloud-sql-proxy.cnf
+  cat <<EOF >"${mysql_conf_dir}/cloud-sql-proxy.cnf"
```
medb, line 326: Please uncomment test on Rocky to make sure that change works: https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/cloud-sql-proxy/test_cloud_sql_proxy.py#L59-L60

medb (changes requested):

cloud-sql-proxy/cloud-sql-proxy.sh
functicons: Done.

cloud-sql-proxy/test_cloud_sql_proxy.py
```diff
@@ -56,9 +56,6 @@ def __submit_pyspark_job(self, cluster_name):
       'HA',
   )
   def test_cloud_sql_proxy(self, configuration):
```
medb, line 58: Seems like this test is failing during presubmit, may you fix it?

medb (changes requested):

cloud-sql-proxy/test_cloud_sql_proxy.py
medb: Seems like HA test is still failing on Debian 10 images.

medb (changes requested):

cloud-sql-proxy/test_cloud_sql_proxy.py
functicons: I couldn't reproduce the error. Maybe flaky? I'm rerunning the test.

medb: Seems like it consistently fails in presubmit

medb (approved):

PR #1015, opened 2022-09-10
Enable PostGres as metastore DB.

yigress:
Detect Cloud SQL instance type to act on the DB type
  Add retries of initialialize metastore db as there race condition in HA can cause failure
  Add logging and some code cleanning

### DIFF
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -21,8 +21,21 @@
 # Do not use "set -x" to avoid printing passwords in clear in the logs
 set -euo pipefail
 
-readonly ADDITIONAL_INSTANCES_KEY='attributes/additional-cloud-sql-instances'
+declare -A DEFAULT_DB_PORT=(['MYSQL']='3306' ['POSTGRES']='5432' ['SQL']='1433')
+declare -A DEFAULT_DB_ADMIN_USER=(['MYSQL']='root' ['POSTGRES']='postgres' ['SQL']='#TODO')
+declare -A DEFAULT_DB_PROTO=(['MYSQL']='mysql' ['POSTGRES']='postgresql' ['SQL']='sqlserver')
+declare -A DEFAULT_DB_DRIVER=(['MYSQL']='com.mysql.jdbc.Driver' ['POSTGRES']='org.postgresql.Driver' ['SQL']='com.microsoft.sqlserver.jdbc.SQLServerDriver')
 
+function err() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')] [$(hostname)]: $*" >&2
+  return 1
+}
+
+function log() {
+  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')] [$(hostname)]: $*" >&2
+}
+
+readonly ADDITIONAL_INSTANCES_KEY='attributes/additional-cloud-sql-instances'
 readonly PROXY_DIR='/var/run/cloud_sql_proxy'
 readonly PROXY_BIN='/usr/local/bin/cloud_sql_proxy'
 readonly INIT_SCRIPT='/usr/lib/systemd/system/cloud-sql-proxy.service'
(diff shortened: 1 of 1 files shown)


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -21,8 +21,21 @@
 # Do not use "set -x" to avoid printing passwords in clear in the logs
 set -euo pipefail
 
-readonly ADDITIONAL_INSTANCES_KEY='attributes/additional-cloud-sql-instances'
+declare -A DEFAULT_DB_PORT=(['MYSQL']='3306' ['POSTGRES']='5432' ['SQL']='1433')
+declare -A DEFAULT_DB_ADMIN_USER=(['MYSQL']='root' ['POSTGRES']='postgres' ['SQL']='#TODO')
```
functicons, line 25: Add a comment for '#TODO'.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -42,11 +55,54 @@ readonly ENABLE_PROXY_ON_WORKERS
...
+  if ! [[ "${metastore_instance}" =~ .+:.+:.+ ]]; then
+    err 'hive-metastore-instance must be of form project:region:instance'
+  fi
+  local project=${metastore_instance%*:*:*}
+  local instance=${metastore_instance##*:}
+  local database=$(gcloud sql instances describe --project=${project} ${instance} | grep 'databaseVersion')
+  if [[ -z "${database}" ]]; then
+    err 'Unable to find hive_metastore_instance'
+  fi
+  # Trim off version and whitespaces and use upper case
```
functicons, line 76: Add an example, "(original)" -> "(result)".

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -364,92 +420,124 @@ function run_validation() {
...
+    return
+  fi
+
+  log "Installing POSTGRES CLI ..."
+  if command -v apt >/dev/null; then
+    apt update && apt install postgresql-client -y
+  elif command -v yum >/dev/null; then
+    yum -y update && yum -y install postgresql
+  fi
+  echo "POSTGRES CLI installed" >&2
```
functicons, line 461: `log "POSTGRES CLI installed"`?

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -364,92 +420,124 @@ function run_validation() {
...
+function install_db_cli() {
+  case ${METASTORE_INSTANCE_TYPE} in
+    MYSQL)
+      install_mysql_cli
+      ;;
+    POSTGRES)
+      install_postgres_cli
+      ;;
+    SQL)
+      # TODO: add SQL support
```
functicons, line 473: If SQLServer is not supported yet, consider fail fast.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -476,11 +564,13 @@ function update_master() {
...
   start_cloud_sql_proxy
-  configure_sql_client
 
   if [[ $ENABLE_CLOUD_SQL_METASTORE == "true" ]]; then
-    install_mysql_cli
-    initialize_metastore_db
+    install_db_cli
+
+    # Retry as there may be failures due to race condition
+    run_with_retries initialize_metastore_db
```
functicons, line 572: What's the race condition? Should we call it only on m-0?

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -21,8 +21,21 @@
 # Do not use "set -x" to avoid printing passwords in clear in the logs
 set -euo pipefail
 
-readonly ADDITIONAL_INSTANCES_KEY='attributes/additional-cloud-sql-instances'
+declare -A DEFAULT_DB_PORT=(['MYSQL']='3306' ['POSTGRES']='5432' ['SQL']='1433')
```
functicons, line 24: Is it possible to do a conversion, replace SQL with SQLSERVER?

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -364,92 +422,125 @@ function run_validation() {
...
+  case ${METASTORE_INSTANCE_TYPE} in
+    MYSQL)
+      install_mysql_cli
+      ;;
+    POSTGRES)
+      install_postgres_cli
+      ;;
+    SQL)
+      # TODO: add SQL support
+      err 'Fail fast here if SQL support is not enabled.'
```
functicons, line 476: "SQLServer is not supported".

cloud-sql-proxy/cloud-sql-proxy.sh
functicons: I don't see an answer or change for this comment.

cloud-sql-proxy/cloud-sql-proxy.sh
yigress: good catch! I just created one sqlserver instance, the instance type is actually SQLSERVER. updated relevant info

cloud-sql-proxy/cloud-sql-proxy.sh
yigress: forgot to click the comment button earlier. 

The race condition is when all master nodes trying to create database user and database. From testing it seems postgres is more prone to this issue. There was a PR https://github.com/GoogleCloudDataproc/initialization-actions/pull/288 a while back trying to solve this issue, not sure why it is reverted, maybe inadvertently.  retry works here as the retry only checks whether db user or database exists so it is a fast check. only the node that creates the database gets to run hive --initSchema.
Alternatively we could try to pick a specific master node, but in case some master node is faster than others and only a few master nodes, retry works better IMO

functicons (approved):
LGTM

PR #1019, opened 2022-09-29
Add an initialization action for installing the Toree Jupyter kernel

ojarjur:
Dataproc clusters currently ship with the Spylon kernel for Scala notebooks.

This new init action enables you to setup Toree as an alternative kernel for Scala

### DIFF
--- a/toree/README.md
+++ b/toree/README.md
@@ -1,0 +1,43 @@
+# Apache Toree Kernel
+
+This init action installs the [Apache Toree](https://toree.apache.org/) kernel for
+Jupyter notebooks.
+
+This init action assumes that the Jupyter optional component and the `toree` pip package
+are both installed in the cluster.
+
+## Usage
+
+**:warning: NOTICE:** See [best practices](/README.md#how-initialization-actions-are-used) of using initialization actions in production.
+
+You can create a cluster with Jupyter and the Apache Toree kernel configured using a
+command like the following:
+
+### Dataproc 2.0 (and later) images
+
+```bash
+REGION=<region>
+CLUSTER_NAME=<cluster_name>
+gcloud dataproc clusters create ${CLUSTER_NAME?} \
+  --region ${REGION?} \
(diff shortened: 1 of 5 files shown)


toree/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+```bash
+REGION=<region>
+CLUSTER_NAME=<cluster_name>
+gcloud dataproc clusters create ${CLUSTER_NAME?} \
+  --region ${REGION?} \
+  --image-version=2.0 \
+  --enable-component-gateway \
+  --optional-components=JUPYTER \
+  --properties=dataproc:pip.packages='toree==0.5.0' \
```
KoopaKing, line 26: Is there a reason you need to specify this here instead of it getting installed in toree.sh?

KoopaKing (approved):

toree/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Dataproc 2.0 (and later) images
+
+```bash
+REGION=<region>
+CLUSTER_NAME=<cluster_name>
+gcloud dataproc clusters create ${CLUSTER_NAME?} \
+  --region ${REGION?} \
+  --image-version=2.0 \
+  --enable-component-gateway \
+  --optional-components=JUPYTER \
```
KoopaKing, line 25:
```
Component JUPYTER requires that component ANACONDA is also selected. (FAILED_PRECONDITION)
    [google.rpc.error_details_ext] { message: "Component JUPYTER requires that component
    ANACONDA is also selected." }'
```

toree/README.md
ojarjur: My thinking was that it was a best practice to explicitly specify the packages (including versions) to install at cluster creation time, as it allows you to change the versions without changing the init actions code.

However, I think it's reasonable to make this optional rather than strictly required, so I've updated the code to manually install version `0.5.0` if the package is not already installed.

KoopaKing (approved):

PR #1020, opened 2022-10-06
Add an initialization action to change the node manager's log directory path to local SSDs

datasherlock:
- If there are no local SSDs, the script will not add any property to yarn-site and the default boot disk path will be used
- If there are multiple mounts, it will create a comma separated list of paths for the property based on the MAX_MNT_DISK_FOR_LOGS setting.
- If MAX_MNT_DISK_FOR_LOGS is greater than the actual disks, then the actual disk count will be used. If not, the configuration will be honoured.
- The fluentd configuration file is modified to monitor the new path for pushing to Cloud Logging

### DIFF
--- a/logging/change_to_local_ssd.sh
+++ b/logging/change_to_local_ssd.sh
@@ -1,0 +1,100 @@
+#!/bin/bash
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS-IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This script sets the yarn.nodemanager.log-dirs value to the Local SSD mounts.
+
+# Dataproc configurations
+# MAX_MNT_DISK_FOR_LOGS is the maximum number of disks to be used for creating the logs
+# LOGPATH is the path under /mnt/<mount number> where the application logs are created
+
+readonly HADOOP_CONF_DIR='/etc/hadoop/conf'
+readonly MAX_MNT_DISK_FOR_LOGS=3
+readonly LOGPATH='hadoop/yarn/userlogs'
+
+function set_hadoop_property() {
+    local -r config_file=$1
+    local -r property=$2
+    local -r value=$3
+    bdconfig set_property \
+        --configuration_file "${HADOOP_CONF_DIR}/${config_file}" \
+        --name "${property}" --value "${value}" \
+        --clobber
+}
+
+# This function validates the MAX_MNT_DISK_FOR_LOGS variable and ensures that the value is capped at the max disks available
+function get_max_local_ssd() {
+    max_disk_no=0
+    for line in $(df -h --output=target); do
+        if [ ${line:0:4} = '/mnt' ]; then
+            disk_no=$(echo $line | cut -d "/" -f3)
+            if [ $disk_no ] >$max_disk_no; then
+                max_disk_no=${disk_no}
+            fi
+        fi
+    done
+    echo $((max_disk_no > MAX_MNT_DISK_FOR_LOGS ? MAX_MNT_DISK_FOR_LOGS : max_disk_no)) # Return the least of total disks and maximum disks to be used for logging
+}
+
+# This function constructs the log directory paths
+function construct_log_dirs() {
+    max_disk_no=$(get_max_local_ssd)
+    #echo "Max Disk = "$max_disk_no
+    disk_no=1
+    log_dirs=""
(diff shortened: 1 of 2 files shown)


datasherlock:
@jainyogesh  - Can you please help me find a reviewer for this PR?

jainyogesh:
@ranu010101 , @vinayakumarb, @medb   - can one of you please help review it or suggest another reviewr

logging/change_to_local_ssd.sh
```diff
@@ -0,0 +1,103 @@
...
+        --configuration_file "${HADOOP_CONF_DIR}/${config_file}" \
+        --name "${property}" --value "${value}" \
+        --clobber
+}
+
+# This function validates the MAX_MNT_DISK_FOR_LOGS variable and ensures that the value is capped at the max disks available
+function get_max_local_ssd() {
+    max_disk_no=0
+    for line in $(df -h --output=target); do
+        if [ ${line:0:4} = '/mnt' ]; then
```
jayadeep-jayaraman, line 42: Logic dependent on hard coded values should be avoided. We should be able to get the localssd information from the cluster configuration

logging/change_to_local_ssd.sh
```diff
@@ -0,0 +1,103 @@
...
+                max_disk_no=${disk_no}
+            fi
+        fi
+    done
+    echo $((max_disk_no > MAX_MNT_DISK_FOR_LOGS ? MAX_MNT_DISK_FOR_LOGS : max_disk_no)) # Return the least of total disks and maximum disks to be used for logging
+}
+
+# This function constructs the log directory paths
+function construct_log_dirs() {
+    max_disk_no=$(get_max_local_ssd)
```
jayadeep-jayaraman, line 54: Pls use double quotes as following "${get_max_local_ssd)"

logging/change_to_local_ssd.sh
```diff
@@ -0,0 +1,103 @@
...
+    while [ $disk_no -le $max_disk_no ]; do
+        log_dirs+="/mnt/"$disk_no"/"$LOGPATH","
+        disk_no=$((${disk_no} + 1))
+    done
+    echo $log_dirs | sed 's/,$//g'
+}
+
+function configure_yarn() {
+    ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
+    echo "Role = "$ROLE
```
jayadeep-jayaraman, line 67: Pls use {} to avoid globbing of values

logging/change_to_local_ssd.sh
```diff
@@ -0,0 +1,103 @@
...
+    logdirs=$(construct_log_dirs)
+    echo "Log dir = " $logdirs
```
jayadeep-jayaraman, line 69: Same comment as above

logging/change_to_local_ssd.sh
```diff
@@ -0,0 +1,103 @@
...
+    if [[ "${ROLE}" != 'Master' && ! -z $logdirs ]]; then
```
jayadeep-jayaraman, line 70: Same comment as above

logging/change_to_local_ssd.sh
```diff
@@ -0,0 +1,103 @@
...
+
+# Dataproc configurations
+# Modify these configurations as needed
+# MAX_MNT_DISK_FOR_LOGS is the maximum number of disks to be used for creating the logs
+# LOGPATH is the path under /mnt/<mount number> where the application logs are created
+readonly MAX_MNT_DISK_FOR_LOGS=3
+readonly LOGPATH='hadoop/yarn/userlogs'
+
+# DO NOT CHANGE THESE VARIABLES UNLESS THE SOURCE FRAMEWORKS MODIFY IT
+readonly FLUENTD_YARN_USERLOGS='/etc/google-fluentd/config.d/dataproc-yarn-userlogs.conf'
```
jayadeep-jayaraman, line 25: Is there anyother way to get this value without hardocoding it ?

logging/change_to_local_ssd.sh
datasherlock: As discussed, the dataproc metadata only exposes the types of disks and not the path itself. The hardcoding here is based off assumptions in GCE as well as Dataproc APIs

jayadeep-jayaraman (changes requested):

logging/change_to_local_ssd.sh
```diff
@@ -0,0 +1,103 @@
...
+    bdconfig set_property \
+        --configuration_file "${HADOOP_CONF_DIR}/${config_file}" \
+        --name "${property}" --value "${value}" \
+        --clobber
+}
+
+# This function validates the MAX_MNT_DISK_FOR_LOGS variable and ensures that the value is capped at the max disks available
+function get_max_local_ssd() {
+    max_disk_no=0
+    for line in $(df -h --output=target); do
```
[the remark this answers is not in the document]
datasherlock: df returns only the filesystems and the mounts points they are mounted on. Even if someone were to create a directory called 'mnt' in /, df wouldn't return the path unless there's a device/filesystem mounted at that mount point.

jayadeep-jayaraman (approved):

jayadeep-jayaraman (approved):

PR #1021, opened 2022-10-11
Fix Rstudio installation on Ubuntu 20.04

ahzaz:
Installing Rstudio fails with following on ubuntu 20.04. 
```
Err:12 http://cran.r-project.org/bin/linux/ubuntu focal-cran40/ InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 51716619E084DAB9
Reading package lists...
W: GPG error: http://cran.r-project.org/bin/linux/ubuntu focal-cran40/ InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 51716619E084DAB9
E: The repository 'http://cran.r-project.org/bin/linux/ubuntu focal-cran40/ InRelease' is not signed.
```

Adding siging key as per cran documentaiton

### DIFF
--- a/rstudio/rstudio.sh
+++ b/rstudio/rstudio.sh
@@ -79,6 +79,10 @@ if [[ "${ROLE}" == 'Master' ]]; then
   # Install RStudio Server
   REPOSITORY_KEY=95C0FAF38DB3CCAD0C080A7BDC78B2DDEABC47B7
   run_with_retries apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys ${REPOSITORY_KEY}
+  # https://cran.r-project.org/bin/linux/ubuntu/
+  if [[ "${OS_ID}" == "ubuntu" ]]; then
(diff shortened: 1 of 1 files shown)


rstudio/rstudio.sh
```diff
@@ -79,6 +79,10 @@ if [[ "${ROLE}" == 'Master' ]]; then
   # Install RStudio Server
   REPOSITORY_KEY=95C0FAF38DB3CCAD0C080A7BDC78B2DDEABC47B7
   run_with_retries apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys ${REPOSITORY_KEY}
+  # https://cran.r-project.org/bin/linux/ubuntu/
+  if [[ "${OS_ID}" == "ubuntu" ]]; then
+    wget -qO- https://cloud.r-project.org/bin/linux/ubuntu/marutter_pubkey.asc | sudo tee -a /etc/apt/trusted.gpg.d/cran_ubuntu_key.asc
```
medb, line 84: Use `curl -fsSL --retry-connrefused --retry 10 --retry-max-time 30 ... | ...` for consistency with other init actions and better reliability.

medb, line 84: Do we need `sudo`? I thought that init actions are run as `root` already.

medb (approved):

rstudio/rstudio.sh
ahzaz: Updated

PR #1018, opened 2022-09-21
[rapids]split spark-rapids and rapids init scripts

nvliyuan:
Signed-off-by: liyuan <yuali@nvidia.com>
restructure rapids init scripts, split spark-rapids and rapids to the individual script

PR #1023, opened 2022-10-20
[gpu] Updated GPU driver install script

cjac:
* capture distribution from /etc/os-release for use in logging

* create a matrix of supported nvidia kernel driver, cuda, cudnn and nccl versions

* simplify default cuda version specification

* exit early on cuda versions known to be unsupported

* remove unused dataproc node role variable

* allow user to specify gpu-driver-version from metadata argument

* simplify rocky nccl version delta on cuda 11.2

* provide reasonable default cudnn version

* include current kernel version in kernel-devel package name

* use mkdir -p to ensure idempotency

* add support for CUDA > 11.5 on Debian

### DIFF
--- a/gpu/test_gpu.py
+++ b/gpu/test_gpu.py
@@ -38,7 +38,7 @@ class NvidiaGpuDriverTestCase(DataprocTestCase):
                                      driver_provider):
     if self.getImageVersion() < pkg_resources.parse_version("2.0") or self.getImageOs() == "rocky":
       self.skipTest("Not supported in pre 2.0 or Rocky images")
-      
+
     metadata = None
     if driver_provider is not None:
       metadata = "gpu-driver-provider={}".format(driver_provider)
@@ -64,7 +64,7 @@ class NvidiaGpuDriverTestCase(DataprocTestCase):
                                      driver_provider):
     if self.getImageVersion() < pkg_resources.parse_version("2.0") or self.getImageOs() == "rocky":
       self.skipTest("Not supported in pre 2.0 or Rocky images")
-        
+
     metadata = "install-gpu-agent=false"
     if driver_provider is not None:
       metadata += ",gpu-driver-provider={}".format(driver_provider)
@@ -90,7 +90,7 @@ class NvidiaGpuDriverTestCase(DataprocTestCase):
                                   driver_provider):
     if self.getImageVersion() < pkg_resources.parse_version("2.0") or self.getImageOs() == "rocky":
       self.skipTest("Not supported in pre 2.0 or Rocky images")
-      
+
     metadata = "install-gpu-agent=true"
     if driver_provider is not None:
       metadata += ",gpu-driver-provider={}".format(driver_provider)
@@ -119,9 +119,15 @@ class NvidiaGpuDriverTestCase(DataprocTestCase):
   def test_install_gpu_cuda_nvidia(self, configuration, machine_suffixes,
                                    master_accelerator, worker_accelerator,
                                    cuda_version):
-    if self.getImageVersion() < pkg_resources.parse_version("2.0") or self.getImageOs() == "rocky":
-      self.skipTest("Not supported in pre 2.0 or Rocky images")
-        
+    image_os = self.getImageOs()
+
+    if self.getImageVersion() < pkg_resources.parse_version("2.0"):
+      self.skipTest("Not supported in pre 2.0 images")
+
+    if ( image_os == "rocky" and (cuda_version < 11.2 and cuda_version != 11.0) ) or \
+       ( image_os == "debian" and cuda_version < 11.1 ):
+      self.skipTest(f'CUDA version {cuda_version} is not supported on os {image_os}')
+
     metadata = "gpu-driver-provider=NVIDIA,cuda-version={}".format(cuda_version)
     self.createCluster(
         configuration,
@@ -143,7 +149,7 @@ class NvidiaGpuDriverTestCase(DataprocTestCase):
                                   driver_provider, zone):
     if self.getImageVersion() < pkg_resources.parse_version("2.0") or self.getImageOs() == "rocky":
       self.skipTest("Not supported in pre 2.0 or Rocky images")
-        
+
     self.createCluster(
         configuration,
         self.INIT_ACTIONS,
@@ -155,7 +161,7 @@ class NvidiaGpuDriverTestCase(DataprocTestCase):
         metadata=None,
         timeout_in_minutes=30,
         startup_script="gpu/mig.sh")
-        
+
     for machine_suffix in ["w-0", "w-1"]:
       self.verify_mig_instance("{}-{}".format(self.getClusterName(),
                                           machine_suffix))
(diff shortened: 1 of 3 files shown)


cjac:
tested with rocky8:
* without specifying versions
* with --metadata cuda-version=11.1
This required another initialization action to revert the kernel to a previous version
* with --metadata cuda-version=11.2
* with --metadata cuda-version=11.5
* with --metadata cuda-version=11.6
* with --metadata cudnn-version=8.0.5.39 \
    --metadata cuda-version=11.1 \
    --metadata nccl-version=2.8.4 \
    --metadata gpu-driver-version=455.45.01 \
    --metadata gpu-driver-url=https://download.nvidia.com/XFree86/Linux-x86_64/455.45.01/NVIDIA-Linux-x86_64-455.45.01.run \
    --metadata cuda-url=https://developer.download.nvidia.com/compute/cuda/11.1.0/local_installers/cuda_11.1.0_455.23.05_linux.run
This required another initialization action to revert the kernel to a previous version

cjac:
tested with debian10:
* without specifying versions
* with --metadata cuda-version=11.1
* with --metadata cuda-version=11.2
* with --metadata cuda-version=11.5
* with --metadata cudnn-version=8.0.5.39
--metadata cuda-version=11.1
--metadata nccl-version=2.8.4
--metadata gpu-driver-version=455.45.01
--metadata gpu-driver-url=https://download.nvidia.com/XFree86/Linux-x86_64/455.45.01/NVIDIA-Linux-x86_64-455.45.01.run
--metadata cuda-url=https://developer.download.nvidia.com/compute/cuda/11.1.0/local_installers/cuda_11.1.0_455.23.05_linux.run

cjac:
tested with 2.0-ubuntu18:
* without specifying versions
* with --metadata cuda-version=10.1
* with --metadata cuda-version=10.2
* with --metadata cuda-version=11.1
* with --metadata cuda-version=11.2
* with --metadata cuda-version=11.5
* with --metadata cuda-version=11.6
* with --metadata cudnn-version=8.0.5.39
--metadata cuda-version=11.1
--metadata nccl-version=2.8.4
--metadata gpu-driver-version=455.45.01
--metadata gpu-driver-url=https://download.nvidia.com/XFree86/Linux-x86_64/455.45.01/NVIDIA-Linux-x86_64-455.45.01.run
--metadata cuda-url=https://developer.download.nvidia.com/compute/cuda/11.1.0/local_installers/cuda_11.1.0_455.23.05_linux.run

nvliyuan:
> @nvliyuan - Can you pls confirm if the changes that you are making will take care of the changes being proposed in this PR?

since in this pr #1018 I merged the [install_gpu_driver.sh](https://github.com/nvliyuan/initialization-actions/blob/rapids-2208-restructure/gpu/install_gpu_driver.sh) and [install_gpu_driver.sh](https://github.com/nvliyuan/initialization-actions/blob/rapids-2208-restructure/rapids/rapids.sh) to [spark-rapids.sh](https://github.com/nvliyuan/initialization-actions/blob/rapids-2208-restructure/sparkRapids/spark-rapids.sh), so I think the new scripts is independent and will not impact other scripts

jayadeep-jayaraman:
The below line has been deleted in the new script and because of that we are getting the error

```
ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
readonly ROLE
```

```
/etc/google-dataproc/startup-scripts/dataproc-initialization-script-0: line 480: ROLE: unbound variable
```

jayadeep-jayaraman:
2.x rocky and 2.x ubuntu tests are passing along with all the 1.5 distro's.

For 2.x debian 10 there are multiple errors all pointing to invalid url addresses as shown below

```
+ curl -fsSL --retry-connrefused --retry 10 --retry-max-time 30 https://download.nvidia.com/XFree86/Linux-x86_64/450.51.06/NVIDIA-Linux-x86_64-450.51.06.run -o driver.run
curl: (22) The requested URL returned error: 404

Setting up cuda-keyring (1.0-1) ...
+ curl -fsSL --retry-connrefused --retry 10 --retry-max-time 30 https://download.nvidia.com/XFree86/Linux-x86_64/440.33.01/NVIDIA-Linux-x86_64-440.33.01.run -o driver.run
curl: (22) The requested URL returned error: 404

```

PTAL

cjac:
Thanks for the review, Jaydeep.  I'll re-run with debian10 and correct the failures.

cjac:
The 440 kernel driver doesn't compile on the debian kernel.  Has this CUDA version worked in the past?  What metadata arguments were you passing when you received the second 404?  You would have received that error if you used --metadata cuda-version=10.2.  I can patch the script to get past the 404, but then it will bail out during kernel driver build[1].

I could maybe get it working with a variaton on the rocky kernel revert script[2] I put together.  But it feels messy, and there has been no customer requesting the work, to my knowledge.

Kernel driver 450 (CUDA 11.0) similarly fails to build[3].

I should probably have adjusted the conditional that prints unsupported versions...

[1] https://web.c9h.org/~cjac/for-dataproc/nvidia-installer-440.64.log
[2] https://raw.githubusercontent.com/cjac/initialization-actions/revert-rocky-kernel-202210/util/revert-kernel.sh
[3] https://web.c9h.org/~cjac/for-dataproc/nvidia-installer-450.51.log

cjac:
for what it's worth, getting CUDA > 11.5 working on the debian10 images wouldn't be much work.  So I did it.

cjac:
I was able to get CUDA 10.2 and 11.0 working on debian10 with kernel 5.7, but more modern kernels changed various APIs that the NVidia kernel driver source depends on.

gpu/test_gpu.py
```diff
@@ -119,9 +119,15 @@ def test_install_gpu_with_agent(self, configuration, machine_suffixes,
...
                                    cuda_version):
-    if self.getImageVersion() < pkg_resources.parse_version("2.0") or self.getImageOs() == "rocky":
-      self.skipTest("Not supported in pre 2.0 or Rocky images")
-        
+    image_os = self.getImageOs()
+
+    if self.getImageVersion() < pkg_resources.parse_version("2.0"):
+      self.skipTest("Not supported in pre 2.0 images")
+
+    if ( image_os == "rocky" and (cuda_version < 11.2 and cuda_version != 11.0) ) or \
```
jayadeep-jayaraman, line 127: 11.2, 11.0 etc should be string.

cjac: oops.

gpu/test_gpu.py
```diff
@@ -119,9 +119,15 @@ def test_install_gpu_with_agent(self, configuration, machine_suffixes,
...
-    if self.getImageVersion() < pkg_resources.parse_version("2.0") or self.getImageOs() == "rocky":
-      self.skipTest("Not supported in pre 2.0 or Rocky images")
-        
+    image_os = self.getImageOs()
+
+    if self.getImageVersion() < pkg_resources.parse_version("2.0"):
+      self.skipTest("Not supported in pre 2.0 images")
+
+    if ( image_os == "rocky" and (cuda_version < "11.2" and cuda_version != "11.0") ) or \
+       ( image_os == "debian" and cuda_version < "11.1' ):
```
jayadeep-jayaraman, line 128: One side is double and other side is single quote. This needs to be fixed, both sides should have double quotes

cjac: yikes.  don't know how that happened.

jayadeep-jayaraman (approved):
LGTM

PR #1024, opened 2022-10-20
Do not fail when no metastore instances is set

yigress:
This fix covers the scenario when no hive-metastore-stance is specified, user may provide additional-cloud-sql-instances or not provide any instance at all, in such case we first check if additional-cloud-sql-instance is provided, if not, default to mysql for backward compatibility.

### DIFF
--- a/cloud-sql-proxy/cloud-sql-proxy.sh
+++ b/cloud-sql-proxy/cloud-sql-proxy.sh
@@ -27,12 +27,12 @@ declare -A DEFAULT_DB_PROTO=(['MYSQL']='mysql' ['POSTGRES']='postgresql' ['SQLSE
 declare -A DEFAULT_DB_DRIVER=(['MYSQL']='com.mysql.jdbc.Driver' ['POSTGRES']='org.postgresql.Driver' ['SQLSERVER']='com.microsoft.sqlserver.jdbc.SQLServerDriver')
 
 function err() {
-  echo "[$(date +'%Y-%m-%dT%H:%M:%S%z')] [$(hostname)]: $*" >&2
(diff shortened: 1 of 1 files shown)


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -58,37 +58,49 @@ readonly USE_CLOUD_SQL_PRIVATE_IP
...
 
-# CLOUD SQL instance type is one of MYSQL, POSTGRES, SQL
-METASTORE_INSTANCE_TYPE=$(get_metastore_instance_type)
+# CLOUD SQL instance type is one of MYSQL, POSTGRES, SQLSERVER.
+# Type inferred in order of METASTORE_INSTANCE, ADDITIONAL_INSTANCES, default to MYSQL
+if [[ -n "${METASTORE_INSTANCE}" ]]; then
+  METASTORE_INSTANCE_TYPE=$(get_metastore_instance_type "${METASTORE_INSTANCE}")
+elif [[ -n "${ADDITIONAL_INSTANCES}" ]]; then
+  METASTORE_INSTANCE_TYPE=$(get_metastore_instance_type "${ADDITIONAL_INSTANCES}")
+fi
```
functicons, line 98: As discussed offline, we should not get type from additional instances, instead add a metadata for type, if not provided, default to MYSQL.

functicons (changes requested):

cloud-sql-proxy/cloud-sql-proxy.sh
yigress: added metadata cloud-sql-instance-type that takes priority

PR #1028, opened 2022-11-02
[rapids]update the outofdate agent to collect GPU metrics

nvliyuan:
Signed-off-by: liyuan <yuali@nvidia.com>
this pr is to :
1. update the outofdate agent to collect GPU metrics
2. rename sparkRapids folder to spark-rapids
3. add readme to spark-rapids
4. verify 2.1.0-RC2 images
5. fix #1012 to support customized driver version

PR #1030, opened 2022-11-09
refactor gpu installer to be used by other scripts

cjac:
common functions and variables are defined in gpu/installer_common.sh

these common functions and variables can be sourced and used by other initialization actions which require some or all of the gpu packages to be installed


cjac:
@pulkit-jain-G, @nvliyuan, @viadea,  @mengdong 

What do you think about this?  There's a chance that the user won't be able to fetch that shared file.  Should we accept that risk, considering that it allows us to use common functions across different initialization actions?

cjac:
yikes.  we've got a bit of a chicken-before-the-egg problem here.  This is hacky, but will probably get the tests to pass.  I would want to know what the other members of the team think about committing this before doing so, though.

viadea:
> @pulkit-jain-G, @nvliyuan, @viadea, @mengdong
> 
> What do you think about this? There's a chance that the user won't be able to fetch that shared file. Should we accept that risk, considering that it allows us to use common functions across different initialization actions?

From Spark RAPIDS point of view, we are fine because we are going to use a new [spark-rapids.sh](https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/sparkRapids/spark-rapids.sh) that @nvliyuan  worked on.

cjac:
What I mean is that there is a lot of shared code between spark-rapids.sh and install_gpu_driver.sh

this shared function library could be sourced by spark-rapids.sh rather than re-writing all of the code that's already been tested during development of install_gpu_driver.sh

I have already run into a bug in spark-rapids.sh that was fixed in install_gpu_driver.sh, and I was hoping to avoid more of the same.

nvliyuan:
> this shared function library could be sourced by spark-rapids.sh rather than re-writing all of the code that's already been tested during development of install_gpu_driver.sh

@cjac yes, that makes sense. I can merge this pr to #1028 if this pr goes first.

cjac:
@jayadeep-jayaraman - I'd also like to hear your opinion on this one

jayadeep-jayaraman:
The changes look good, pls go ahead and merge the changes when you are ready

cjac:
I did not mean to close that.

cjac:
tests are failing because we use an obsolete argument, --requests in the script which kicks off the script.  See #1052 for details

cjac:
blocked on #1050

cjac:
I don't think this is important enough to maintain.

PR #1036, opened 2022-12-05
[rapids] remove duplicate folder

nvliyuan:
Signed-off-by: liyuan <yuali@nvidia.com>
Since #1028 merged, we should remove sparkRapids folder


viadea:
@cjac

cjac:
I think we may need to do some tidying up in order to get *all* of the tests to run.  But here's to hoping!

cjac:
jobs are timing out due to io.  I've increased the default pd size and gotten rid of all 'yum update' and 'dnf update' calls.

cjac:
looks like we don't have enough capacity for 1024G disks.  I've reduce the size to 50G again, but changed type from pd-standard to pd-ssd

nvliyuan:
Hi @cjac, not sure why we need these commits just remove a folder, anything I can do?

cjac:
Removing the folder unfortunately triggers a full run of all tests.  These
changes are fixes to the failing actions.  It would probably be better to
fix all of these in their own PRs.

nvliyuan:
> Removing the folder unfortunately triggers a full run of all tests.

Hi @cjac , can we disable this action? Not sure it is reasonable if we want to remove a deprecated folder that will trigger all test.

cjac:
I'm going to prioritize getting all of our tests passing all of the time
instead.  That way, we get both.  And more regular runs of these tests is
critical.  I'm helping to shape Dataproc's direction this coming year, so
we should be able to fit it in if I have my way.

It's been on my mind, as well.

cjac:
@jayadeep-jayaraman - is there any way to temporarily disable the test so we can just remove this excess directory?  I started going down the path of making sure the tests run for all initialization actions, and I didn't get very far.  But we know that deleting the un-used directory, sparkRapids, will not cause any tests to break.  Nothing depends on that directory.  There is just logic to cause commits which create or delete directories to run the entire suite of tests.  Maybe we could change the logic to not trigger any tests on directory removal instead of triggering *all* tests on directory removal.

jayadeep-jayaraman:
I don't think there is anyway today to disable the test. The logic [here](https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/cloudbuild/presubmit.sh#L56-L81) needs to be changed to handle deletions.

jayadeep-jayaraman:
I see that there are lot of changes which are not related to removing the sparkRapids folder. Can you please check the PR once again and remove unwanted changes. Example:- changes to h2o init script, cloud-sql-proxy init script etc.

cjac:
Yes, I'll go through and clean those up tonight.  Thank you for your help with this!

cjac:
sorry, my vacation took priority!

cjac:
This will take a while:

```
2023-02-25T11:01:29.800997373Z All tests will be run: the following BUILD files 'sparkRapids/BUILD' were removed
2023-02-25T11:01:29.801014673Z + echo 'All tests will be run: the following BUILD files '\''sparkRapids/BUILD'\'' were removed'
```

cjac:
well, it hasn't failed yet, which is a good sign.

cjac:
okay, that took too long.  @jayadeep-jayaraman - can you help me to increase the timeout for this deletion?

jayadeep-jayaraman:
Hi @cjac - did you increase the value of timeout in cloudbuild ? If yes can you pls submit that PR so that I can approve that first before testing this ?

cjac:
I did not.  I'll give that a go when I'm at my desk, thanks!

nvliyuan:
Hi @jayadeep-jayaraman , could we remove this folder now?

jayadeep-jayaraman (approved):

PR #1047, opened 2023-02-08
Fix node manager failure during script execution

jster1357:
current script fails. This fixes the issue where the node manager doesn't restart due to the insertion of new settings while the node manager is running.

### DIFF
--- a/prometheus/prometheus.sh
+++ b/prometheus/prometheus.sh
@@ -142,6 +142,13 @@ function start_services() {
 }
 
 function configure_hadoop() {
+
+if [[ "${ROLE}" == 'Master' ]]; then
+  stop_service_gracefully hadoop-yarn-resourcemanager.service
+  else
+    stop_service_gracefully hadoop-yarn-nodemanager.service
+  fi
+  
   cat <<EOF >/etc/hadoop/conf/hadoop-metrics2.properties
 resourcemanager.sink.statsd.class=org.apache.hadoop.metrics2.sink.StatsDSink
 resourcemanager.sink.statsd.server.host=${HOSTNAME}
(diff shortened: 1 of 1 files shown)


prometheus/prometheus.sh
```diff
@@ -142,6 +142,13 @@ function start_services() {
 }
 
 function configure_hadoop() {
+
+if [[ "${ROLE}" == 'Master' ]]; then
+  stop_service_gracefully hadoop-yarn-resourcemanager.service
+  else
+    stop_service_gracefully hadoop-yarn-nodemanager.service
+  fi
```
Arun-Krishna, line 150: Please fix the indentation

Arun-Krishna, line 150: Please fix the indentation

prometheus/prometheus.sh
```diff
@@ -156,10 +163,10 @@ mrappmaster.sink.statsd.skip.hostname=true
...
 EOF
 
-  if [[ "${ROLE}" == 'Master' ]]; then
-    restart_service_gracefully hadoop-yarn-resourcemanager.service
+if [[ "${ROLE}" == 'Master' ]]; then
+  start_service_gracefully hadoop-yarn-resourcemanager.service
   else
-    restart_service_gracefully hadoop-yarn-nodemanager.service
+    start_service_gracefully hadoop-yarn-nodemanager.service
   fi
```
Arun-Krishna, line 170: Please fix the indentation

Arun-Krishna (commented):
Is the issue related to restart not picking up the conf changes?

prometheus/prometheus.sh
jster1357: this is done

prometheus/prometheus.sh
jster1357: this is done

PR #1049, opened 2023-02-16
Add 2.1 images to presubmit

jayadeep-jayaraman:


jayadeep-jayaraman:
@cjac @medb - Can you please review the changes ?

cloudbuild/run-presubmit-on-k8s.sh
```diff
@@ -14,7 +14,6 @@ LOGS_SINCE_TIME=$(date --iso-8601=seconds)
 
 kubectl run "${POD_NAME}" \
   --image="${IMAGE}" \
-  --requests='cpu=750m,memory=2Gi,ephemeral-storage=2Gi' \
```
medb, line 17: Does this make presubmit flaky?

cloudbuild/cloudbuild.yaml
```diff
@@ -75,8 +75,41 @@ steps:
       - 'CLOUDSDK_COMPUTE_REGION=us-central1'
       - 'CLOUDSDK_CONTAINER_CLUSTER=init-actions-presubmit'
 
+# Run presubmit tests in parallel for 2.1 Debian image
+  - name: 'gcr.io/cloud-builders/kubectl'
+    id: 'dataproc-2.0-debian10-tests'
```
medb, line 80: Change IDs to reflect tested images.

medb (changes requested):

cloudbuild/run-presubmit-on-k8s.sh
jayadeep-jayaraman: This is not being used since last few versions as GKE Autopilot has been ignoring this parameter. This line is printed `--requests has been deprecated, has no effect and will be removed in 1.24` in the last few runs

medb: Yeah, I think that we need to migrate it to the new API - we added this flag because w/o presubmits were flaky. I propose to move it the separate PR.

cloudbuild/cloudbuild.yaml
jayadeep-jayaraman: updated

medb (approved):

cloudbuild/run-presubmit-on-k8s.sh
medb: Reverted, please file a separate PR for this change.

PR #1050, opened 2023-02-17
Modified the script to be compatible with new K8S version

jayadeep-jayaraman:

PR #1052, opened 2023-02-20
removed obsoleted argument

cjac:
the --requests argument has been deprecated and has now become obsolete

### DIFF
--- a/cloudbuild/run-presubmit-on-k8s.sh
+++ b/cloudbuild/run-presubmit-on-k8s.sh
@@ -14,7 +14,6 @@ LOGS_SINCE_TIME=$(date --iso-8601=seconds)
 
 kubectl run "${POD_NAME}" \
   --image="${IMAGE}" \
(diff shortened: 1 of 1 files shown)


jayadeep-jayaraman:
The default cpu and memory quotas are lower than what was set in the request arg and it was to remove some flakiness from the tests as mentioned [here](https://github.com/GoogleCloudDataproc/initialization-actions/pull/1004). We should move towards a deployment spec model instead of directly running Pods

jayadeep-jayaraman:
Also, this [open PR](https://github.com/GoogleCloudDataproc/initialization-actions/pull/1050/files) is to move to the new K8S version

jayadeep-jayaraman (approved):

jayadeep-jayaraman:
Hi @cjac - I am approving this temporary fix to test the pending GPU changes. I will open a separate issue to fix the API's permanently. I will not be able to get to it for another 1 week due to other priorities and hence approving this temporary change

PR #1064, opened 2023-06-23
Fixes to enable connectors on 2.1 images

cjac:
* In 2.1 images, environment variable is DATAPROC_IMAGE_VERSION
* added functions to compare versions
* When image version is >= 2.1, use scala v2.13

### DIFF
--- a/connectors/connectors.sh
+++ b/connectors/connectors.sh
@@ -27,15 +27,6 @@ readonly BIGQUERY_CONNECTOR_URL=$(/usr/share/google/get_metadata_value attribute
 readonly SPARK_BIGQUERY_CONNECTOR_URL=$(/usr/share/google/get_metadata_value attributes/spark-bigquery-connector-url || true)
 readonly HIVE_BIGQUERY_CONNECTOR_URL=$(/usr/share/google/get_metadata_value attributes/hive-bigquery-connector-url || true)
 
-is_worker() {
-  local role
-  role="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
-  if [[ $role != Master ]]; then
-    return 0
-  fi
-  return 1
-}
-
 min_version() {
   echo -e "$1\n$2" | sort -r -t'.' -n -k1,1 -k2,2 -k3,3 | tail -n1
 }
@@ -63,7 +54,8 @@ function get_connector_url() {
 
   # spark-bigquery
   if [[ $name == spark-bigquery ]]; then
-    readonly -A LT_DP_VERSION_SCALA=([2.0]="2.11" # On Dataproc images < 2.0, use Scala 2.11
+    readonly -A LT_DP_VERSION_SCALA=([1.5]="2.11" # On Dataproc images < 1.5, use Scala 2.11
+                                     [2.0]="2.12" # On Dataproc images < 2.0, use Scala 2.12
                                      [2.1]="2.12" # On Dataproc images < 2.1, use Scala 2.12
                                      [2.2]="2.13" # On Dataproc images < 2.2, use Scala 2.13
                                     )
@@ -71,7 +63,7 @@ function get_connector_url() {
     # cluster.  We will use this to determine the appropriate connector to use
     # based on the scala version.
 
-    for LT_DP_VER in 2.0 2.1 2.2 ; do
+    for LT_DP_VER in 1.5 2.0 2.1 2.2 ; do
       if ! test -v scala_version && compare_versions_lt $DATAPROC_IMAGE_VERSION ${LT_DP_VER} ; then
         local -r scala_version=${LT_DP_VERSION_SCALA[${LT_DP_VER}]}
         continue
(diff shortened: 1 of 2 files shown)


cjac:
Warning  FailedScheduling

connectors/connectors.sh
```diff
@@ -72,7 +84,7 @@ get_connector_url() {
   fi
 
   # bigquery
-  if [[ $(min_version "$DATAPROC_VERSION" 2.0) == 2.0 ]]; then
+  if [[ $(min_version "$DATAPROC_IMAGE_VERSION" 2.0) == 2.0 ]]; then
```
functicons, line 87: Some older 1.5 and 2.0 images might not have DATAPROC_IMAGE_VERSION defined, instead they use DATAPROC_VERSION. So, for backward compatibility, you might need to check both.

connectors/connectors.sh
```diff
@@ -50,13 +58,17 @@ get_connector_url() {
...
     # We will use this to determine the appropriate connector to use
     # based on the scala version.
-    if [[ $(min_version "$DATAPROC_VERSION" 1.5) == 1.5 ]]; then
-      local -r scala_version=2.12
-    else
+    if compare_versions_lt $DATAPROC_IMAGE_VERSION 2.0 ; then
       local -r scala_version=2.11
+    else if compare_versions_lt $DATAPROC_IMAGE_VERSION 2.1 ; then
+           local -r scala_version=2.12
+         else if compare_versions_lt $DATAPROC_IMAGE_VERSION 2.2 ; then
```
functicons, line 68: In the future, when a new Dataproc version is released, `scala_version` will be undefined. So, consider fail explicitly for unknown Dataproc version.

functicons (changes requested):

connectors/connectors.sh
cjac: done

connectors/connectors.sh
cjac: done

connectors/connectors.sh
```diff
@@ -50,13 +54,32 @@ get_connector_url() {
...
     # based on the scala version.
-    if [[ $(min_version "$DATAPROC_VERSION" 1.5) == 1.5 ]]; then
-      local -r scala_version=2.12
-    else
-      local -r scala_version=2.11
+
+    for LT_DP_VER in 1.5 2.0 2.1 2.2 ; do
+      if ! test -v scala_version && compare_versions_lt $DATAPROC_IMAGE_VERSION ${LT_DP_VER} ; then
+        local -r scala_version=${LT_DP_VERSION_SCALA[${LT_DP_VER}]}
+        continue
```
functicons, line 69: Why continue instead of break?

connectors/connectors.sh
```diff
@@ -50,13 +54,32 @@ get_connector_url() {
...
+      fi
+    done
+
+    if compare_versions_lt $DATAPROC_IMAGE_VERSION 2.0 ; then
+      if ! compare_versions_lt ${version} 0.30.0 ; then
+        echo "spark-bigquery-with-dependencies version ${version} does not target scala versions below 2.12 (${scala_version})" >&2
+        exit -1
```
functicons, line 76: We usually exit with 1 for error.

connectors/connectors.sh
```diff
@@ -50,13 +54,32 @@ get_connector_url() {
...
+      fi
+    fi
+
+    if ! test -v scala_version ; then
```
functicons, line 80: Instead of testing scala version to detect unsupported image version, prefer check image version directly at the beginning.

connectors/connectors.sh
```diff
@@ -50,13 +54,32 @@ get_connector_url() {
...
+      echo "unsupported DATAPROC_IMAGE_VERSION" >&2
```
functicons, line 81: Include the actual value in the message.

functicons (changes requested):

connectors/connectors.sh
```diff
@@ -50,13 +54,32 @@ get_connector_url() {
...
+    # DATAPROC_IMAGE_VERSION is built from an environment variable set on the
+    # cluster.  We will use this to determine the appropriate connector to use
     # based on the scala version.
-    if [[ $(min_version "$DATAPROC_VERSION" 1.5) == 1.5 ]]; then
-      local -r scala_version=2.12
-    else
-      local -r scala_version=2.11
+
+    for LT_DP_VER in 1.5 2.0 2.1 2.2 ; do
+      if ! test -v scala_version && compare_versions_lt $DATAPROC_IMAGE_VERSION ${LT_DP_VER} ; then
```
functicons, line 67: nit: prefer `"${var}"` over `$var`. This is the convention of Dataproc shell script.

functicons (changes requested):

connectors/connectors.sh
cjac: my bad!  I meant last;

connectors/connectors.sh
cjac: I'll take care of it

cjac:
Thank you for the thorough review, Dagang!  Let's see if we can build from this one.

/gcbrun

functicons (approved):
LGTM, thanks!

PR #1072, opened 2023-07-26
Added OpenTelemetry Initialization action for Dataproc clusters

shambhavirai16:
- Added [OpenTelemetry](https://opentelemetry.io/) initialization script for installing and configuring [otelcol-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) agent for monitoring purposes. This enables users to collect and export metrics/logs/traces and supports numerous sinks.
- Support for GCM exporter (enabled by default)
- Support for scraping prometheus metrics from any arbitrary http endpoints passed in `prometheus-scrape-endpoints` metadata property
- Added README for the new init script

### DIFF
--- a/otel/README.md
+++ b/otel/README.md
@@ -36,3 +36,7 @@ OTel installed on every node:
     ```
 NOTE: Add a string of space separated prometheus scrape urls in `PROMETHEUS_ENDPOINTS`, example value: `0.0.0.0:1234 0.0.0.0:1235`
 
+---
+
+### Usage warning
+This initialisation action is provided without support and currently does not have an integration test. Read more [here](https://github.com/GoogleCloudDataproc/initialization-actions/#why-these-samples-are-provided).


otel/otel.sh
```diff
@@ -0,0 +1,79 @@
...
+    metrics:
+      receivers: [prometheus]
+      processors: [resourcedetection]
+      exporters: [googlecloud]
+EOF
+}
+
+function start_services(){
+  systemctl daemon-reload
+  systemctl restart otelcol-contrib
```
akshatbhargava123, line 69: Would this be `start` or `restart`?

akshatbhargava123: if the service is not already running, what is it's behaviour then?

otel/otel.sh
```diff
@@ -0,0 +1,79 @@
...
+
+  cat >/etc/otelcol-contrib/config.yaml<<EOF
+receivers:
+  prometheus:
+    config:
+      scrape_configs:
+      - job_name: 'otel-collector'
+        scrape_interval: 10s
+        static_configs:
+        - targets: [${prometheus_scrape_endpoints}]
```
akshatbhargava123, line 43: Can we confirm there are no logs when no scrape endpoints are passed? Else it would be too noisy in case users are not using prometheus receiver and still getting logs regarding empty scrape targets.

otel/otel.sh
```diff
@@ -0,0 +1,79 @@
+# This initialization action installs OpenTelemetry Collector Contrib (otelcol-contrib) in a Dataproc cluster
+# configures OTel and pulls metrics from Prometheus endpoints provided by the customer. 
+# The metrics are then exported to Google Cloud Monitoring (GCM).
+
+set -euxo pipefail
+
+function install_otel() {
+  # Install otelcol-contrib as package (https://github.com/open-telemetry/opentelemetry-collector-contrib)
+  wget https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.81.0/otelcol-contrib_0.81.0_linux_amd64.deb
+  dpkg -i otelcol-contrib_0.81.0_linux_amd64.deb
```
akshatbhargava123, line 10: You can pull the OTel version (`0.81.0`) as a readonly variable on top of this file.

otel/otel.sh
```diff
@@ -0,0 +1,79 @@
+# This initialization action installs OpenTelemetry Collector Contrib (otelcol-contrib) in a Dataproc cluster
+# configures OTel and pulls metrics from Prometheus endpoints provided by the customer. 
+# The metrics are then exported to Google Cloud Monitoring (GCM).
+
+set -euxo pipefail
```
akshatbhargava123, line 5: Can't seem to find pipes being used anywhere in this file, can you confirm if this is needed?

otel/otel.sh
```diff
@@ -0,0 +1,79 @@
+# This initialization action installs OpenTelemetry Collector Contrib (otelcol-contrib) in a Dataproc cluster
```
akshatbhargava123, line 1: We need to add `#!/usr/bin/env bash` on top of this file.

otel/otel.sh
```diff
@@ -0,0 +1,79 @@
...
+set -euxo pipefail
+
+function install_otel() {
+  # Install otelcol-contrib as package (https://github.com/open-telemetry/opentelemetry-collector-contrib)
+  wget https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.81.0/otelcol-contrib_0.81.0_linux_amd64.deb
+  dpkg -i otelcol-contrib_0.81.0_linux_amd64.deb
+}
+
+function configure_endpoints() {
+  readonly PROMETHEUS_ENDPOINTS="$(/usr/share/google/get_metadata_value attributes/prometheus-scrape-endpoints || '')"
```
akshatbhargava123, line 14: All readonly variables (hardcoded or reading metadata values) can be moved to top area before functions are defined to keep consistency with other init actions.

akshatbhargava123 (changes requested):

otel/README.md
```diff
@@ -0,0 +1,31 @@
...
+    REGION=<region>
+    CLUSTER_NAME=<cluster_name>
+    PROMETHEUS_ENDPOINTS=<prometheus_endpoints>
+    gcloud dataproc clusters create ${CLUSTER_NAME} \
+        --region ${REGION} \
+        --metadata=prometheus-scrape-endpoints="${PROMETHEUS_ENDPOINTS}"
+        --initialization-actions gs://goog-dataproc-initialization-actions-${REGION}/otel/otel.sh
+    ```
+
+NOTE: Add a string of space separated prometheus scrape urls in `PROMETHEUS_ENDPOINTS`, example value: `0.0.0.0:1234 0.0.0.0:1235`
```
Arun-Krishna, line 30: Can we make this comma separated instead? quotes are not mandatory for passing the value, and in their absence this is just a silly mistake that can happen

otel/otel.sh
shambhavirai16: If no endpoint is passed in, the system will not perform any operations and no errors will be logged while if an endpoint is provided but there are no metrics available, it will repeatedly generate error messages.

otel/otel.sh
shambhavirai16: Added.

otel/otel.sh
shambhavirai16: Pipes are not used, removed this.

otel/otel.sh
shambhavirai16: Added.

otel/otel.sh
shambhavirai16: Done.

otel/otel.sh
```diff
@@ -0,0 +1,79 @@
...
+
+processors:
+  resourcedetection:
+    detectors: [gcp]
+    timeout: 10s
+
+exporters:
+  googlecloud:
+    log:
+      default_log_name: opentelemetry.io/collector-exported-log
```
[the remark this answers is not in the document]
shambhavirai16: Removed the log to prevent further increase in cost to customers.

otel/otel.sh
shambhavirai16: Updated.

otel/README.md
shambhavirai16: Commas cannot be added to the value of any metadata because the system interprets them as delimiters between two metadata properties.

akshatbhargava123 (approved):

PR #1073, opened 2023-07-31
Hue: install libmysqlclient20 for 2.0

functicons:


hue/hue.sh
```diff
@@ -53,8 +54,20 @@ function update_repo() {
   if command -v apt-get >/dev/null; then
     retry_command "apt-get update"
   else
-    retry_command "yum -y update"
```
yigress, line 56: if '-y' causing issue, can we add comment why not use it.

functicons: Good catch. I should not have removed the `-y`.

yigress (approved):

Issue #1077, opened 2023-08-14
[oozie] intermittent error writing to HDFS during init action


cjac:
Some users may experience script failure when clusters start the oozie init action script prior to HDFS being fully online.

```
+ hadoop fs -put -f /tmp/oozie-install-m95M/share /user/oozie/
2023-08-07 20:41:35,938 WARN hdfs.DataStreamer: DataStreamer Exception
org.apache.hadoop.ipc.RemoteException(java.io.IOException): File /user/oozie/share/lib/pig/hadoop-yarn-client-3.3.3.jar._COPYING_ could only be written to 0 of the 1 minReplication nodes. There are 0 datanode(s) running and 0 node(s) are excluded in this operation.
        at org.apache.hadoop.hdfs.server.blockmanagement.BlockManager.chooseTarget4NewBlock(BlockManager.java:2315)
```

cjac:
fixed in #1089 

May need to be paired with `--metadata startup-script-url="${INIT_ACTIONS_ROOT}/delay-masters-startup.sh"`

```
#!/bin/bash

set -x

readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
if [[ "${ROLE}" != 'Master' ]]; then set +x; exit 0; fi

node_number=$(echo ${HOSTNAME} | perl -ne '/-m-(\d+)/; print $1')
delay_seconds=$((${node_number} * 60))
sleep ${delay_seconds}s

NOW=$(date +"%F-%T")
echo "instance #${node_number} (${HOSTNAME}) proceeds at ${NOW}" | tee /var/log/delay-masters.log

set +x

```

cjac:
This issues appears to be resolved

PR #1078, opened 2023-08-14
[oozie] Create and call function to await HDFS Data Nodes

cjac:
Some users are experiencing errors when init actions begin running before dfs has come online.  This patch pauses before the first call to dfs in order to avoid this race condition.

This patch is intended as the fix for #1077

### DIFF
--- a/hue/hue.sh
+++ b/hue/hue.sh
@@ -56,7 +56,7 @@ function update_repo() {
   if command -v apt-get >/dev/null; then
     retry_command "apt-get update"
   else
-    retry_command "yum -y update"
+    retry_command "yum check-update"
   fi
 }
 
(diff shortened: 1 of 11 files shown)


cjac:
The pause needs to happen before the initialization actions are launched.  We can stagger the cluster node startup with a script such as the following.

```
cjac@cjac11:~/src/team/cloud-support-tses/products/dataproc/dataproc-repro$ cat init/delay-masters-startup.sh                    #!/bin/bash

set -x

readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
if [[ "${ROLE}" != 'Master' ]]; then set +x; exit 0; fi

node_number=$(echo ${HOSTNAME} | perl -ne '/-m-(\d+)/; print $1')
delay_seconds=$((${node_number} * 150))
sleep ${delay_seconds}s

NOW=$(date +"%F-%T")
echo "instance #${node_number} (${HOSTNAME}) proceeds at ${NOW}" | tee /var/log/delay-masters.log

set +x
```

Once the agent has launched and kicks off the init actions, the patched script should be capable of running even if the HDFS store does not come online ever.

cjac:
I'm going to close this CL and open a new one.  Somehow we got one of nvliyuan's ephemeral email addresses as a contributor.  This was likely due to me rebasing my changes on a different branch.

cjac:
continued in #1089

PR #1086, opened 2023-09-05
[serial] new initialization action to execute in serial

cjac:
If you are otherwise unable to execute initialization actions in a set order, you may use this script to execute them in the order specified.


cjac:
This is not necessary.  Terraform provides a way to specify init actions sequentially.

PR #1089, opened 2023-09-19
[oozie] improve support for 2.1 images

cjac:
This PR includes updates that we've found improve the experience for some of our customers.

* declare master_node and ROLE variables in header
* define sample secret name to provide more complete defaults
* create function await_hdfs_datanodes to pause for quorum
* only interact with dfs if quorum has been reached
* moved oozie.service.ProxyUserService.proxyuser.oozie.* property definitions down a few lines
* called out whitepaper recommending these properties
* purge local guava jar and replace it with octets from an https url


cjac:
continuation of #1078

kuldeepkk-dev (approved):

PR #1094, opened 2023-10-03
[general] add a table of supported versions

cjac:
Making note in README.md that no versions of these initialization actions are certified to be in working order.

Let's fill in this table with known good image versions and the initialization actions which work on them.

I'm inviting general participation for this commit, please.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,56 @@ You can decide when to sync your copy of the initialization action with any chan
 
 ## Why these samples are provided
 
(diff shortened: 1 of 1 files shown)


cjac:
@nvliyuan - can you tell me the latest dataproc image versions your team has checked the rapids/rapids.sh and/or gpu/install_gpu_driver.sh on?  Should we also include that recent run of the dask tests?

nvliyuan:
> @nvliyuan - can you tell me the latest dataproc image versions your team has checked the rapids/rapids.sh and/or gpu/install_gpu_driver.sh on? Should we also include that recent run of the dask tests?

Currently we only test spark-rapids/* on dataproc image2.0* and image2.1*, we don't test dask/*.

cjac:
Thank you!  I'll include that information in the patch this week.

PR #1098, opened 2023-10-16
[livy] Livy 202301016

cjac:
this change should have gone into the previous livy change.


cjac:
@functicons would you mind helping with this one, too?  The change should have been applied to both 2.0 and 2.1 and the previous change only applied it to 2.1

medb (approved):

PR #1101, opened 2023-10-17
[zeppelin] corrected DATAPROC_VERSION to DATAPROC_IMAGE_VERSION

erickMunz:
This will not make zeppelin work on 2.1 image, instead users should use  --optional-components=ZEPPELIN

### DIFF
--- a/zeppelin/zeppelin.sh
+++ b/zeppelin/zeppelin.sh
@@ -26,7 +26,7 @@ fi
 
 
 readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_IMAGE_VERSION}.
-Use Zeppelin Component instead: https://cloud.google.com/dataproc/docs/concepts/components/zeppelin"
+Use Zeppelin Optional Component instead: https://cloud.google.com/dataproc/docs/concepts/components/zeppelin"
 [[ "${DATAPROC_IMAGE_VERSION}" != 1.* ]] && echo "${NOT_SUPPORTED_MESSAGE}" && exit 1
 
 readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
--- a/zeppelin/zeppelin.sh
+++ b/zeppelin/zeppelin.sh
@@ -19,9 +19,15 @@
 
 set -euxo pipefail
 
-readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_VERSION}.
+# Detect dataproc image version from its various names
+if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_IMAGE_VERSION; then
+  DATAPROC_IMAGE_VERSION="${DATAPROC_IMAGE_VERSION}"
+fi
+
+
+readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_IMAGE_VERSION}.
 Use Zeppelin Component instead: https://cloud.google.com/dataproc/docs/concepts/components/zeppelin"
-[[ $DATAPROC_VERSION != 1.* ]] && echo "$NOT_SUPPORTED_MESSAGE" && exit 1
+[[ $DATAPROC_IMAGE_VERSION != 1.* ]] && echo "$NOT_SUPPORTED_MESSAGE" && exit 1
 
 readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
 readonly INTERPRETER_FILE='/etc/zeppelin/conf/interpreter.json'
--- a/zeppelin/zeppelin.sh
+++ b/zeppelin/zeppelin.sh
@@ -20,14 +20,14 @@
 set -euxo pipefail
 
 # Detect dataproc image version from its various names
-if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_IMAGE_VERSION; then
+if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_VERSION; then
   DATAPROC_IMAGE_VERSION="${DATAPROC_IMAGE_VERSION}"
 fi
 
 
 readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_IMAGE_VERSION}.
 Use Zeppelin Component instead: https://cloud.google.com/dataproc/docs/concepts/components/zeppelin"
-[[ $DATAPROC_IMAGE_VERSION != 1.* ]] && echo "$NOT_SUPPORTED_MESSAGE" && exit 1
+[[ "${DATAPROC_IMAGE_VERSION}" != 1.* ]] && echo "${NOT_SUPPORTED_MESSAGE}" && exit 1
 
 readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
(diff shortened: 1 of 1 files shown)


zeppelin/zeppelin.sh
```diff
@@ -19,9 +19,15 @@
 
 set -euxo pipefail
 
-readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_VERSION}.
+# Detect dataproc image version from its various names
+if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_IMAGE_VERSION; then
```
functicons, line 23: This doesn't look correct. I guess you might want:

```
if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_VERSION; then
  DATAPROC_IMAGE_VERSION="${DATAPROC_VERSION}"
fi
```

zeppelin/zeppelin.sh
```diff
@@ -19,9 +19,15 @@
...
+  DATAPROC_IMAGE_VERSION="${DATAPROC_IMAGE_VERSION}"
+fi
+
+
+readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_IMAGE_VERSION}.
 Use Zeppelin Component instead: https://cloud.google.com/dataproc/docs/concepts/components/zeppelin"
-[[ $DATAPROC_VERSION != 1.* ]] && echo "$NOT_SUPPORTED_MESSAGE" && exit 1
+[[ $DATAPROC_IMAGE_VERSION != 1.* ]] && echo "$NOT_SUPPORTED_MESSAGE" && exit 1
```
functicons, line 30: Prefer this form: "${var}". Both the "" and {} are needed.

functicons (changes requested):

zeppelin/zeppelin.sh
cjac: You're right!  We were too aggressive about the search and replace.  Thanks for catching this!

zeppelin/zeppelin.sh
cjac: Sure.  We'll make this update as well once he returns to his desk.

erickMunz: @functicons made the changes, how does it look?

zeppelin/zeppelin.sh
```diff
@@ -19,9 +19,15 @@
...
 set -euxo pipefail
 
-readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_VERSION}.
+# Detect dataproc image version from its various names
+if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_VERSION; then
+  DATAPROC_IMAGE_VERSION="${DATAPROC_IMAGE_VERSION}"
+fi
+
+
+readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_IMAGE_VERSION}.
```
cjac, line 28: looks like we missed the trailing " here Erick!

cjac (changes requested):
missing trailing "

zeppelin/zeppelin.sh
cjac: nope.  I was wrong.  You had it correct.

Ready for your review, @functicons

cjac (approved):
changes were correct initially.

zeppelin/zeppelin.sh
```diff
@@ -20,14 +20,14 @@
 set -euxo pipefail
 
 # Detect dataproc image version from its various names
-if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_IMAGE_VERSION; then
+if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_VERSION; then
   DATAPROC_IMAGE_VERSION="${DATAPROC_IMAGE_VERSION}"
```
functicons, line 24: This should be `DATAPROC_IMAGE_VERSION="${DATAPROC_VERSION}"`

zeppelin/zeppelin.sh
functicons, line 28: Missing the trailing ", and you might want to it actionable by adding "use Zeppelin optional component please".

zeppelin/zeppelin.sh
```diff
@@ -66,8 +72,8 @@ function install_zeppelin() {
   local asm_3_1=/usr/lib/zeppelin/lib/asm-3.1.jar
   local asm_5_0_4=/usr/lib/zeppelin/lib/asm-5.0.4.jar
 
-  if [[ "$zeppelin_version" == "0.7."* && -f "$asm_3_1" && -f "$asm_5_0_4" ]]; then
-    rm "$asm_5_0_4"
+  if [[ "${zeppelin_version}" == "0.7."* && -f "${asm_3_1}" && -f "${asm_5_0_4}" ]]; then
```
functicons, line 75: Add a comment to explain why it is needed.

functicons (changes requested):

zeppelin/zeppelin.sh
cjac: you are correct once again

zeppelin/zeppelin.sh
cjac: no, the string is continued to the next line.  I thought the same thing when I looked at it, though, so you're in good company.  Erick corrected me.  The full text reads:

readonly NOT_SUPPORTED_MESSAGE="Zeppelin initialization action is not supported on Dataproc ${DATAPROC_IMAGE_VERSION}.
Use Zeppelin Component instead: https://cloud.google.com/dataproc/docs/concepts/components/zeppelin"

Would you like it corrected from `Use Zeppelin Component instead` to `Use Zeppelin Optional Component instead` ?

zeppelin/zeppelin.sh
cjac: I don't think we know why it is needed.  We just replaced $... with ${...} as you requested above

cjac: @erickMunz - do you have context here?  I don't know Zeppelin well enough to say anything useful about this block of code.

zeppelin/zeppelin.sh
cjac: @erickMunz - we need to change the rvalue on line 24 here

zeppelin/zeppelin.sh
erickMunz: Added Use Zeppelin Optional Component instead

zeppelin/zeppelin.sh
erickMunz: @cjac  not something I have worked on, but I believe this involves some version compatibility change to use libraries, but we did not changed that

cjac: I think we should leave it without more than the superficial variable syntax changes, @functicons

zeppelin/zeppelin.sh
cjac: This is the last unaddressed change I think Erick

cjac (approved):
Assuming the gcbrun succeeds, I think this is GTG

cjac:
Erick, have you manually tested this init script on 2.1 and 1.5?  I'm having a hard time getting the tests to run.

cjac:
oh, it looks like the BUILD file in the zeppelin/ directory was removed.

functicons (approved):
LGTM

PR #1111, opened 2023-11-14
Add Spark BQ DSv2 connectors for Dataproc 2.2

vishalkarve15:


connectors/connectors.sh
```diff
@@ -7,6 +7,7 @@ set -euxo pipefail
```
medb, line 7: Should we deprecate this init action, instead of supporting it, as we natively support connectors upgrade in the DPGCE images now?

medb (changes requested):

connectors/connectors.sh
vishalkarve15: The property names used here are still slightly different. E.g. Here we use spark-bigquery-connector-version, whereas in native support we use SPARK_BQ_CONNECTOR_VERSION. 
Sure, we can move these properties into native and deprecate this one.

It will still require us to make changes here so that it's at least no-op for 2.2 to avoid breakage (installation directory is different for 2.2).
I just felt this is easier.

vishalkarve15:
Handling this in the native code.

PR #1112, opened 2023-11-28
Support debian12 for rstudio installation

vinayakumarb:
Debian12 (codename bookworm) is not supported in the current version of rstudio init script.

`rstudio-server-2023.09.1-494-amd64.deb` works same as ubuntu22 on debian12 as well.


functicons (approved):
LGTM

PR #1113, opened 2023-11-29
modifying the tests for both mysql and postgresql instances on hive m…

prince-cs:
The test written is modified to check for both databases present MySQL and PostgreSQL. Earlier, the test used to create the cluster and instance on default settings with the arguments being passed. Now, with the new changes it creates the instances for both MySQL and PostgreSQL which has hive metastore DB.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -61,7 +61,7 @@ py_test(
     name = "test_cloud_sql_proxy",
     size = "enormous",
     srcs = ["cloud-sql-proxy/test_cloud_sql_proxy.py"],
-    data = ["cloud-sql-proxy/cloud-sql-proxy.sh"],
+    data = ["cloud-sql-proxy/cloud-sql-proxy.sh", "cloud-sql-proxy/hivetest.hive"],
     local = True,
     shard_count = 3,
     deps = [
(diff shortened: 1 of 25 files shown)


yigress:
I am a little confused about the tests. It runs cloud-sql-proxy init script on the same cluster with both mysql and postgres DB and set up them for the same hive metastore service? how does that work?

prince-cs:
It was asked to run it for both the database types MySQL and PostgreSQL.

https://screenshot.googleplex.com/8b8NdGQdjsVg4Af

yigress:
we want to keep the testing content the same for different metastore DBs, can we use same test script file TEST_SCRIPT_FILE_NAME, but passing different variables for different metastore DB

cloud-sql-proxy/pyspark_metastore_test.py
```diff
@@ -24,25 +34,54 @@
...
+  print('Successfully found table {} in Cloud SQL Hive metastore'.format(
+    test_table.TBL_NAME))
+
+  # Clean up table.
+  sqlContext.sql('DROP TABLE ' + test_table_name)
+
+if args.instance_type == INSTANCE_MYSQL:
+  configure_table_mysql()
+elif args.instance_type == INSTANCE_POSTGRESQL:
+  configure_table_postgresql()
```
yigress, line 87: the test logic is the same, can we deduplicated the code here
something like extract the jdbc connection string out, something like 
jdbcUri = if mysql jdbc:mysql://.... else jdbc:postgres...
then test with the jdbcUri?

PR #1116, opened 2023-12-01
Fix install_gpu_driver.sh failures in rocky 2.0 and 2.1 images

bcheena:
1. install_gpu_driver.sh init script fails with the following error in rocky 2.0 and 2.1 images:
```
++ dnf -y -q update
Error: 
 Problem: The operation would result in removing the following protected packages: systemd
```

We should exclude systemd from dnf update.

2. In 2.0 rocky, the available version of kernel-devel is `4.18.0-513.9.1.el8_9.x86_64` which does not match the running kernel version `4.18.0-477.27.1.el8_8.x86_64`.

```
++ dnf -y -q install kernel-devel-4.18.0-477.27.1.el8_8.x86_64
Error: Unable to find a match: kernel-devel-4.18.0-477.27.1.el8_8.x86_64
```
There should be a condition to check if the kernel needs to be upgraded. 

`upgrade_kernel` method has been taken from https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/spark-rapids/spark-rapids.sh#L474.

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -71,7 +71,7 @@ readonly DRIVER=${NVIDIA_DEBIAN_GPU_DRIVER_VERSION_PREFIX}
 # As of Rocky 8.7, kernel 4.18.0-425 is unable to build older nvidia kernel drivers
 ROCKY_BINARY_INSTALL="false"
 if [[ "${OS_NAME}" == "rocky" &&  "${DRIVER}" < "510" ]]; then
-  readonly ROCKY_BINARY_INSTALL="true"
+  ROCKY_BINARY_INSTALL="true"
 fi
 readonly ROCKY_BINARY_INSTALL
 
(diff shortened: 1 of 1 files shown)


gpu/install_gpu_driver.sh
```diff
@@ -486,6 +488,75 @@ EOF
...
+    CURRENT_KERNEL_VERSION="${KERN_VER}-${KERN_REL}"
+  else
+    echo "unsupported OS: ${OS_NAME}!"
+    exit -1
+  fi
+
+  # Get latest version available in repos
+  if [[ "${OS_NAME}" == "debian" ]]; then
+    apt-get -qq update
+    TARGET_VERSION=$(apt-cache show --no-all-versions linux-image-amd64 | awk '/^Version/ {print $2}')
```
cjac, line 510: perhaps allow the user to specify a target version with metadata value?

cjac: Have you taken a look at this?  It does look like some of your code mirrors the work I did in this incomplete patch...

https://github.com/cjac/initialization-actions/blob/change-kernel-version-202211/kernel/

cjac (commented):
This feels unsafe.  I stopped working on this a while ago because it felt flimsy...

cjac:
`upgrade_kernel method has been taken from https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/spark-rapids/spark-rapids.sh#L474.`

oh gosh.  I forgot that we put that into production.  So sketchy...

cjac:
Hey there Cheena,

I've been taking with Gregory from Rocky.  I think I should set up a call
with them, Nvidia, and some representatives from the Dataproc team to
discuss the problem.

I hope this helps me to remember to set it up!

C.J.

cjac:
This is looking good.  I'll test it next.

cjac:
These changes have been tested with
* 2.0-rocky8
* 2.1-rocky8
* 2.2-rocky9
* 2.0-debian10

cjac (approved):
LGTM

PR #1124, opened 2024-01-03
Init action for installing ops agent

vamshikrishna-g:
With Dataproc 2.2, we recommend installing ops agent for VM system metrics as well as third party application observability. See release note: https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-release-2.2

This PR provides an init action for installing Ops Agent.


opsagent/opsagent.sh
```diff
@@ -0,0 +1,34 @@
...
+
+cat <<EOF >> /etc/google-cloud-ops-agent/config.yaml
+logging:
+  service:
+    pipelines:
+      default_pipeline:
+        receivers: []
+EOF
+
+service google-cloud-ops-agent restart
```
igorpeshansky, line 34: It's preferred to use `systemctl restart google-cloud-ops-agent` (https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/installation#restart).

opsagent/README.md
```diff
@@ -0,0 +1,27 @@
+# Ops Agent
+
+With [Dataproc 2.2 image version](https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-release-2.2), we recommend installing [Google Cloud Ops Agent](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent) to obtain system metrics. 
+
+This initialization action will install Ops Agent on a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster and provide similar metrics as the [--metric-sources=monitoring-agent-defaults setting](https://cloud.google.com/dataproc/docs/guides/dataproc-metrics#monitoring_agent_metrics) which was supported until Dataproc 2.1.
+[This page](https://cloud.google.com/monitoring/api/metrics_agent#oagent-vs-magent) highlights differences in metric collection between Ops Agent and legacy monitoring agent.
+
+This initialization action also specifies a user configuration in order to skip syslogs collection from your cluster nodes. If this is not specified, Ops Agent will collect syslogs besides the system (node) metrics. You can further customize this configuration to collect logs and metrics from other third-party applications.
```
igorpeshansky, line 8: Just curious, why are we disabling syslog collection? Would it make sense to provide two init actions for customers — one that simply installs the Ops Agent with the default config, and one that installs it without syslog collection?

Better yet, can't you specify multiple initialization actions? Why don't we provide one that disables syslog collection after installing the Ops Agent, so it can be specified as:
```bash
gcloud dataproc clusters create ${CLUSTER_NAME} \
    --image-version=2.2 \
    --region=${REGION} \
    --initialization-actions=gs://goog-dataproc-initialization-actions-${REGION}/opsagent/opsagent.sh,gs://goog-dataproc-initialization-actions-${REGION}/opsagent/disable_syslog.sh
```
?

opsagent/README.md
```diff
@@ -0,0 +1,27 @@
...
+- Use the `gcloud` command to create a new cluster with this initialization
+    action.
+
+    ```bash
+    REGION=<region>
+    CLUSTER_NAME=<cluster_name>
+    gcloud dataproc clusters create ${CLUSTER_NAME} \
+        --image-version 2.2 \
+        --region ${REGION} \
+        --initialization-actions gs://goog-dataproc-initialization-actions-${REGION}/opsagent/opsagent.sh
```
igorpeshansky, line 25: Minor nit: the [documentation](https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/init-actions#gcloud-command) uses `=` in gcloud flags, so:
```suggestion
    gcloud dataproc clusters create ${CLUSTER_NAME} \
        --image-version=2.2 \
        --region=${REGION} \
        --initialization-actions=gs://goog-dataproc-initialization-actions-${REGION}/opsagent/opsagent.sh
```

igorpeshansky (commented):
Some initial thoughts.

opsagent/README.md
```diff
@@ -0,0 +1,27 @@
+# Ops Agent
+
+With [Dataproc 2.2 image version](https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-release-2.2), we recommend installing [Google Cloud Ops Agent](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent) to obtain system metrics. 
+
+This initialization action will install Ops Agent on a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster and provide similar metrics as the [--metric-sources=monitoring-agent-defaults setting](https://cloud.google.com/dataproc/docs/guides/dataproc-metrics#monitoring_agent_metrics) which was supported until Dataproc 2.1.
```
igorpeshansky, line 5: Forgot one.

Formatting:
```suggestion
This initialization action will install Ops Agent on a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster and provide similar metrics as the [`--metric-sources=monitoring-agent-defaults` setting](https://cloud.google.com/dataproc/docs/guides/dataproc-metrics#monitoring_agent_metrics) which was supported until Dataproc 2.1.
```

vamshikrishna-g: Done

opsagent/README.md
vamshikrishna-g: Done

opsagent/opsagent.sh
vamshikrishna-g: Done

opsagent/README.md
vamshikrishna-g: Dataproc versions 2.1 and earlier did not collect syslogs by default.
Hence, thought to provide an example user configuration for achieving backwards compatibility assuming customers will be interested in keeping the same behaviour on clusters running 2.2 image version.

vinayakumarb: +1 for keeping the same behavior as before. 
syslog might add cloud logging cost for the customer, if they don't need it.

I like idea of providing multiple init actions.
Instead of having init action to disable syslog, we can provide one more init action to enable it.
This can be done as a follow up PR.

vinayakumarb (approved):
LGTM

opsagent/README.md
igorpeshansky: @vinayakumarb Providing the action to enable syslog ingestion would effectively undo the last bit of the currently defined action, which doesn't compose well, and would be confusing, IMO.

I'd rather `opsagent.sh` provided the [documented default behavior](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#default), which would match any other GCE environment. We can then either have a separate standalone action that installs the Ops Agent without syslog ingestion (e.g., `opsagent_nosyslog.sh`, or even `opsagent_21compat.sh`) or an add-on action specifically to disable syslog (e.g., `disable_syslog.sh`, as proposed above).

@vamshikrishna-g Regardless of the above approach, we can document the equivalent settings to the previous behavior to be the action (or a sequence of actions) that installs the Ops Agent with syslog collection disabled. The reason I suggested separating the two is so customers can use the provided (and supported) init action to install the Ops Agent with the documented default behavior.

If we choose to only provide one action in this PR, I strongly recommend renaming that action to `opsagent_nosyslog.sh` to avoid confusing customers.

vamshikrishna-g: Added init action `opsagent.sh` that provides default behavior and renamed the one providing user configuration override as `opsagent_nosyslog.sh`

opsagent/opsagent.sh
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+# This script installs the Google Cloud Ops Agent on each node in the cluster.
+# See https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#default
+# for built-in configuration of Ops Agent.
+
+# Detect dataproc image version from its various names
+if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_VERSION; then
+  DATAPROC_IMAGE_VERSION="${DATAPROC_VERSION}"
+fi
```
igorpeshansky, line 22: [Optional] Not going to block the PR on this, but this is just `DATAPROC_IMAGE_VERSION="${DATAPROC_IMAGE_VERSION:-${DATAPROC_VERSION}}"` (see https://tldp.org/LDP/abs/html/parameter-substitution.html and https://tldp.org/LDP/abs/html/refcards.html#AEN22728). Also in the other script.

opsagent/opsagent_nosyslog.sh
```diff
@@ -0,0 +1,43 @@
...
+curl -sSO https://dl.google.com/cloudagents/add-google-cloud-ops-agent-repo.sh
+bash add-google-cloud-ops-agent-repo.sh --also-install
+
+cat <<EOF >> /etc/google-cloud-ops-agent/config.yaml
+logging:
+  service:
+    pipelines:
+      default_pipeline:
+        receivers: []
+EOF
```
igorpeshansky, line 41: [Really optional] Also not going to block the PR, but if you swapped these, you wouldn't need to restart the agent afterwards, as the installed package would pick up the new config… You probably wouldn't save any script lines, though, because you'd need to ensure the directory exists (`mkdir -p /etc/google-cloud-ops-agent`) anyway…

opsagent/README.md
```diff
@@ -0,0 +1,40 @@
+# Ops Agent
+
+With [Dataproc 2.2 image version](https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-release-2.2), we recommend installing [Google Cloud Ops Agent](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent) to obtain system metrics. 
+
+This initialization action will install Ops Agent on a [Google Cloud Dataproc](https://cloud.google.com/dataproc) cluster and provide similar metrics as the [`--metric-sources=monitoring-agent-defaults`](https://cloud.google.com/dataproc/docs/guides/dataproc-metrics#monitoring_agent_metrics) setting which was supported until Dataproc 2.1.
```
igorpeshansky, line 5: Minor nit: whenever used as a subject in a sentence, our docs refer to the agent as "the Ops Agent"… Up to you if you want to be consistent.

opsagent/README.md
```diff
@@ -0,0 +1,40 @@
...
+[This page](https://cloud.google.com/monitoring/api/metrics_agent#oagent-vs-magent) highlights differences in metric collection between Ops Agent and legacy monitoring agent.
+
+We provide two variants of this initialization action:
+- `opsagent.sh` installs Ops Agent. [By default](https://cloud.google.com/stackdriver/docs/solutions/agents/ops-agent/configuration#default), it collects syslogs and system (node) metrics.
+- `opsagent_nosyslog.sh` installs Ops Agent and also specifies a user configuration in order to skip syslogs collection from your cluster nodes. If the user configuration is not specified, Ops Agent will collect syslogs besides the system (node) metrics. You can further customize this configuration to collect logs and metrics from other third-party applications.
+
+If you are looking for backward compatibility w.r.t. Dataproc image versions (2.1 and prior) that did not ingest syslogs from Dataproc clusters and want to retain the same behavior, please use `opsagent_nosyslog.sh`.
```
igorpeshansky, line 12: I would be careful about making backward compatibility statements — the Ops Agent's host metrics *are* different than those of the standalone ("legacy") agent. I'd also remove the parentheses around the versions… How about something like:
```suggestion
If you are looking to match the behavior of Dataproc image versions up to 2.1 with `--metric-sources=monitoring-agent-defaults`, which did not ingest syslogs from Dataproc nodes, please use `opsagent_nosyslog.sh`.
```
WDYT?

opsagent/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Install Ops Agent collecting system metrics only (no syslogs)
+
+```bash
+REGION=<region>
+CLUSTER_NAME=<cluster_name>
+gcloud dataproc clusters create ${CLUSTER_NAME} \
+    --image-version=2.2 \
+    --region=${REGION} \
+    --initialization-actions=gs://goog-dataproc-initialization-actions-${REGION}/opsagent/opsagent_nosyslog.sh
+```
```
igorpeshansky, line 40: If the recommended approach is to match Dataproc 2.1 and earlier, should we put the `opsagent_nosyslog.sh` action first?

opsagent/README.md
vamshikrishna-g: Updated

opsagent/README.md
vamshikrishna-g: Done

opsagent/README.md
vamshikrishna-g: Good one, incorporated.

igorpeshansky (approved):
LGTM :shipit:

PR #1127, opened 2024-01-17
Improving the init action test for Kafka

prince-cs:
For Kafka, adding the tests for submission of jobs and verifying instances to increase the coverage for overall scope. The test now runs kafka jobs with cruise-control and kafka-manager. Also, adding the test for cruise-control and kafka-manager instance verification.

### DIFF
--- a/kafka/kafka.sh
+++ b/kafka/kafka.sh
@@ -25,6 +25,7 @@ readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
 readonly RUN_ON_MASTER="$(/usr/share/google/get_metadata_value attributes/run-on-master || echo false)"
 readonly KAFKA_ENABLE_JMX="$(/usr/share/google/get_metadata_value attributes/kafka-enable-jmx || echo false)"
 readonly KAFKA_JMX_PORT="$(/usr/share/google/get_metadata_value attributes/kafka-jmx-port || echo 9999)"
+readonly INSTALL_KAFKA_PYTHON="$(/usr/share/google/get_metadata_value attributes/install-kafka-python || echo false)"
 
 # The first ZooKeeper server address, e.g., "cluster1-m-0:2181".
 ZOOKEEPER_ADDRESS=''
@@ -40,6 +41,11 @@ function retry_apt_command() {
     sleep 5
   done
   return 1
+}
+
+function recv_keys() {
+  retry_apt_command "apt-get install -y gnupg2 &&\
+                     apt-key adv --keyserver keyserver.ubuntu.com --recv-keys B7B3B788A8D3785C"
 }
 
 function update_apt_get() {
@@ -168,8 +174,29 @@ function install_and_configure_kafka_server() {
   wait_for_kafka
 }
 
+function install_kafka_python_package() {
+  KAFKA_PYTHON_PACKAGE="kafka-python==2.0.2"
+  if [[ "${INSTALL_KAFKA_PYTHON}" != "true" ]]; then
+    return
+  fi
+
+  if [[ "$(echo "${DATAPROC_IMAGE_VERSION} > 2.0" | bc)" -eq 1 ]]; then
+    /opt/conda/default/bin/pip install "${KAFKA_PYTHON_PACKAGE}" || { sleep 10; /opt/conda/default/bin/pip install "${KAFKA_PYTHON_PACKAGE}"; }
+  else
+    OS=$(. /etc/os-release && echo "${ID}")
+    if [[ "${OS}" == "rocky" ]]; then
+      yum install -y python2-pip
+    else
+      apt-get install -y python-pip
+    fi
+    pip2 install "${KAFKA_PYTHON_PACKAGE}" || { sleep 10; pip2 install "${KAFKA_PYTHON_PACKAGE}"; } || { sleep 10; pip install "${KAFKA_PYTHON_PACKAGE}"; }
+  fi
+}
+
 function main() {
+  recv_keys || err 'Unable to receive keys.'
   update_apt_get || err 'Unable to update packages lists.'
+  install_kafka_python_package
 
   # Only run the installation on workers; verify zookeeper on master(s).
   if [[ "${ROLE}" == 'Master' ]]; then
--- a/kafka/test_kafka.py
+++ b/kafka/test_kafka.py
@@ -1,5 +1,6 @@
 import os
 
+import pkg_resources
 from absl.testing import absltest
 from absl.testing import parameterized
 
@@ -9,7 +10,12 @@ from integration_tests.dataproc_test_case import DataprocTestCase
 class KafkaTestCase(DataprocTestCase):
     COMPONENT = 'kafka'
     INIT_ACTIONS = ['kafka/kafka.sh']
+    KAFKA_CRUISE_CONTROL_INIT_ACTION = ['kafka/kafka.sh', 'kafka/cruise-control.sh']
+    KAFKA_MANAGER_INIT_ACTION = ['kafka/kafka.sh', 'kafka/kafka-manager.sh']
     TEST_SCRIPT_FILE_NAME = 'validate.sh'
+    PYTHON2_VERSION = 'python2.7'
+    PYTHON3_VERSION = 'python3'
+    KAFKA_VERIFY_SCRIPT_PYTHON = 'verify_kafka_running.py'
 
     def verify_instance(self, name):
         self.upload_test_file(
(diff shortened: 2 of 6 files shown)


kafka/test_kafka.py
```diff
@@ -9,7 +10,15 @@
 class KafkaTestCase(DataprocTestCase):
     COMPONENT = 'kafka'
     INIT_ACTIONS = ['kafka/kafka.sh']
+    KAFKA_CRUISE_CONTROL_INIT_ACTION = ['kafka/kafka.sh', 'kafka/cruise-control.sh']
+    KAFKA_MANAGER_INIT_ACTION = ['kafka/kafka.sh', 'kafka/kafka-manager.sh']
     TEST_SCRIPT_FILE_NAME = 'validate.sh'
+    KAFKA_PYTHON_INIT_ACTION_PATH = 'gs://dataproc-kafka-test/kafka_python.sh'
+    KAFKA_PYTHON3_INIT_ACTION_PATH = 'gs://dataproc-kafka-test/kafka_python3.sh'
+    PYTHON2_VERSION = 'python2.7'
+    PYTHON3_VERSION = 'python3.10'
```
functicons, line 19: 3.10 might not be the right version for all Dataproc versions. I think we should use `python3` here.

kafka/verify_kafka_running.py
```diff
@@ -0,0 +1,159 @@
...
+def test_produce_and_consume_messages():
+    cluster_name = get_cluster_name()
+    kafka_runner = KafkaTestRunner(cluster_name)
+    kafka_runner.create_test_topic()
+    kafka_runner.publish_messages()
+    kafka_runner.consume_messages()
+
+
+def main():
+    time.sleep(10)
```
functicons, line 153: Is it to make sure Kafka is up and running? I don't think we should add it in the test, because users won't do the same. If wait is really necessary, it should be in the Kafka init action.

kafka/test_kafka.py
```diff
@@ -23,6 +32,22 @@ def __run_test_script(self, name):
         self.assert_instance_command(
             name, "bash {}".format(self.TEST_SCRIPT_FILE_NAME))
 
+    def __submit_pyspark_job(self, cluster_name):
+        if self.getImageVersion() >= pkg_resources.parse_version("2.1"):
+            verify_script = self.KAFKA_VERIFY_SCRIPT_PYTHON3
+            python_version = self.PYTHON3_VERSION
+        else:
+            verify_script = self.KAFKA_VERIFY_SCRIPT_PYTHON2
```
functicons, line 40: verify_kafka_running_py3.py and verify_kafka_running.py have a lot in common. I think we should merge then, and pass in parameters or detect different image versions inside the script.

kafka/test_kafka.py
```diff
@@ -23,6 +32,22 @@ def __run_test_script(self, name):
...
+            python_version = self.PYTHON2_VERSION
```
functicons, line 41: Do you know what happens if we use Python3 in 2.0?

kafka/test_kafka.py
```diff
@@ -9,7 +10,15 @@
 class KafkaTestCase(DataprocTestCase):
     COMPONENT = 'kafka'
     INIT_ACTIONS = ['kafka/kafka.sh']
+    KAFKA_CRUISE_CONTROL_INIT_ACTION = ['kafka/kafka.sh', 'kafka/cruise-control.sh']
+    KAFKA_MANAGER_INIT_ACTION = ['kafka/kafka.sh', 'kafka/kafka-manager.sh']
     TEST_SCRIPT_FILE_NAME = 'validate.sh'
+    KAFKA_PYTHON_INIT_ACTION_PATH = 'gs://dataproc-kafka-test/kafka_python.sh'
```
functicons, line 16: I think we should merge the 2 init actions into one, pass parameters or detect the env inside the init action.

functicons (changes requested):

kafka/test_kafka.py
prince-cs: done

kafka/test_kafka.py
prince-cs: based on the dataproc image version both the scripts are merged and tested locally.

kafka/test_kafka.py
prince-cs: merging them would require to change the script at the location at which it is stored as it is custom made init action script for kafka and while running with the repo it is not able to fetch this file. So,

functicons: I don't quite understand why it needs to be in this GCS bucket. Cannot it simply use the one from this repo? Just like the kafka.sh itself.

functicons (changes requested):

kafka/test_kafka.py
prince-cs: I tried that but it started throwing error saying file not found.

functicons: It should be the same as the kafka init action itself. Or you can extend the kafka init action with parameters to allow you do the work, this will bypass the problem and mabye some users also like it.

functicons (changes requested):

kafka/test_kafka.py
prince-cs: Passing the parameter and based on the dataproc image version being used for the init action, installing the kafka-python package accordingly with kafka init action itself.

kafka/kafka.sh
```diff
@@ -25,6 +25,7 @@ readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
 readonly RUN_ON_MASTER="$(/usr/share/google/get_metadata_value attributes/run-on-master || echo false)"
 readonly KAFKA_ENABLE_JMX="$(/usr/share/google/get_metadata_value attributes/kafka-enable-jmx || echo false)"
 readonly KAFKA_JMX_PORT="$(/usr/share/google/get_metadata_value attributes/kafka-jmx-port || echo 9999)"
+readonly DATAPROC_IMAGE="$(/usr/share/google/get_metadata_value attributes/image-version)"
```
functicons, line 28: Why is this added? It's not consistent with how other init actions get the image version.

kafka/kafka.sh
```diff
@@ -168,8 +174,25 @@ function install_and_configure_kafka_server() {
...
+      apt-get install -y python-pip
+    fi
+    pip2 install "${KAFKA_PYTHON_PACKAGE}" || { sleep 10; pip2 install "${KAFKA_PYTHON_PACKAGE}"; } || { sleep 10; pip install "${KAFKA_PYTHON_PACKAGE}"; }
+  fi
+}
+
 function main() {
+  recv_keys || err 'Unable to receive keys.'
   update_apt_get || err 'Unable to update packages lists.'
+  install_kafka_python_package
```
functicons, line 195: This new functionality should be guarded by a new meta data, e.g., install-kafka-python with default false.

functicons (changes requested):

kafka/kafka.sh
prince-cs: Okay got it. We have the DATAPROC_IMAGE_VERSION for this. No need to define it under the metadata.

kafka/kafka.sh
prince-cs: Understood.

kafka/kafka.sh
```diff
@@ -168,8 +174,27 @@ function install_and_configure_kafka_server() {
   wait_for_kafka
 }
 
+function install_kafka_python_package() {
+  KAFKA_PYTHON_PACKAGE="kafka-python==2.0.2"
+  if [[ "${INSTALL_KAFKA_PYTHON}" == "true" ]]; then
```
functicons, line 179: Prefer shortcut:
```
if [[ "${INSTALL_KAFKA_PYTHON}" != "true" ]]; then
  return
fi
```

kafka/kafka.sh
```diff
@@ -25,6 +25,7 @@ readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
 readonly RUN_ON_MASTER="$(/usr/share/google/get_metadata_value attributes/run-on-master || echo false)"
 readonly KAFKA_ENABLE_JMX="$(/usr/share/google/get_metadata_value attributes/kafka-enable-jmx || echo false)"
 readonly KAFKA_JMX_PORT="$(/usr/share/google/get_metadata_value attributes/kafka-jmx-port || echo 9999)"
+readonly INSTALL_KAFKA_PYTHON="$(/usr/share/google/get_metadata_value attributes/install-kafka-python || echo false)"
```
functicons, line 28: The new metadata should be documented in the README file.

functicons (changes requested):

kafka/kafka.sh
prince-cs: Done

kafka/kafka.sh
prince-cs: Done.

functicons (approved):
LGTM

PR #1130, opened 2024-01-19
Extending the Oozie test to submit the jobs

prince-cs:
For Oozie test, creating new test cases to submit the pyspark jobs for multiple components. The components covered are as follows:

1. distcp
2. hive
3. map-reduce
4. shell
5. pig
6. spark
7. sqoop

### DIFF
--- a/oozie/test_oozie.py
+++ b/oozie/test_oozie.py
@@ -1,5 +1,6 @@
 import os
 
+import pkg_resources
 from absl.testing import absltest
 from absl.testing import parameterized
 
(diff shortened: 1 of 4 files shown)


oozie/test_oozie.py
```diff
@@ -41,5 +46,57 @@ def test_oozie(self, configuration, machine_suffixes):
...
+                                         self.COMPONENT,
+                                         self.TEST_SCRIPT_JOB,
+                                         python_version,
+                                         python_version,
+                                         job_args))
+
+    def test_distcp_job(self):
+        self.__submit_pyspark_job(self.CONFIGURATION, "distcp")
+
+    def test_hive_job(self):
```
saravind7, line 77: We should also add a hive2 test. Do you want to do this as part of this or do you have a separate plan for new tests?

prince-cs: The test added for hive is passing "hive2" as an argument. So, I guess this would suffice the condition. I can take this on as a separate enhancement for this test.

prince-cs: Adding functionality for hive is being taken as a separate task and will be considered as an enchancement.

saravind7 (approved):

PR #1131, opened 2024-01-22
Expanding the livy test

prince-cs:
Expanding the livy test to manipulate the metadata fields and submit the pyspark job with livy up and running verification script.

### DIFF
--- a/livy/livy.sh
+++ b/livy/livy.sh
@@ -107,6 +107,14 @@ function main() {
 
   wget -nv --timeout=30 --tries=5 --retry-connrefused "${LIVY_URL}" -P "${temp}"
 
+  local OS
+  OS=$(. /etc/os-release && echo "${ID}")
+  if [[ "${OS}" == "rocky" ]]; then
+    yum install -y unzip
+  else
+    apt-get install -y unzip
+  fi
+
   unzip -q "${temp}/${LIVY_PKG_NAME}.zip" -d /usr/local/lib/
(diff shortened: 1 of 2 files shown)


saravind7 (approved):

livy/livy.sh
```diff
@@ -107,6 +107,14 @@ function main() {
 
   wget -nv --timeout=30 --tries=5 --retry-connrefused "${LIVY_URL}" -P "${temp}"
 
+  local OS
+  OS=$(. /etc/os-release && echo "${ID}")
```
cnauroth, line 111: Nitpick: please use just one space after `==`.

livy/livy.sh
```diff
@@ -107,6 +107,14 @@ function main() {
...
+  if [[ "${OS}" ==  "rocky" ]]; then
+    yum install -y unzip
+  else
+    sudo apt-get install -y unzip
```
cnauroth, line 115: Is there a reason the `else` path needs `sudo` but not the `if` path? I'd expect `sudo` is unnecessary on either path, because initialization actions always run as root:

https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/init-actions

> Initialization actions run as the **root** user. You **do not** need to use `sudo`.

cnauroth (changes requested):

livy/livy.sh
prince-cs: done

livy/livy.sh
prince-cs: done

cnauroth (approved):

PR #1138, opened 2024-02-08
Cloudsql proxy test for both mysql and postgresql which submits a hive job

prince-cs:


yigress:
there are build failures when create the postgres instance

ERROR: (gcloud.sql.instances.create) HTTPError 400: Invalid request: Only custom or shared-core instance Billing Tier type allowed for PostgreSQL database.
Step #2 - "dataproc-1.5-debian10-tests": 2024-02-08T17:56:27.192242838Z

PR #1139, opened 2024-02-20
Adding configuration in cloud build to run the tests for 2.2 Dataproc Images

prince-cs:
Adding the configuration for running the init action tests for 2.2 dataproc images as well and alongside making minor changes in accordance to support the init action for 2.2 images.

### DIFF
--- a/dask/dask.sh
+++ b/dask/dask.sh
@@ -43,7 +43,7 @@ readonly KNOX_HOME=/usr/lib/knox
 readonly KNOX_DASK_DIR=${KNOX_HOME}/data/services/dask/0.1.0
 readonly KNOX_DASKWS_DIR=${KNOX_HOME}/data/services/daskws/0.1.0
 
-CONDA_PACKAGES=("dask=${DASK_VERSION}" 'dask-bigquery' 'dask-ml' 'dask-sql')
+CONDA_PACKAGES=('dask-bigquery' 'dask-ml' 'dask-sql')
 
 if [[ "${DASK_RUNTIME}" == 'yarn' ]]; then
   # Pin `distributed` package version because `dask-yarn` 0.9
@@ -343,6 +343,8 @@ EOF
 
 
 function main() {
+  #Install dask with the help of conda as installing with mamba causes version conflicts
+  execute_with_retries "conda install -y dask=${DASK_VERSION}"
   # Install conda packages
   execute_with_retries "mamba install -y ${CONDA_PACKAGES[*]}"
 
--- a/hue/test_hue.py
+++ b/hue/test_hue.py
@@ -60,6 +60,8 @@ class HueTestCase(DataprocTestCase):
         'STANDARD',
     )
     def test_hue_job(self, configuration):
+        if self.getImageVersion() >= pkg_resources.parse_version("2.2"):
+            self.skipTest("Not supported in 2.2 image")
         self.createCluster(configuration, self.INIT_ACTIONS)
         self.__submit_pyspark_job(self.getClusterName())
 
--- a/livy/test_livy.py
+++ b/livy/test_livy.py
@@ -27,10 +27,16 @@ class LivyTestCase(DataprocTestCase):
         self.remove_test_script(self.TEST_SCRIPT_FILE_NAME, name)
 
     def _run_python_test_file(self, name):
-        self.assert_instance_command(
-            name,
-            "sudo apt-get install -y python3-pip && sudo pip3 install requests"
-        )
+        if self.getImageVersion() >= pkg_resources.parse_version("2.2"):
+            self.assert_instance_command(
+                name,
+                "sudo apt install python3-requests"
+            )
+        else:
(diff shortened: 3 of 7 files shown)


dask/dask.sh
```diff
@@ -43,7 +43,7 @@ readonly KNOX_HOME=/usr/lib/knox
 readonly KNOX_DASK_DIR=${KNOX_HOME}/data/services/dask/0.1.0
 readonly KNOX_DASKWS_DIR=${KNOX_HOME}/data/services/daskws/0.1.0
 
-CONDA_PACKAGES=("dask=${DASK_VERSION}" 'dask-bigquery' 'dask-ml' 'dask-sql')
+CONDA_PACKAGES=('dask-bigquery' 'dask-ml' 'dask-sql')
```
functicons, line 46: Why make this change?

dask/dask.sh
```diff
@@ -344,6 +344,7 @@ EOF
 
 function main() {
   # Install conda packages
+  execute_with_retries "conda install -y dask=2022.1"
```
functicons, line 347: Why don't use the env var DASK_VERSION?

livy/test_livy.py
```diff
@@ -27,10 +27,16 @@ def _verify_instance(self, name):
...
     def _run_python_test_file(self, name):
-        self.assert_instance_command(
-            name,
-            "sudo apt-get install -y python3-pip && sudo pip3 install requests"
-        )
+        if self.getImageVersion() == pkg_resources.parse_version("2.2"):
+            self.assert_instance_command(
+                name,
+                "sudo apt install python3-requests"
+            )
```
functicons, line 34: Why do this in 2.2?

functicons (changes requested):

dask/dask.sh
prince-cs: It was causing conflicts while being installed with mamba.

dask/dask.sh
prince-cs: ack

livy/test_livy.py
prince-cs: The generic command was throwing error.

dask/dask.sh
```diff
@@ -344,6 +344,7 @@ EOF
 
 function main() {
   # Install conda packages
+  execute_with_retries "conda install -y dask=${DASK_VERSION}"
```
functicons, line 347: Add a comment on why install dask with conda before other packages.

hue/test_hue.py
```diff
@@ -60,6 +60,8 @@ def test_hue(self, configuration, machine_suffixes):
         'STANDARD',
     )
     def test_hue_job(self, configuration):
+        if self.getImageVersion() == pkg_resources.parse_version("2.2"):
```
functicons, line 63: Can you make this change future-proof? Check >=2.2 instead of 2.2.

livy/test_livy.py
```diff
@@ -27,10 +27,16 @@ def _verify_instance(self, name):
         self.remove_test_script(self.TEST_SCRIPT_FILE_NAME, name)
 
     def _run_python_test_file(self, name):
-        self.assert_instance_command(
-            name,
-            "sudo apt-get install -y python3-pip && sudo pip3 install requests"
-        )
+        if self.getImageVersion() == pkg_resources.parse_version("2.2"):
```
functicons, line 30: ditto.

spark-rapids/test_spark_rapids.py
```diff
@@ -64,6 +64,9 @@ def test_spark_rapids(self, configuration, machine_suffixes, accelerator):
     if self.getImageVersion() == pkg_resources.parse_version("2.1") or self.getImageOs() == "rocky":
       self.skipTest("Not supported in image2.1 or rocky images")
 
+    if self.getImageVersion() == pkg_resources.parse_version("2.2") and self.getImageOs() == "debian":
+        self.skipTest("The Debian version (12) is not supported")
```
functicons, line 68: The condition and the message don't match, in theory, the Debian version can change (because of fix for high-severity CVEs or something similar) and the message will become misleading. Either change the condition or the message. If the reason is about OS distro and version, it is better to check OS directly.

Also update other places.

functicons (changes requested):

hue/test_hue.py
prince-cs: done

livy/test_livy.py
prince-cs: done

dask/dask.sh
prince-cs: done

spark-rapids/test_spark_rapids.py
prince-cs: The message is exactly copied from the error logs from the test. But the tests are passing for debian 10 and debian 11. Any ideas on what to exactly do here?

functicons (approved):

PR #1141, opened 2024-02-22
[oozie] Using $(hostname -s) rather than ${HOSTNAME}

cjac:


cjac:
that run failed due to resource constraints.  I'll re-launch and check on the build later today.

functicons (approved):
LGTM

PR #1142, opened 2024-02-23
[flink] Using $(hostname -s) rather than ${HOSTNAME}

cjac:


cjac:
no nodes available to schedule pods

kuldeepkk-dev (commented):
@cjac 

Is the newline required for "if" condition on line 224? Can we put conditions in one line or use '\' ?

kuldeepkk-dev (approved):

PR #1143, opened 2024-02-23
[hbase] Using $(hostname -s) vs ${HOSTNAME}

cjac:


kuldeepkk-dev (approved):
LGTM

PR #1144, opened 2024-02-23
[presto] Using $(hostname -s) rather than ${HOSTNAME}

cjac:


kuldeepkk-dev:
readonly PRESTO_MASTER_FQDN="$(/usr/share/google/get_metadata_value attributes/dataproc-master)"

@cjac - Can we change the variable name since value doesn't contain FQDN?

kuldeepkk-dev (approved):
LGTM

PR #1153, opened 2024-03-01
Added new init action for Datasketches

kuldeepkk-dev:
Added new initialization action for Apache Datasketches.

### DIFF
--- a/datasketches/datasketches.sh
+++ b/datasketches/datasketches.sh
@@ -35,13 +35,9 @@ readonly DS_LIBPATH="/usr/lib/datasketches"
 readonly SPARK_VERSION=$(spark-submit --version 2>&1 | sed -n 's/.*version[[:blank:]]\+\([0-9]\+\.[0-9]\).*/\1/p' | head -n1)
 readonly SPARK_JAVA_EXAMPLE_JAR="gs://spark-lib/datasketches/spark-java-thetasketches-1.0-SNAPSHOT.jar"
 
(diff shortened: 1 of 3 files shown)


cjac:
You'll need to add a BUILD file and some tests for the CI framework for gcbrun to trigger.  Check with Erick ; he recently added some to an action that was missing them.

datasketches/datasketches.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+#set -x
```
functicons, line 2: This should not be commented out. And the best practice is to use `set -euxo pipefail`

datasketches/datasketches.sh
```diff
@@ -0,0 +1,73 @@
...
+
+declare -A all_components=( [java]="3.1.0" [hive]="1.2.0" [memory]="2.0.0" [pig]="1.1.0" )
+
+#Download and configure datasketches libraries
+for lib in "${!all_components[@]}" 
+do
+	configure_libraries $lib ${all_components[$lib]}
+done
+
+#Deploy spark java thetasketches example jar
```
functicons, line 62: Prefer extracting helper functions over comments. Consider structure like:

function main() {
  init
  configure_libraries
  download_example_jar
}

functicons (changes requested):

functicons (approved):
LGTM

PR #1160, opened 2024-04-15
Fix oozie.sh - Debian 10 backports repo issue.

kuldeepkk-dev:


kuldeepkk-dev:
Tests are passing on 2.X and failing on 2.2 which are unrelated to this change.

oozie/oozie.sh
```diff
@@ -679,6 +697,10 @@ EOF
 
 
 function main() {
+  #Remove debian backports
+  if [[ ${OS_NAME} == debian ]]; then
```
prince-cs, line 701: Adding a check for dataproc version is also required, as the issue only affected debian 10 which is being used by dataproc 2.0

prince-cs (changes requested):

oozie/oozie.sh
kuldeepkk-dev: Done, added condition.

prince-cs (approved):
LGTM

cjac (approved):
LGTM

cjac:
2.1-debian11 fails with

Error from server (NotFound): pods "presubmit-2-1-debian11-32dbb602-ca42-47a1-aff2-cd66385a2c81" not found

This looks like an infrastructure failure in k8s...

Issue #1157, opened 2024-04-14
Kafka initialization actions script fails due to missing Debian repository


olbapjose:
Very recently, Dataproc clusters started to fail at creation, due to an error in the Kafka initialization script, caused by a Debian repository no longer available: 

**`https://deb.debian.org/debian buster backports-Release`**

The error says: 

- Initialization action failed. Failed action '**gs://goog-dataproc-initialization-actions-europe-west1/kafka/kafka.sh**', see output in: gs://XXXXX/google-cloud-dataproc-metainfo/a697c722-2bd7-440b-b4da-9494892703ac/XXXXXX-m/dataproc-initialization-script-0_output"

The contents of that file is the following. Any advice or workaround is more than welcome.

<img width="943" alt="image" src="https://github.com/GoogleCloudDataproc/initialization-actions/assets/34943494/2e313202-6481-4684-94bd-b22ede95ce4c">

kishida-yuki:
I'm  in the same situation.

cjac:
Thank you for the report.  We are addressing this issue with the highest priority.

akhanna213:
The fix https://github.com/GoogleCloudDataproc/initialization-actions/pull/1161 for gpu init actions has been verified. We are already working on the same fix patch for other init actions which are failing with the same error. 

For urgent fix, customers/developers can clone the init action and add the same lines of code as in the fix in their copy, and use it for cluster creation. Please note that we do not encourage our customers to use cloned init script as they will not have updated init actions, and they will have to clone it every time there is a change in the init actions repository. So unless urgent, please wait for the other fixes to go in :)

akhanna213:
Hi @ahmedetefy @olbapjose could you confirm if the error message is still the same. We have already rolled out the fix a while back.

olbapjose:
@akhanna213  Please see the image below and the attachment, which is the output file mentioned in the error.

![image](https://github.com/GoogleCloudDataproc/initialization-actions/assets/34943494/c21cf805-9468-4b41-8eda-4e0a08a88013)

[google-cloud-dataproc-metainfo_initialization-script-0_output.txt](https://github.com/GoogleCloudDataproc/initialization-actions/files/15387532/google-cloud-dataproc-metainfo_initialization-script-0_output.txt)

akhanna213:
[...]

PR #1161, opened 2024-04-15
Remove backports for releases older than oldstable

cjac:
https://github.com/GoogleCloudDataproc/initialization-actions/issues/1157

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -16,6 +16,24 @@
 
 set -euxo pipefail
 
+function remove_old_backports {
+  # This script uses 'apt-get update' and is therefore potentially dependent on
+  # backports repositories which have been archived.  In order to mitigate this
+  # problem, we will remove any reference to backports repos older than oldstable
+
+  # https://github.com/GoogleCloudDataproc/initialization-actions/issues/1157
+  oldstable=$(curl -s https://deb.debian.org/debian/dists/oldstable/Release | awk '/^Codename/ {print $2}');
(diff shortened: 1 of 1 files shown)


gpu/install_gpu_driver.sh
```diff
@@ -16,6 +16,28 @@
...
+  local stable
+  stable=$(curl -s https://deb.debian.org/debian/dists/stable/Release | awk '/^Codename/ {print $2}');
+
+  local matched_files
+  matched_files="$(grep -rsil '\-backports' /etc/apt/sources.list*)"
+  local filename
+  if [[ -n "$matched_files" ]]; then
+    for filename in "$matched_files"; do
+      grep -e "$oldstable-backports" -e "$stable-backports" "$filename" || \
+        sed -i -e 's/^.*-backports.*$//' "$filename"
```
yigress, line 36: this replace the whole line? should we just replace the expression of {non-stable}-backport

yigress: the format on older sources and new sources is different. this will remove too much on dataproc-2.2
dataproc-2.2
cat /etc/apt/sources.list.d/debian.sources
Types: deb deb-src
URIs: mirror+file:///etc/apt/mirrors/debian.list
Suites: bookworm bookworm-updates bookworm-backports
Components: main

Types: deb deb-src
URIs: mirror+file:///etc/apt/mirrors/debian-security.list
Suites: bookworm-security
Components: main

jayadeep-jayaraman (commented):
Can we ensure that this function is invoked only for 2.0 and 1.5

jayadeep-jayaraman (approved):

PR #1162, opened 2024-04-16
[gpu] add support for Dataproc 2.2 images

cjac:
Include open drivers on Debian 12


cjac:
tests keep failing due to lack of quota.  I think the tests need to be run during India business times since us-central1 might be quieter at that time.

cjac:
Fixed in another branch

PR #1164, opened 2024-04-18
Fix for bigtable.sh - Debian buster repo issue

kuldeepkk-dev:


prince-cs (approved):
LGTM

bigtable/bigtable.sh
```diff
@@ -243,6 +261,9 @@ function install_hbase() {
 }
 
 function main() {
+  if [[ ${OS_NAME} == debian ]] && [[ ${DATAPROC_IMAGE_VERSION} == 2.0 || ${DATAPROC_IMAGE_VERSION} == 1.5 ]]; then
```
cjac, line 264: I think 2.1 should not be excluded.  Maybe 2.2 since we changed the format and filename of the archive list, but 2.1 will eventually be oldstable and we'll want the code to work there, too.

cjac (commented):
I think the conditional around remove_old_backports could use a tweak...

cjac (approved):
LGTM

jayadeep-jayaraman (approved):

bigtable/bigtable.sh
```diff
@@ -243,6 +261,9 @@ function install_hbase() {
 }
 
 function main() {
+  if [[ ${OS_NAME} == debian ]] && [[ $(echo "${DATAPROC_IMAGE_VERSION} <= 2.1" | bc -l) == 1 ]]; then
```
jayadeep-jayaraman, line 264: OS_NAME is not defined in this script, therefore it will fail. Also, please ensure that /gcbrun runs fine.

kuldeepkk-dev:
Tests are passing on 2.0 & 2.1. 2.2 failures doesn't seem to be related to this change, merging this.

PR #1166, opened 2024-04-18
Fix for Debian 10 backports repo issue.

prince-cs:


cjac (commented):
Please include and call the function we've developed for the GPU installer.  Review Kuldeep's recent changes for examples

prince-cs:
> Please include and call the function we've developed for the GPU installer. Review Kuldeep's recent changes for examples

The changes are included which you added for GPU installer but in Oozie replacing the main repo with archive repo was required as the same was done for fixing image builds as well.

kuldeepkk-dev (approved):
LGTM

PR #1170, opened 2024-04-19
[zookeeper] remove backports for releases older than oldstable

cjac:
oldoldstable backports have been moved to archive ; attempting to fetch the release data will result in init action failure.


cjac:
Since there are no tests for this action, I propose that it be accepted without tests.

zookeeper/zookeeper.sh
```diff
@@ -84,6 +103,12 @@ if (($NODE_NUMBER > 2)); then
   exit 0
 fi
 
+# Remove > oldstable debian backports
+if [[ ${OS_NAME} == debian ]] &&
```
jayadeep-jayaraman, line 107: OS_NAME is not declared and hence this script will fail

PR #1172, opened 2024-04-22
[user-environment]Fix for Debian 10 backports repo issue.

prince-cs:


prince-cs:
Since there are no tests for this action, I propose that it be accepted without tests.

kuldeepkk-dev (approved):
LGTM

Issue #1175, opened 2024-04-22
rstudio.sh is unable to get the receive keys. Maybe due to invalid repo key.


kuldeepkk-dev:
```
$ apt-key adv --keyserver hkp://pool.sks-keyservers.net:80 --recv-keys XXXX
Executing: /tmp/apt-key-gpghome.uDjKWqitdH/gpg.1.sh --keyserver hkp://pool.sks-keyservers.net:80 --recv-keys XXX

gpg: keyserver receive failed: Server indicated a failure
```

vishalkarve15:
I was able to get the same repo key to work with a different keyserver.

PR #1179, opened 2024-05-02
Issue #1175: [rstudio] use keyserver.ubuntu.com

vishalkarve15:
Based on https://stackoverflow.com/questions/67251078/gpg-keyserver-send-failed-no-keyserver-available-when-sending-to-hkp-pool


kuldeepkk-dev (approved):
Tests are failing in 2.2-debian12 due to no internet access. This is because in 2.2 Dataproc clusters, we set `internal_ip_only=true` by default as per https://cloud.google.com/dataproc/docs/release-notes#February_16_2024

This should work by configuring CloudNAT or proxy on those images.

PR #1181, opened 2024-05-15
Allow private Oozie installation.

majduk:
Add support for metadata `oozie-repo-url` parameter, which allows to specify GCS bucket and folder where RPM/APT and JAR resources are stored. That allows running this action in an environment with no Internet access and thus control the source of packages.


oozie/oozie.sh
```diff
@@ -203,6 +205,39 @@ function configure_ssl() {
...
+  else
+    wget -nv --timeout=30 --tries=5 --retry-connrefused \
+      $url/$file -P "$dst"
+  fi
+}
+
+function dnf_install() {
+  pkg=$1
+  if [ -n "${OOZIE_REPO_URL}" ]; then
+    gsutil cp "${OOZIE_REPO_URL}/$pkg*" "/tmp/fname.rpm"
```
davidrabinowitz, line 223: Please convert `fname.rpm` to a random file name (add timestamp, uuid, etc) to avoid potential collusions.

majduk: Solved.

PR #1182, opened 2024-05-16
[gpu-driver]apt-get update fails due to changes in apt repo.

prince-cs:


jayadeep-jayaraman:
@viadea - pls take a look

viadea:
> @viadea - pls take a look

Actually there is a rick to use `--allow-releaseinfo-change`.
I prefer we fix the root cause instead of using above workaround.
But if GCP team are ok to use this workaround and take the risk for now, i am fine.

jayadeep-jayaraman:
@cjac - can you pls take a look at the previous comment

jayadeep-jayaraman:
We can revert this change now that we have a permanent fix

prince-cs:
Closing this PR as it is no longer required.

PR #1185, opened 2024-06-02
Fixing bigtable.sh

xby-G:
Fix the script to fix two scenarios:

 2.  Multiple source list files need to be updated

Test result: manually testing done.


cjac (approved):
LGTM

PR #1191, opened 2024-06-28
[RStudio] fixing the test failure for 2.0 ubuntu images.

prince-cs:
**libdeflate-dev** and **libdeflate0** are removed from the ubuntu library being used as source. 

- Make a temporary directory
- Download the archived version for both mentioned packages in the temp directory
- Using **dpkg** install both the packages and call the function only when the **OS** is **ubuntu** and the **image version** is **2.0**


kuldeepkk-dev (approved):
LGTM

PR #1192, opened 2024-06-28
Presubmits stability

prince-cs:


gpu/test_gpu.py
```diff
@@ -158,8 +158,8 @@ def test_install_gpu_with_mig(self, configuration, machine_suffixes,
         configuration,
         self.INIT_ACTIONS,
         zone=zone,
-        master_machine_type="n1-standard-4",
-        worker_machine_type="n1-standard-4",
+        master_machine_type="g2-standard-4",
```
cjac, line 161: replace all of the GPU_T4 with GPU_L4 if you want to make this change, plz

prince-cs: This change is only specific to this test, as GPU_L4 supports mig instances and machine type required is g2 series.

prince-cs:
`ERROR: (gcloud.dataproc.clusters.create) INVALID_ARGUMENT: Insufficient 'SSD_TOTAL_GB' quota. Requested 250.0, available 150.0. Your resource request exceeds your available quota. See https://cloud.google.com/compute/resource-usage. Use https://cloud.google.com/docs/quotas/view-manage#requesting_higher_quota to request additional quota.`

Can we bump this quota as well.

gpu/test_gpu.py
cjac: makes sense. thanks for clarifying.

cjac:
I just rebased onto the hue work, the test from which was failing.  I will look for other failing tests before running gcbrun again, since it's so expensive.

cjac:
2.0-rocky8 passed all tests, will you look at that.
rapids is failing on 2.2-rocky9 - but I think the rebase may fix that

there seems like a lot of things that might run now that I've rebased onto my latest work.  I'll try a run.

cjac:
The rapids test is failing on 2.2-rocky9

cjac:
Failed tests:
livy - 2.2-debian12
sql-proxy - 2.1-ubuntu20, 2.2-ubuntu-22
bigtable - 2.2-rocky9
gpu - 2.2-ubuntu22, 2.0-ubuntu18
rapids - 2.2-rocky9
tony - 2.1-ubuntu20

and many more, but that should get us through tomorrow.

Deependra-Patel:
We should split this CL into separate CLs each can be reviewed separately with context. Right now it is too broad.
Eg. all `boot_disk_size` change can be bundled together with reasoning why we are doing this, similarly separate CLs can be for each init action change, etc.

prince-cs:
In the latest build logs, the error is with dask init action script, as some dependency issues are present. Resolving those and apart from that I don't see any other failing init action tests.

PR #1193, opened 2024-06-28
[gpu] do not require signing metadata

cjac:
* wrapped signing code in conditional that checks for private key
* replaced tabs with spaces
* made use of get_metadata_attribute more frequently
* using default value of "" unless otherwise supplied
* using CUDA subversion of 12.4.1
* building source in /opt/install-nvidia-driver


prince-cs (approved):
LGTM

kuldeepkk-dev (approved):
LGTM

PR #1195, opened 2024-07-08
[bigtable] using modern url for shc-core and shc-examples jar packages

cjac:
Thanks to @dilipgodhia for the patch

tests done were manually, and only exercised the variable changed:

```
SCH_REPO="https://repo.hortonworks.com/content/repositories/releases/com/hortonworks"
SHC_VERSION='1.1.1-2.1-s_2.11'
SHC_JAR="shc-core-${SHC_VERSION}.jar"
SHC_EXAMPLES_JAR="shc-examples-${SHC_VERSION}.jar"
SHC_URL="${SCH_REPO}/shc-core/${SHC_VERSION}/${SHC_JAR}"
SHC_EXAMPLES_URL="${SCH_REPO}/shc-examples/${SHC_VERSION}/${SHC_EXAMPLES_JAR}"
cjac@cjac:~/src/github/cjac/initialization-actions$ wget "${SHC_EXAMPLES_URL}" ; echo $?
--2024-07-08 09:54:06--  https://repo.hortonworks.com/content/repositories/releases/com/hortonworks/shc-examples/1.1.1-2.1-s_2.11/shc-examples-1.1.1-2.1-s_2.11.jar
Resolving repo.hortonworks.com (repo.hortonworks.com)... 54.149.82.252, 35.164.245.218, 44.233.237.20
Connecting to repo.hortonworks.com (repo.hortonworks.com)|54.149.82.252|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 128482 (125K) [application/java-archive]
Saving to: ‘shc-examples-1.1.1-2.1-s_2.11.jar.2’

shc-examples-1.1.1-2.1-s_ 100%[===================================>] 125.47K  --.-KB/s    in 0.03s   

2024-07-08 09:54:06 (4.20 MB/s) - ‘shc-examples-1.1.1-2.1-s_2.11.jar.2’ saved [128482/128482]

0
```

### DIFF
--- a/bigtable/bigtable.sh
+++ b/bigtable/bigtable.sh
@@ -42,7 +42,7 @@ readonly BIGTABLE_HBASE_CLIENT_2X_VERSION='2.12.0'
 readonly BIGTABLE_HBASE_CLIENT_2X_JAR="bigtable-hbase-2.x-hadoop-${BIGTABLE_HBASE_CLIENT_2X_VERSION}.jar"
 readonly BIGTABLE_HBASE_CLIENT_2X_URL="${BIGTABLE_HBASE_CLIENT_2X_REPO}/${BIGTABLE_HBASE_CLIENT_2X_VERSION}/${BIGTABLE_HBASE_CLIENT_2X_JAR}"
 
-readonly SCH_REPO="https://repo.hortonworks.com/content/groups/public/com/hortonworks"
+readonly SCH_REPO="https://repo.hortonworks.com/content/repositories/releases/com/hortonworks"
 readonly SHC_VERSION='1.1.1-2.1-s_2.11'
 readonly SHC_JAR="shc-core-${SHC_VERSION}.jar"
 readonly SHC_EXAMPLES_JAR="shc-examples-${SHC_VERSION}.jar"
--- a/bigtable/test_bigtable.py
+++ b/bigtable/test_bigtable.py
@@ -70,7 +70,7 @@ class BigTableTestCase(DataprocTestCase):
     )
     def test_bigtable(self, configuration, machine_suffixes):
         self.createCluster(
-            configuration, self.INIT_ACTIONS, metadata=self.metadata)
+            configuration, self.INIT_ACTIONS, metadata=self.metadata, timeout_in_minutes=15)
 
         for machine_suffix in machine_suffixes:
             self.verify_instance("{}-{}".format(self.getClusterName(),


cjac:
rocky8 is not provisioning.  I'm going to try in a reproduction environment.  Please stand by.

cjac:
That may have just been a resource exhaustion failure due to the build project being heavily loaded.

I spun up a 2.0-rocky8 image without problem.

```
Mon Jul  8 11:28:05 AM PDT 2024
Mon Jul  8 11:33:03 AM PDT 2024
+ gcloud dataproc clusters create cluster-1718310842 \
  --master-boot-disk-type pd-ssd \
  --worker-boot-disk-type pd-ssd \
  --secondary-worker-boot-disk-type pd-ssd \
  --master-boot-disk-size 100 \
  --worker-boot-disk-size 100 \
  --secondary-worker-boot-disk-size 100 \
  --master-machine-type n1-standard-4 \
  --worker-machine-type g2-standard-4 \
  --num-masters=3 \
  --num-workers=2 \
  --num-secondary-workers=0 \
  --region us-west4 \
  --zone us-west4-a \
  --subnet subnet-cluster-1718310842 \
  --no-address \
  --shielded-secure-boot \
  --service-account=sa-cluster-1718310842@cjac-2021-00.iam.gserviceaccount.com \
  --tags=tag-cluster-1718310842 \
  --bucket cjac-dataproc-repro-1718310842 \
  --enable-component-gateway \
  --initialization-action-timeout=25m \
  --initialization-actions gs://cjac-dataproc-repro-1718310842/dataproc-initialization-actions/bigtable/bigtable.sh \
  --metadata bigtable-instance=cjac-bigtable0 \
  --worker-accelerator type=nvidia-l4 \
  --initialization-action-timeout=15m \
  --no-shielded-secure-boot \
  --image-version 2.0-rocky8 \
  --metadata dask-runtime=standalone \
  --metadata install-gpu-agent=true \
  --metadata cuda-version=12.4 \
  --max-idle=8h \
  --scopes https://www.googleapis.com/auth/cloud-platform
Waiting on operation [projects/cjac-2021-00/regions/us-west4/operations/3f3567e6-06f5-395b-9163-6d6fabfaded5].
Waiting for cluster creation operation...⠛                                                                                                                                                                                                                                                                                                   
WARNING: Consider using Auto Zone rather than selecting a zone manually. See https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/auto-zone
WARNING: The specified custom staging bucket 'cjac-dataproc-repro-1718310842' is not using uniform bucket level access IAM configuration. It is recommended to update bucket to enable the same. See https://cloud.google.com/storage/docs/uniform-bucket-level-access.
Waiting for cluster creation operation...done.                                                                                                                                                                                                                                                                                               
Created [https://dataproc.googleapis.com/v1/projects/cjac-2021-00/regions/us-west4/clusters/cluster-1718310842] Cluster placed in zone [us-west4-a].

real    4m58.068s
user    0m3.056s
sys     0m0.299s
+ date
Mon Jul  8 11:33:03 AM PDT 2024
```

I haven't exercised the tests, but the cluster is ONLINE after having run the init action with metadata parameters supplied.

cjac:
the build serveer is blocking all image variants from being tested.  We're blocking on this final log entry:

`2024-07-08T18:43:41.916598447Z INFO: Analyzed target //bigtable:test_bigtable (94 packages loaded, 835 targets configured).`

I am not terribly familiar with our Bazel test framework, so I will need to reach out to engineering for an assist here.

cjac:
timeout.

/gcbrun

cjac:
@hercynium are we using the correct url here?

hercynium:
I’ve only been at Cloudera one day and my job is to keep the cloud stuff running BUT, something that stands out to me is that the env var in question starts with ‘SCH’ while all the others start with ‘SHC’. Maybe that’s relevant, maybe not. At least by commenting I will get notifications on the issue. ☺

hercynium:
Oh, I think I see you’re asking for the correct URL.

I’ll ask around in between my onboarding stuff tomorrow.

cjac:
Thanks!  And any information about how long we can expect the jars to be maintained at their current location, where to go for planned updates, etc.  If it's not too much trouble.  Welcome to the team!

hercynium:
@cjac I think I know who to ask now but I just want to be clear on the question. Are you trying to determine the most canonical/stable public URL for Hortonworks' Maven repos? Or is there something missing from one of those URLs and you're looking for where it really lives?

cjac:
Yes.  Reliable/stable is the goal.

hercynium:
Looks like the most appropriate one is `https://repo.hortonworks.com/content/repositories/releases/` though `https://repo.hortonworks.com/content/groups/public/` should also contain all the same "stuff". The releng guy told me "they both should work" so I poked around and the public repo is a "virtual group" repo, which has *many* other repos mapped into it, including the `releases` repo. I'm guessing it includes virtual/mapped mirrors of the repos for all transient dependencies, including the stuff most people would normally get from places like Maven Central et al.

My advice? Use the `releases` repo unless it's missing something you need that's in the `public` one.

As I poked around, I found one more possibility: `https://repo.hortonworks.com/content/groups/releases/`

That one *might* be better than either of the others since it's also a virtual group repo, but seems to only contain the content from the `releases` repo... though stuff could be added in the future.

Good luck!

hercynium:
Oh! Somebody else just pinged me, said they always just used `https://repo.hortonworks.com/content/groups/public/` so maybe that's fine. The `releases` one should be the most "stable" in that releases should never just go away, but since the `public` repo has `releases` mapped into it... there shouldn't be any concern for those dependecies.

cjac:
Thanks for letting me know Steven.  The reason we're asking is because we saw reduced performance at various times with the `groups/public` url where the `repositories/releases` has been more resilient, from anecdotal evidence.  Have there been any changes to the infrastructure or maintenance windows that we should know of to avoid over-loading your systems?

I imagine you might ask us to keep a local cache for our installation scripts to pull from so that we are not impacted by any reduced performance on any network equipment between our two  points of presence or on the servers in the http pool for your load balancer.

hercynium:
I learned that earlier this year the maven repos were moved from being hosted on jfrog artifactory to sonatype nexus. It might be that the Hortonworks repos aren't cached as well, too. (Also, customers have access to dedicated repo servers that probably will perform better)

cjac:
Okay, I think we've confirmed that the url we're using now:

`https://repo.hortonworks.com/content/repositories/releases/com/hortonworks`

is better than or equal to the previous url:

`https://repo.hortonworks.com/content/groups/public/com/hortonworks`

With that, I think we're only pending a review from Kuldeep.

hercynium:
I just got one more piece of potentially useful info: All the Hortonworks repos are EOL, since Hortonworks/HDP was sunsetted in favor of Cloudera/CDP four years ago. It might be worth looking for equivalent jars/artifacts from CDP... or mirroring the hortonworks repos yourself. Those hortonworks repo servers will eventually go away.

hercynium:
This guide might be useful if you're running an entire HDP cluster: https://medium.com/hashmapinc/hortonworks-data-platform-2-6-5-in-place-upgrade-to-cloudera-data-platform-7-65dfc35091fd

cjac:
Thanks Prince!

Can you work with eng to cache this file on gcs so we don't lose support when the old repo goes offline as per Cloudera SRE?  Maybe this init action can be archived after 2.0 goes out of support.

bigtable/bigtable.sh
```diff
@@ -42,7 +42,7 @@ readonly BIGTABLE_HBASE_CLIENT_2X_VERSION='2.12.0'
 readonly BIGTABLE_HBASE_CLIENT_2X_JAR="bigtable-hbase-2.x-hadoop-${BIGTABLE_HBASE_CLIENT_2X_VERSION}.jar"
 readonly BIGTABLE_HBASE_CLIENT_2X_URL="${BIGTABLE_HBASE_CLIENT_2X_REPO}/${BIGTABLE_HBASE_CLIENT_2X_VERSION}/${BIGTABLE_HBASE_CLIENT_2X_JAR}"
 
-readonly SCH_REPO="https://repo.hortonworks.com/content/groups/public/com/hortonworks"
+readonly SCH_REPO="https://repo.hortonworks.com/content/repositories/releases/com/hortonworks"
```
prince-cs, line 45: I think it should be SHC instead of SCH (typo)

prince-cs:
So we just need to download the jars and store it in our GCS bucket. Right?

bigtable/bigtable.sh
cjac: I checked when Cloudera SRE suggested same.  This matches variable definition and we will remain backwards compatible with historic code if don't go fixing what ain't broke.  I advise leaving the potential typo.

kuldeepkk-dev (approved):
LGTM.

cjac:
Prince,

That is correct, and then update the same url section to fetch from our copy once Cloudera SRE brings their service down.  Or before.  Before would reduce potential of customer impact.

PR #1202, opened 2024-07-24
[hue] update action for July 2024

cjac:
modified:   hue/README.md

* line-wrapped and clarified important note

modified:   hue/hue.sh

* added functions os_id, is_debian, is_debian10, is_ubuntu
* only adding buster-backports to sources.list for debian10
* update_repo now uses is_debian || is_ubuntu rather than checking for apt-get in path
* setting DEBIAN_FRONTEND=noninteractive before installation of hue ; mysql-community-server had an important message that it was prompting the user to answer on a non-interactive console, halting tests.
* added note that we may need to specify this option with debconf

* jettisoned my function remove_old_backports with preference for Prince's much better replace_backports_repo

* using slightly different logic to determine whether we replace the backports repo

### DIFF
--- a/hue/README.md
+++ b/hue/README.md
@@ -32,7 +32,8 @@ installed:
 
 ## Important notes
 
-*   If you wish to use Oozie in Hue, it must be installed before running this
-    initialization action e.g. put the
-    [Oozie initialization action](../oozie/README.md) before this one in the
-    list to cloud.
+*   If you wish to use Oozie in Hue, it must be installed before
+    running this initialization action e.g. put the [Oozie
+    initialization action](../oozie/README.md) before this one in the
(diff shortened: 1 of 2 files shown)


cjac:
The previous run didn't seem to get the message that the installation was noninteractive and stopped again with the prompt.  I do not see this behavior in my reproduction environment.

cjac:
last test revealed that I had failed to define os_version on which is_debian depends.  This should fix that last failure.

cjac:
@prince-cs - this PR is ready for squash and merge.  I believe i granted you authorization to perform reviews.  Please review and approve if you believe we're ready to push this one out.  I know that #1192 is blocked on hue and probably others.

prince-cs (approved):

PR #1203, opened 2024-07-25
Update kafka.sh

akhanna213:
With primary workers naming convention getting changed, we cannot rely on the worker number anymore, hence changing the broker id for primary workers.

### DIFF
--- a/kafka/kafka.sh
+++ b/kafka/kafka.sh
@@ -150,7 +150,7 @@ function install_and_configure_kafka_server() {
     fi
   else
     # For worker nodes, broker ID is the worker ID.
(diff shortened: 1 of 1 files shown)


kafka/kafka.sh
```diff
@@ -150,7 +150,7 @@ function install_and_configure_kafka_server() {
     fi
   else
     # For worker nodes, broker ID is the worker ID.
-    BROKER_ID=$(hostname | sed 's/.*-w-\([0-9]*\)$/\1/g')
+    BROKER_ID=$((RANDOM % 10000))
```
jayadeep-jayaraman, line 153: Please add comment on why 10000

jayadeep-jayaraman (approved):

PR #1207, opened 2024-08-03
[gpu] Account for rocky9 kernel-devel-$(uname -r) moving to vault

cjac:
modified:   gpu/install_gpu_driver.sh

* inlining conditional blocks to reduce lines while maintaining readability
* cache results from some frequently-used programs
* decrease verbosity of apt-get with additional -q
* wrap nvidia-smi in a function which manages cache some
* add a work-around for kernel-devel-$(uname -r) missing from rocky9 repo
* use time to determine how long the long tasks take

modified:   gpu/test_gpu.py

* removed code skipping rocky tests


prince-cs (approved):

cjac:
https://github.com/GoogleCloudDataproc/initialization-actions/pull/1207#issuecomment-2267084553
> /gcbrun

This test passed.
BGiLKotSkedyuBN

PR #1208, opened 2024-08-05
[bigtable] retry failed hbase download attempts

cjac:
* using better implementation of debian backports repo update
* fetching shc jar from a local cache rather than from cloudera's deprecated repos


cjac:
@prince-cs these files should also be cached to GCS in case these archives go end of life, plz

prince-cs (approved):

PR #1210, opened 2024-08-08
Fixing installation of mysql-client

prince-cs:


cjac:
can we also update any `apt-key add -` ?

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -480,7 +480,7 @@ function install_mysql_cli() {
 
   log "Installing MySQL CLI ..."
   if command -v apt >/dev/null; then
-    apt update && apt install default-mysql-client -y
```
cjac, line 483: To support legacy clusters, we could instead:

```
mysql_pkg="mysql-client"
old_pkg_name="default-mysql-client"
if apt-cache search -n "${old_pkg_name}" | grep mysql ; then mysql_pkg="${old_pkg_name}" ; fi
apt install -yq "${mysql_pkg}"
```

cjac (approved):
maybe test with 1.5 if you have time for that.

PR #1212, opened 2024-08-08
Fixing dask test for presubmits

prince-cs:


dask/verify_dask_standalone.py
```diff
@@ -1,11 +1,6 @@
-from dask.distributed import Client
 import dask.array as da
-import dask_ml
-import dask_bigquery
 
 import numpy as np
 
-client = Client("localhost:8786")
```
cjac, line 8: Can you convince me that we don't care about dask.distributed ?

dask/verify_dask_yarn.py
```diff
@@ -1,13 +1,6 @@
-from dask_yarn import YarnCluster
-from dask.distributed import Client
 import dask.array as da
-import dask_ml
-import dask_bigquery
 
 import numpy as np
 
-cluster = YarnCluster()
```
cjac, line 9: This is the crux of the test.  We can't remove it and still call this a test.

cjac (commented):
Please ensure that the tests pass

dask/verify_dask_standalone.py
prince-cs: This variable is not being used in the script. It is only declared.

dask/verify_dask_yarn.py
prince-cs: This variable is not being used in the script. It is only declared.

dask/verify_dask_standalone.py
cjac: The process of instantiating the object is the test.  If an exception is thrown, the test is a FAIL as it should be

dask/verify_dask_yarn.py
cjac: That's not a declaration, that's an instantiation

PR #1213, opened 2024-08-08
Fixing horovod test for presubmits

prince-cs:


horovod/test_horovod.py
```diff
@@ -8,7 +9,7 @@ class HorovodTestCase(DataprocTestCase):
   COMPONENT = "horovod"
   INIT_ACTIONS = ["horovod/horovod.sh"]
   GPU_INIT_ACTIONS = ["gpu/install_gpu_driver.sh"] + INIT_ACTIONS
-  GPU_P100 = "type=nvidia-tesla-p100"
```
cjac, line 11: A100 or H100 should probably replace the P100.

horovod/test_horovod.py
```diff
@@ -26,6 +27,8 @@ def _submit_spark_job(self, script):
   def test_horovod_cpu(self, configuration, controller):
     if self.getImageOs() == 'rocky':
       self.skipTest("Not supported in Rocky Linux-based images")
+    if self.getImageVersion() > pkg_resources.parse_version("2.0"):
```
cjac, line 30: What is the reason that it's not supported on newer images?  I think the job we're trying to do here is to ensure that it *does* work on newer versions.  Horovod was last updated two weeks ago, so it's getting regular maintenance.  Let's see what we need to install in order to get this working.

cjac (changes requested):
Let's get this test passing rather than skipping on newer releases.

PR #1221, opened 2024-08-09
Fixing rstudio test for presubmits

prince-cs:


cjac:
what did this change accomplish other than changing the functionality for 1.5 and below?  I don't want to break legacy use cases if we can avoid it.

prince-cs:
> what did this change accomplish other than changing the functionality for 1.5 and below? I don't want to break legacy use cases if we can avoid it.

This change is done, as it is required by 1.5 ubuntu images. If we can merge this change then fine,, otherwise we would have to ignore it for the image candidate.

kuldeepkk-dev:
Have you tested this for 1.4? 1.5, 1.4 both are unsupported images per https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-version-clusters#unsupported-dataproc-image-versions

kuldeepkk-dev (approved):
LGTM. 
Prince confirmed that changes are working on 1.4 and 1.5 Dataproc images.

PR #1225, opened 2024-08-15
[dask] Updated to work with modern environments

cjac:
Updated to work with modern packages

dask.sh:
* removed most os condition tests ; this action is generally OS agnostic
* using the specific conda environment rather than the "default"
* declaring variables where they are used
* broke systemd units into scheduler and worker
* * only starting scheduler on primary master
* * only starting worker on master if dask-worker-on-master is not set to false
* updated knox xml to reference master node instead of localhost
* since dask-yarn only works with older libs, we are much more specific about the mamba/conda install packages

test_dask.py:
* removed rocky test inhibitions
* changed test machine type from e2-standard-2 to n1-standard-16
* self.verify_dask_standalone now takes an argument pointing to the master node
* using correct conda environment

verify_dask_standalone.py: 
* Scheduler now only runs on primary master node
* tests now specify the node of the scheduler
* Helpful debug messages now printed during test

verify_dask_yarn.py:
* YarnCluster object instantiated using skein.Client to reduce complexity
* YarnCluster object instantiation now wrapped in an exception handler with retry due to its flakiness
* Helpful debug messages now printed during test

### DIFF
--- a/dask/dask.sh
+++ b/dask/dask.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2020 Google LLC
+# Copyright 2020,2021,2023,2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


prince-cs (approved):

kuldeepkk-dev (approved):
LGTM

PR #1226, opened 2024-08-16
[gpu] non-spark rapids runtime should also use CUDA 12.4

cjac:
* using CUDA 12.4 by default for all use cases
* fetch nvidia kernel code from tarball instead of git clone


cjac:
@jakirkham - I've invited you to help maintain this repo since you have some good context about installing this software.  If you're available to review and approve this change once the tests complete in about an hour, it would help unblock me.

I'm knocking on wood as I say that.

prince-cs (approved):

PR #1232, opened 2024-09-10
[spark-rapids] Updating spark-rapids version to 24.08.1 and enabling the tests to run only for 2.1 and 2.2 images

prince-cs:


nvliyuan:
LGTM

kuldeepkk-dev (commented):
"Not supported in pre 2.0 images" - Can we modify this wordings since we don't support 2.0 either? 

Something like "Not supported in 2.0 and earlier images"?

kuldeepkk-dev (approved):

PR #1240, opened 2024-09-26
[gpu] workaround for broken apt trust database

cjac:
write trust keys to files under /usr/share/keyrings/
encode that path as signed-by into sources.list file for each repository
rename trusted.gpg
full-update to latest packages
force install of held packages to clear error

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -1127,6 +1127,30 @@ EOF
   fi
 
   #
+  # cran-r
+  #
+  if [[ -f /etc/apt/sources.list.d/cran-r.list ]]; then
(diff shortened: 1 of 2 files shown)


cjac:
With retries, Rocky did finish within 30 minutes.  This increase from 30m to 90m will improve reliability at the expense of longer test runs on some images.

cjac:
This patch appears to resolve the issue on all targeted platforms.  @santhoshvly - please try this script and report whether it works on your specific image version.  I did not test on a specific subversion and instead used 2.2-debian12

I have an internal issue open to place those trust keys in the right locations when they are imported, and to encode the path to the keyrings in the sources.list file.

prince-cs (approved):

PR #1242, opened 2024-09-30
[gpu] updates as result of rocky tests and running script with custom image generator

cjac:
Updates for Rocky, plus updates to allow the script to be run by the custom image generator

* reduce log noise a little
* Only unhold systemd on debian12
* encode get_metadata_value logic inline so this script works where `/usr/share/google/get_metadata_value` is not installed (custom-images)
* clean up argument passing to `execute_with_retries`
* allow access to DKMS keys for the duration of the install process instead of just during driver signing
* On Rocky, try to build drivers using DKMS with a fallback of installing pre-built binaries
* Do not install driver aliases on rocky
* remove extra `modprobe -r`
* in `configure_yarn`, skip update if `${HADOOP_CONF_DIR}` does not exist
* only `chown yarn:yarn` for directory list if directory list is not empty (including existing as a single element with element value 'None')
* corrected logic to modify the bigtop `sources.list` file rather than generating a new one
* using `$(os_codename)` rather than `${OS_CODENAME}`
* google cloud + logging/monitoring `sources.list` section tidy-up
* MySQL key is now fetched from `keyserver.ubuntu.com` rather than including inline
* making a backup of `/etc/apt/trusted.gpg` instead of removing without backup
* unholding `systemd` and `libsystemd0` on bookworm instead of force-upgrading

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -30,9 +30,10 @@ function is_debian()   { [[ "$(os_id)" == 'debian' ]] ; }
 function is_debian10() { is_debian && [[ "$(os_version)" == '10'* ]] ; }
 function is_debian11() { is_debian && [[ "$(os_version)" == '11'* ]] ; }
 function is_debian12() { is_debian && [[ "$(os_version)" == '12'* ]] ; }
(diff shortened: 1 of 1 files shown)


prince-cs (approved):
LGTM

PR #1245, opened 2024-10-04
pinning libsystemd0 version for debian 12 OS

prince-cs:


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -480,7 +485,11 @@ function install_mysql_cli() {
 
   log "Installing MySQL CLI ..."
   if command -v apt >/dev/null; then
-    apt update && apt install mysql-client -y
+    if is_debian12; then
+      apt update && apt install libsystemd0=252.26-1~deb12u2 mysql-client -y
```
cjac, line 489: apt does not have a stable abi.  Please use apt-get instead

cjac: sed -i -e 's/apt /apt-get /g'

cjac: Also: what are we doing here?  Why are we pinning the libsystemd0 package version in this way?  Shouldn't we add an entry to /etc/apt/preferences.d/mysql-repository-pin-600 so that it's global and not specific to this use case?

cjac: Again, what are we doing here?  Do you have a bug or a link describing the reason we're pinning systemd?  Can we get the oracle team to publish a mysql-client that matches Debian stable's version?

We have an internal issue from 2023 about pinning systemd for 1.5 images, but I would prefer not to pin packages, and instead talk to the maintainers about fixing bugs.

cjac: From a 2.12 system:

```
cjac@cluster-1718310842-w-0:~$ apt-cache show mysql-client
Package: mysql-client
Source: mysql-community
Version: 8.0.39-1debian12
Architecture: amd64
Maintainer: MySQL Release Engineering <mysql-build@oss.oracle.com>
Installed-Size: 75
Depends: mysql-community-client (= 8.0.39-1debian12)
Homepage: http://www.mysql.com/
Priority: optional
Section: database
Filename: pool/mysql-8.0/m/mysql-community/mysql-client_8.0.39-1debian12_amd64.deb
Size: 66660
SHA256: 083aab42c1e892bcc487bc514f7e335f379bca2639931db18e3ae34bc5c4f153
SHA1: d2e5d50a5770ca53a75ccb80096b82fac2cb8d50
MD5sum: 9db8ea8aa99826cbaec60a4a20b990df
Description: MySQL Client meta package depending on latest version
 The MySQL(TM) software delivers a very fast, multi-threaded, multi-user,
 and robust SQL (Structured Query Language) database server. MySQL Server
 is intended for mission-critical, heavy-load production systems as well
 as for embedding into mass-deployed software. MySQL is a trademark of
 Oracle. This is a meta package that depends on the latest mysql client
 package available in the repository.
Description-md5: 9562f9a79c0f92ac81175a806e88ebfd

```

cjac: Maintainer: MySQL Release Engineering <mysql-build@oss.oracle.com>

Let's reach out to them.

PR #1249, opened 2024-10-23
[bigtable] testing to ensure that hbase works with 2.1+

cjac:


cjac:
Test suite says yes.

cjac:
tests passed

PR #1250, opened 2024-10-26
[dask] merge changes from custom-images

cjac:
including improvements from work with custom-images repo

### DIFF
--- a/dask/dask.sh
+++ b/dask/dask.sh
@@ -22,10 +22,10 @@
 set -euxo pipefail
 
 function os_id()       { grep '^ID=' /etc/os-release | cut -d= -f2 | xargs ; }
-function os_version()  { grep '^VERSION_ID=' /etc/os-release | cut -d= -f2 | xargs ; }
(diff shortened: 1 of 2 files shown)


cjac:
2.0-ubuntu18 test failed to run this command.  I've added retry logic to the test code.  It feels like an intermittent error, since all other platforms passed this test.

 gcloud compute ssh test-dask-standard-2-0-20241026-192002-cix8-m --zone=us-central1-a --command="/opt/conda/miniconda3/envs/dask/bin/python verify_dask_yarn.py"

cjac:
This time it was only 2.0-ubuntu18 that failed.  It failed in the same way, but when it went to retry, it threw an exception about time not being imported.  So I've imported it now.  Maybe it will work this time?

cjac:
Rocky failed this time, too.  It looks like a race condition related to /var/cache/dnf/metadata_lock.pid ; wait until it is gone before mounting the ram disk.

cjac:
it looks like the yarn tests are passing here, so I should merge the tests into dask-rapids.

prince-cs (approved):

PR #1251, opened 2024-10-28
[gpu] merge install_gpu_driver.sh from custom-images

cjac:
* add is_debuntu to simplify common is_debian || is_ubuntu use case
    * prepend clean to execute_with_retries to reduce maximum image size
    * downloading assets to ram disk to reduce maximum image size
    * added sync to key points of the script to allow accurate measurement
      of maximum disk usage
    * moved initialization steps to new function prepare_to_install
    * created exit_handler function which gets executed before script exits
    * reduce log noise
    * enable dkms certificate use anywhere in main function
    * this version tested to work with secure boot enabled on all but debian10 and debian11 variants when launched with an image customized with [build-current-images.sh](https://github.com/cjac/custom-images/blob/test-dask-rapids-20241021/examples/secure-boot/build-current-images.sh)

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -120,21 +120,20 @@ readonly ROLE
 
 # CUDA version and Driver version
 # https://docs.nvidia.com/deeplearning/frameworks/support-matrix/index.html
+# https://developer.nvidia.com/cuda-downloads
 readonly -A DRIVER_FOR_CUDA=(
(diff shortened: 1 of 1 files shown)


cjac:
that machine type is way too small.  n1-standard-8 is the correct size for this installer.

cjac:
Everything in the last one worked except for rocky9.  for some reason, the installation of screen is causing the internet to break.  I've added a retry for failed dnf install commands.  I've also increased the timeout from 30m to 90m to make sure rocky9 has enough time to do the installation.

PR #1256, opened 2024-11-05
Running more tests in parallel for faster presubmit outcome

Deependra-Patel:
Now the tests finish < 2 hours for 2.1-deb11 which were earlier timing out in 4 hours, so atleast 2x faster.

`kubectl run` doesn't allow specifying CPU/Memory requirements in single line. The test cluster uses Autopilot so resources will be used only when tests are run.


medb (approved):

Deependra-Patel:
Checked that some integ tests are failing due to unrelated reason which is being investigated separately by @prince-cs. None of them are due to quota issues which could have been caused with this change. Merging.

cjac (approved):
Thanks for the update!  And thanks for catching the line I further to remove.

PR #1264, opened 2024-11-14
Horovod test fix 

prince-cs:
- Adding the correct versions for the framework versions as they will be varying on the Dataproc version being used. 
- Ignoring the tests run for Dataproc versions 2.1 and 2.2 because it is not supported with the current framework versions being suggested and downgrading doesn't allow the frameworks to be installed.


cjac (approved):
LGTM

PR #1269, opened 2024-12-04
[gpu][spark-rapids] Fix MIG script

### LINKED ISSUE #1259: [gpu][spark-rapids] Consolidate mig.sh Scripts and Sync Driver Installation Steps Across Copies
I've observed that there are multiple mig.sh scripts in both [/spark-rapids/mig.sh](https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/spark-rapids/mig.sh) and [/gpu/mig.sh](https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/gpu/mig.sh). 

Additionally, the driver installation steps in [/spark-rapids/mig.sh](https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/spark-rapids/mig.sh) and [/spark-rapids/spark-rapids.sh](https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/spark-rapids/spark-rapids.sh) have diverged, with most updates happening in [/spark-rapids/spark-rapids.sh](https://github.com/GoogleCloudDataproc/initialization-actions/blob/master/spark-rapids/spark-rapids.sh)

I wish we could have a single source for `mig.sh`

SurajAralihalli:
Resolves #1259

This PR updates the MIG script to align the driver installation steps with those in `spark-rapids.sh`.

### DIFF
--- a/spark-rapids/mig.sh
+++ b/spark-rapids/mig.sh
@@ -208,6 +208,9 @@ if (! test -v DATAPROC_IMAGE_VERSION) && test -v DATAPROC_VERSION; then
   DATAPROC_IMAGE_VERSION="${DATAPROC_VERSION}"
 fi
 
+# Fetch Linux Family distro and Dataproc Image version
+readonly OS_NAME=$(lsb_release -is | tr '[:upper:]' '[:lower:]')
+
 # CUDA version and Driver version config
 CUDA_VERSION=$(get_metadata_attribute 'cuda-version' '12.4.1')  #12.2.2
 NVIDIA_DRIVER_VERSION=$(get_metadata_attribute 'driver-version' '550.54.15') #535.104.05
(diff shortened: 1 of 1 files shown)


SurajAralihalli:
Hey @cjac, I can't add you as a reviewer for this PR. Could you please take a look? Thanks!  
This PR was tested for NVIDIA/spark-rapids#11675.

cjac (commented):
Hi there!  Take a look at the latest version of the utility scripts here:

https://github.com/LLC-Technologies-Collier/initialization-actions/blob/db33fce27205535f440fa5cc0545cabc591e6730/gpu/install_gpu_driver.sh#L19

function os_id()       ( set +x ;  grep '^ID=' /etc/os-release | cut -d= -f2 | xargs ; )
function os_version()  ( set +x ;  grep '^VERSION_ID=' /etc/os-release | cut -d= -f2 | xargs ; )
function os_codename() ( set +x ;  grep '^VERSION_CODENAME=' /etc/os-release | cut -d= -f2 | xargs ; )

function version_ge() ( set +x ;  [ "$1" = "$(echo -e "$1\n$2" | sort -V | tail -n1)" ] ; )
function version_gt() ( set +x ;  [ "$1" = "$2" ] && return 1 || version_ge $1 $2 ; )
function version_le() ( set +x ;  [ "$1" = "$(echo -e "$1\n$2" | sort -V | head -n1)" ] ; )
function version_lt() ( set +x ;  [ "$1" = "$2" ] && return 1 || version_le $1 $2 ; )

readonly -A supported_os=(
  ['debian']="10 11 12"
  ['rocky']="8 9"
  ['ubuntu']="18.04 20.04 22.04"
)

# dynamically define OS version test utility functions
if [[ "$(os_id)" == "rocky" ]];
then _os_version=$(os_version | sed -e 's/[^0-9].*$//g')
else _os_version="$(os_version)"; fi
for os_id_val in 'rocky' 'ubuntu' 'debian' ; do
  eval "function is_${os_id_val}() ( set +x ;  [[ \"$(os_id)\" == '${os_id_val}' ]] ; )"

  for osver in $(echo "${supported_os["${os_id_val}"]}") ; do
    eval "function is_${os_id_val}${osver%%.*}() ( set +x ; is_${os_id_val} && [[ \"${_os_version}\" == \"${osver}\" ]] ; )"
    eval "function ge_${os_id_val}${osver%%.*}() ( set +x ; is_${os_id_val} && version_ge \"${_os_version}\" \"${osver}\" ; )"
    eval "function le_${os_id_val}${osver%%.*}() ( set +x ; is_${os_id_val} && version_le \"${_os_version}\" \"${osver}\" ; )"
  done
done

function is_debuntu()  ( set +x ;  is_debian || is_ubuntu ; )

function os_vercat()   ( set +x
  if   is_ubuntu ; then os_version | sed -e 's/[^0-9]//g'
  elif is_rocky  ; then os_version | sed -e 's/[^0-9].*$//g'
                   else os_version ; fi ; )

cjac (commented):
I want to create a template system so that we don't have to keep these functions synchronized manually.

A job for another day.

spark-rapids/mig.sh
```diff
@@ -26,34 +26,207 @@
...
-readonly ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
-DATAPROC_IMAGE_VERSION=$(/usr/share/google/get_metadata_value image|grep -Eo 'dataproc-[0-9]-[0-9]'|grep -Eo '[0-9]-[0-9]'|sed -e 's/-/./g')
-echo "${DATAPROC_IMAGE_VERSION}" >> /usr/local/share/startup-mig-log
 
 # CUDA version and Driver version config
-CUDA_VERSION=$(get_metadata_attribute 'cuda-version' '12.2.2')  #12.2.2
-NVIDIA_DRIVER_VERSION=$(get_metadata_attribute 'driver-version' '535.104.05') #535.104.05
+CUDA_VERSION=$(get_metadata_attribute 'cuda-version' '12.4.1')  #12.2.2
+NVIDIA_DRIVER_VERSION=$(get_metadata_attribute 'driver-version' '550.54.15') #535.104.05
 CUDA_VERSION_MAJOR="${CUDA_VERSION%.*}"  #12.2
```
cjac, line 217: The cuda and driver version selection code is presently getting a revamp, too.  You might want to wait for that...

spark-rapids/mig.sh
```diff
@@ -92,74 +265,168 @@ EOF
...
+      curl -o driver.run \
+        "https://download.nvidia.com/XFree86/Linux-x86_64/${NVIDIA_DRIVER_VERSION}/NVIDIA-Linux-x86_64-${NVIDIA_DRIVER_VERSION}.run"
+      # Install all but kernel driver
+      bash driver.run --no-kernel-modules --silent --install-libglvnd
+      rm driver.run
+
+      WORKDIR=/opt/install-nvidia-driver
+      mkdir -p "${WORKDIR}"
+      pushd $_
+      # Fetch open souce kernel module with corresponding tag
```
cjac, line 348: *source

spark-rapids/mig.sh
```diff
@@ -92,74 +265,168 @@ EOF
...
+    fi
 
-    dpkg -i /tmp/local-installer.deb
-    cp /var/cuda-repo-ubuntu${UBUNTU_VERSION}04-${CUDA_VERSION_MAJOR//./-}-local/cuda-*-keyring.gpg /usr/share/keyrings/
-    execute_with_retries "apt-get update"    
-    
-    execute_with_retries "apt-get install -y -q --no-install-recommends cuda-drivers-${NVIDIA_DRIVER_VERSION_PREFIX}"
-    execute_with_retries "apt-get install -y -q --no-install-recommends cuda-toolkit-${CUDA_VERSION_MAJOR//./-}"
 
     # enable a systemd service that updates kernel headers after reboot
```
cjac, line 412: No, bad idea.  The kernel should not change.  To get a new kernel, upgrade the image.  The kernel version should be pinned, in my opinion.  Otherwise, there are too many failure states.

cjac (commented):
I think these changes should wait on the next PR.  There are a lot of things happening in the GPU driver that should be included,  IMHO.

spark-rapids/mig.sh
SurajAralihalli: I agree. Currently, reboots can trigger kernel upgrades, potentially breaking the drivers. Ideally, we should avoid kernel upgrades.

spark-rapids/mig.sh
SurajAralihalli: Thanks for letting me know. Can we include these improvements in a different PR to unblock the customers (https://github.com/NVIDIA/spark-rapids/issues/11675)

cjac:
> I want to create a template system so that we don't have to keep these functions synchronized manually.
> 
> A job for another day.

That day is today.  Please take a look at #1282

spark-rapids/mig.sh
cjac: I don't know if I can support customers running outdated images.  I guess we've been doing it to date by applying main force.  But it's not really sustainable.  Can you tell me whether your users will be able to move to a new dataproc image to get new kernels?

The customer can only apply new kernel fixes if they reboot into a new kernel.  When they do that, they will lose the drivers that I build for them just for that kernel version.  I do not enable DKMS because we do not recommend changing kernels.  Since DKMS is not installed, the system will not even attempt to build a new version of the kernel driver.  The assumption is that the service account is only granted access to the driver signing private key material during the installation, or more realistically, during custom image generation.

On systems with secure boot enabled, the user will not be able to fetch the signing material required for driver re-build.  It's just a lot less hassle if customers decommission old images and move to new images regularly.

spark-rapids/mig.sh
cjac: the revamp is in #1275 and is ready for review.  If you can take a look and leave a comment, it will give the final reviewers solace to know that our users have exercised the changes.

SurajAralihalli:
> That day is today. Please take a look at #1282

Thanks for sharing! I'm currently traveling, I'll get to it as soon as I can.

spark-rapids/mig.sh
SurajAralihalli: We transitioned away from installing drivers using run files because this approach caused issues with drivers during kernel upgrades and subsequent reboots. This was likely due to the kernel version not being pinned, leading to kernel upgrades.  

Are you suggesting we pin the kernel version for each Dataproc image version (e.g., v2.0, v2.1, v2.2)?  
If a user has a long-running GPU cluster with v2.2 and needs to upgrade the kernel for security patches, what would be the recommended process?

SurajAralihalli: > It's just a lot less hassle if customers decommission old images and move to new images regularly

- Does Dataproc (v2.0, v2.1, v2.2) support minor versions? 
- How can one ensure the latest kernel is installed during cluster creation, specifically at the time of installing the run file?

cjac:
I've moved the mig-specific changes into PR #1284

cjac (approved):
Thank you for the changes.  These look good.

cjac:
mig.sh tests do not work right now, but we have customers indicating that this version works for them.

PR #1273, opened 2024-12-10
Pin bazel version to 7.4.1

prince-cs:


cloudbuild/Dockerfile
```diff
@@ -18,7 +19,12 @@ RUN echo "deb [arch=amd64 signed-by=${bazel_kr_path}] http://storage.googleapis.
     dd of=/etc/apt/sources.list.d/bazel.list status=none && \
     apt-get update -qq
 RUN apt-get autoremove -y -qq && \
-    apt-get install -y -qq openjdk-8-jdk python3-setuptools bazel >/dev/null 2>&1 && \
+    apt-get install -y -qq openjdk-8-jdk python3-setuptools unzip g++ zlib1g-dev >/dev/null 2>&1 && \
     apt-get clean
 
+RUN curl -Lo /tmp/bazel_7.4.1-linux-x86_64.deb https://github.com/bazelbuild/bazel/releases/download/7.4.1/bazel_7.4.1-linux-x86_64.deb
```
Deependra-Patel, line 25: I see CJ has also created a PR https://github.com/GoogleCloudDataproc/initialization-actions/pull/1274#pullrequestreview-2494285572 which doesn't do curl so looks better way to me.

cjac:
it's best to implement a solution using apt-get since it verifies signatures at the time of installation

PR #1274, opened 2024-12-10
[bazel] pinning bazel version used by cloudbuild/Dockerfile

cjac:
Promotion to default of bazel-8 has caused our builds to fail.  Rather than using default, we will use bazel-7 until such time as we patch out codebase to correct for bazel-8's breaking changes

### DIFF
--- a/cloudbuild/Dockerfile
+++ b/cloudbuild/Dockerfile
@@ -9,7 +9,8 @@ COPY --chown=ia-tests:ia-tests . /init-actions
 
 # Install Bazel:
 # https://docs.bazel.build/versions/master/install-ubuntu.html
-ENV bazel_kr_path=/usr/share/keyrings/bazel-keyring.gpg
+ENV bazel_kr_path=/usr/share/keyrings/bazel-keyring.gpg \
+    bazel_version=7.4.0
 RUN apt-get install -y -qq curl >/dev/null 2>&1 && \
     apt-get clean
 RUN /usr/bin/curl https://bazel.build/bazel-release.pub.gpg | \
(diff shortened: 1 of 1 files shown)


cjac:
I am not running gcbrun because it will try to run *all* tests, and that will be expensive and not tell us anything that we don't already know.

cloudbuild/Dockerfile
```diff
@@ -9,16 +9,22 @@ COPY --chown=ia-tests:ia-tests . /init-actions
...
     apt-get update -qq
-RUN apt-get autoremove -y -qq && \
-    apt-get install -y -qq openjdk-8-jdk python3-setuptools bazel >/dev/null 2>&1 && \
+RUN apt-get update -y -qq && \
+    apt-get autoremove -y -qq && \
+    apt-get install -y -qq default-jdk python3-setuptools bazel-${bazel_version} > /dev/null 2>&1 && \
     apt-get clean
 
+# Set bazel-${bazel_version} as the default bazel alternative in this container
+RUN update-alternatives --install /usr/bin/bazel bazel /usr/bin/bazel-${bazel_version} 1 && \
```
Deependra-Patel, line 27: We shouldn't need to set here as 7.4.0 is the only version on this machine. Let's remove it.

Deependra-Patel (changes requested):

cloudbuild/Dockerfile
cjac: I think it's required.  An alternative cannot be selected unless one is defined.  And without this command, bazel was not in the PATH

cjac: yeah, I tested and bazel test fails with the following error:

```
+ bazel test --jobs=10 --local_test_jobs=10 --flaky_test_attempts=3 --action_env=INTERNAL_IP_SSH=true --test_output=errors --test_arg=--image_version=2.0-debian10 gpu:test_gpu
gpu/run-bazel-tests.sh: line 17: bazel: command not found
```

cjac: that is to say that I cannot implement your request to remove the `update-alternatives` commands

cjac:
closing this to yield to an upcoming PR

cjac:
fixed in #1283

Issue #1268, opened 2024-12-02
[gpu] versions installed by gpu/install_gpu_driver.sh do not match requested versions


cjac:
In the previous merge to master, when specifying versions explicitly, install_gpu_driver.sh will allow the OS vendor to select the driver using the version specified as a hint.

Customers want to have instead their input used if it is provided; attempt to select reasonable defaults only if they are not provided.  Version numbers should be extracted from `cuda-url` and `gpu-driver-url` if not explicitly provided in `gpu-driver-version` and `cuda-version` metadata attributes.

Refer to documentation from cuda:

https://docs.nvidia.com/deploy/cuda-compatibility/

Implementation suggestions:

When `cuda-url` is specified but not the `gpu-driver-version`, nor the `gpu-driver-url`, the version should be extracted from the `cuda-url`
* since many of the `gpu-driver-version` [values](https://github.com/LLC-Technologies-Collier/initialization-actions/blob/5c55a699b478e069ad1a6c4c4319d6bdfd9dc6a9/gpu/install_gpu_driver.sh#L389) included in the `cuda-url` are not indexed at [XFree86/Linux-x86_64](https://download.nvidia.com/XFree86/Linux-x86_64/), instead use the preferred `gpu-driver-version` for that `cuda-version`

Create [new function](https://github.com/LLC-Technologies-Collier/initialization-actions/blob/5c55a699b478e069ad1a6c4c4319d6bdfd9dc6a9/gpu/install_gpu_driver.sh#L140) to contain code for better mapping to support matrix.

PR #1275, opened 2024-12-13
[gpu] strict driver and cuda version assignment

cjac:
Resolves Issues
* #1268 

gpu/install_gpu_driver.sh

* Driver version defaults to version in driver .run file if specified
* CUDA version defaults to version in cuda .run file if specified

* exclusively using .run file installation method for cuda and driver installation
* * Installing non-open driver from .run file on rocky8
* build nccl from source, since that is the only mechanism which supports all Dataproc OSs
* wrap expensive functions in completion checks to reduce re-run time when testing manually
* cache build results in GCS
* waiting on apt lock when it exists
* only install build dependencies if build is necessary
* fix problem with ops agent not installing ; using venv
* mark task completion by creating a file rather than setting a variable
* added functions to check and report secure-boot and os version details       
* if the correct metadata attributes are specified, pytorch will be installed, but not by default                      

gpu/manual-test-runner.sh
* order commands correctly
* point to origin rather than staging repo

gpu/test_gpu.py
* clearer test skipping logic


cjac:
Tests took 43 minutes to run this time.  I think it's because the binary driver build cache was invalidated.  Let's see if this one takes more like 14 minutes.

cjac:
the intention is for this PR to resolve issue #1268

cloudbuild/Dockerfile
```diff
@@ -9,16 +9,23 @@ COPY --chown=ia-tests:ia-tests . /init-actions
 
 # Install Bazel:
 # https://docs.bazel.build/versions/master/install-ubuntu.html
-ENV bazel_kr_path=/usr/share/keyrings/bazel-keyring.gpg
+ENV bazel_kr_path=/usr/share/keyrings/bazel-keyring.gpg \
+    bazel_version=7.4.0 \
```
cjac, line 13: Pinning to 7.4.0 is necessary since the release of and default set to version 8

cloudbuild/Dockerfile
```diff
@@ -9,16 +9,23 @@ COPY --chown=ia-tests:ia-tests . /init-actions
...
+    bazel_repo_data="http://storage.googleapis.com/bazel-apt stable jdk1.8" \
+    DEBIAN_FRONTEND=noninteractive
 RUN apt-get install -y -qq curl >/dev/null 2>&1 && \
     apt-get clean
-RUN /usr/bin/curl https://bazel.build/bazel-release.pub.gpg | \
-    gpg --dearmor -o "${bazel_kr_path}"
-RUN echo "deb [arch=amd64 signed-by=${bazel_kr_path}] http://storage.googleapis.com/bazel-apt stable jdk1.8" | \
+RUN /usr/bin/curl -s https://bazel.build/bazel-release.pub.gpg | \
```
cjac, line 18: reduce noise in docker build with the -s (silent) option

cloudbuild/Dockerfile
```diff
@@ -9,16 +9,23 @@ COPY --chown=ia-tests:ia-tests . /init-actions
...
     dd of=/etc/apt/sources.list.d/bazel.list status=none && \
     apt-get update -qq
-RUN apt-get autoremove -y -qq && \
-    apt-get install -y -qq openjdk-8-jdk python3-setuptools bazel >/dev/null 2>&1 && \
+RUN apt-get autoremove -y -qq > /dev/null 2>&1 && \
+    apt-get install -y -qq default-jdk python3-setuptools bazel-${bazel_version} > /dev/null 2>&1 && \
     apt-get clean
 
+# Set bazel-${bazel_version} as the default bazel alternative in this container
+RUN update-alternatives --install /usr/bin/bazel bazel /usr/bin/bazel-${bazel_version} 1 && \
```
cjac, line 28: The file /usr/bin/bazel does not exist in the bazel-${version} package ; set up an alternative and use it

cloudbuild/create-key-pair.sh
```diff
@@ -0,0 +1,135 @@
+#!/bin/bash
```
cjac, line 1: this file copied from https://github.com/GoogleCloudDataproc/custom-images/tree/2024.11/examples/secure-boot

cloudbuild/presubmit.sh
```diff
@@ -70,6 +70,7 @@ determine_tests_to_run() {
     changed_dir="${changed_dir%%/*}/"
     # Run all tests if common directories modified
     if [[ ${changed_dir} =~ ^(integration_tests|util|cloudbuild)/$ ]]; then
+      continue
```
cjac, line 73: must remove this before merging

cjac:
slightly longer than the 14 minutes I predicted, this run completed in 16:29

gpu/test_gpu.py
```diff
@@ -299,7 +355,43 @@ def test_install_gpu_cuda_nvidia_with_spark_job(self, configuration, machine_suf
...
+
+    self.verify_instance_spark()
+
+  @parameterized.parameters(
+#    ("SINGLE", ["m"], GPU_T4, GPU_T4, "11.8", ''),
+#    ("STANDARD", ["m"], GPU_T4, None, "12.0"),
+    ("STANDARD", ["m", "w-0", "w-1"], GPU_T4, GPU_T4, "11.8", 'rocky', '2.0'),
+    ("STANDARD", ["m", "w-0", "w-1"], GPU_T4, GPU_T4, "12.4", 'rocky', '2.1'),
+    ("STANDARD", ["m", "w-0", "w-1"], GPU_T4, GPU_T4, "12.0", 'rocky', '2.2'),
+    ("KERBEROS", ["m", "w-0", "w-1"], GPU_T4, GPU_T4, "12.6", 'rocky', '2.2'),
```
cjac, line 374: We now exercise Kerberos in this test suite.  True, it's only tested on 2.2-rocky9 with CUDA 12.6, but we could change any of the STANDARD definitions above to KERBEROS to improve that set.

cjac:
okay, taking a little break...

jayadeep-jayaraman:
There are too many changes in this single PR. Can the PR be broken into smaller logical chunks so that it can be reviewed better

cjac:
yeah, there has been a lot of change.  I'll re-order the functions a little.  This should reduce the delta significantly.  Testing my change before commit and push...

cjac:
blocked on #1283

cjac (commented):
These change are ready for review.

SurajAralihalli:
Referencing a related discussion: [https://github.com/GoogleCloudDataproc/initialization-actions/pull/1269#discussion_r1904787650](https://github.com/GoogleCloudDataproc/initialization-actions/pull/1269#discussion_r1904787650).

cjac:
> blocked on #1283

1283 has been merged ; continuing.

cjac:
> /usr/bin/nvidia-smi -i 0 --multi-instance-gpu 1

This command causes problems on 11.6.  I'm reviewing nvidia-smi -h for analog for that release.

cjac:
it wants an = for the last key/value pair but the first one must not have an =.  *shrug*

cjac:
rebased on master

cjac:
okay.  I think this will do.  Kuldeep, over to you.

cjac:
I'm going to do a review of the squash and see if anything needs cleaning up.

cjac:
okay.  it's a lot, but it all looks okay and has been thoroughly and repeatedly tested. Nota Bene that 'continue' in presubmit.sh ; I'll remove that once I receive an LGTM.

cjac:
Capacity Scheduler does not support gpu resources on Dataproc 2.0, so in order to use spark with yarn, fall back to Fair Scheduler ; I know there are concurrency concerns at high load, but that concern can be managed, and there is no other option on 2.0 clusters.

cjac:
* if cuda version and gpu driver version are not specified with metadata values, but the customer supplies a cuda runfile url, extract the driver and cuda version from the filename.
* There are inconsistencies between the driver version mentioned in the filename and the actual driver version that can be installed - I have mapped these out, so that if customer uses the latest minor version of any cuda release from 11.7 through 12.6.3 we have a known good driver for that cuda version without their having to think about it.  We also have likely to work configurations from 10.0 through 11.6 if the customer is willing to run an older dataproc subminor version with equally older kernels.
* Customer can explicitly specify driver version with either gpu-driver-version or gpu-driver-url metadata value
* selecting default cuda version based on dataproc image version
* The Yarn Capacity scheduler from hadoop [3.2.4](https://hadoop.apache.org/docs/r3.2.4/) does not recognize the gpu resource type and so must be replaced with the Fair scheduler.  NB: The Fair scheduler has problems at high concurrency when underprovisioned and will time out.
* NCCL is now built from source which takes a long first time for each cuda version per dataproc image the customer deploys. Which reminds me that
* tasks which used to take a long time to perform in previous versions of this script have had their results cached so that subsequent deployments of the same configuration will take significantly less time.

kuldeepkk-dev (commented):
LGTM. 

@cjac I would recommend getting one more review from the product team since there are lot of changes.

cloudbuild/presubmit.sh
```diff
@@ -70,6 +70,7 @@ determine_tests_to_run() {
     changed_dir="${changed_dir%%/*}/"
     # Run all tests if common directories modified
     if [[ ${changed_dir} =~ ^(integration_tests|util|cloudbuild)/$ ]]; then
+      continue # to be removed before merge
```
prince-cs, line 73: this can be removed

cjac:
I met with customer to exercise the cluster and they asked for one additional change.  If the correct metadata key/value pair are passed, we now install pytorch and register a new kernel for use with JupyterLab.

gpu/install_gpu_driver.sh
```diff
@@ -1417,33 +2074,65 @@ function mount_ramdisk(){
...
+    export feature=${feature_map[$ftr]}
+    sshd_config_line="${feature} $(
+      (sshd -T | awk "/^${feature} / {print \$2}" | sed -e 's/,/\n/g';
+       ssh -Q "${ftr}" ) \
+      | sort -u | grep -v -ie sha1 -e md5 | paste -sd "," -)"
+
+    grep -iv "^${feature} " /etc/ssh/sshd_config > /tmp/sshd_config_new
+    echo "$sshd_config_line" >> /tmp/sshd_config_new
+    # TODO: test whether sshd will reload with this change before mv
+    mv /tmp/sshd_config_new /etc/ssh/sshd_config
```
cnauroth, line 2092: Update to `mv -f`.

gpu/install_gpu_driver.sh
```diff
@@ -1453,15 +2142,145 @@ function prepare_to_install(){
...
+    echo "Error: Your Spark version is not supported. Please upgrade Spark to one of the supported versions."
+    exit 1
+  fi
+
+  # Detect dataproc image version
+  if (! test -v DATAPROC_IMAGE_VERSION) ; then
+    if test -v DATAPROC_VERSION ; then
+      DATAPROC_IMAGE_VERSION="${DATAPROC_VERSION}"
+    else
+      if   version_lt "${SPARK_VERSION}" "3.2" ; then DATAPROC_IMAGE_VERSION="2.0"
```
cnauroth, line 2181: Are there any cases where neither `DATAPROC_IMAGE_VERSION` nor `DATAPROC_VERSION` is set so we need to do this detection based on Spark version? I would not have expected this.

gpu/install_gpu_driver.sh
```diff
@@ -1453,15 +2142,145 @@ function prepare_to_install(){
...
+#
+function dnf_add_repo() {
+  local -r repo_name="$1"
+  local -r repo_url="$3" # "http(s)://host/path/filename.repo"
+  local -r kr_path="${5:-/etc/pki/rpm-gpg/${repo_name}.gpg}"
+  local -r repo_path="${6:-/etc/yum.repos.d/${repo_name}.repo}"
+
+  curl -s -L "${repo_url}" \
+    | dd of="${repo_path}" status=progress
+#    | perl -p -e "s{^gpgkey=.*$}{gpgkey=file://${kr_path}}" \
```
cnauroth, line 2218: Can this commented-out code be removed?

cnauroth (changes requested):
Thanks for this, @cjac ! I entered a few comments. In general, can we also review the `curl` calls to make sure they have the necessary retries?

gpu/test_gpu.py
```diff
@@ -64,125 +152,147 @@ def verify_instance_spark(self):
   def test_install_gpu_default_agent(self, configuration, machine_suffixes,
                                      master_accelerator, worker_accelerator,
                                      driver_provider):
-    if ( self.getImageOs() == 'rocky' ) and self.getImageVersion() >= pkg_resources.parse_version("2.2"):
-      self.skipTest("GPU drivers are currently FTBFS on Rocky 9 ; base dataproc image out of date")
+    self.skipTest("No need to regularly test installing the agent on its own cluster ; this is exercised elsewhere")
```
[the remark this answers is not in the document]
cjac: I could go either way.  It feels excessive to me.  This is an expensive test that doesn't really exercise another code path.  I test with install-gpu-agent=true in my reproduction environment during development.

What do you think?

gpu/install_gpu_driver.sh
```diff
@@ -970,13 +1446,12 @@ function configure_yarn_nodemanager() {
...
-  if [[ ${spark_version} != 3.* ]]; then
-    # include exclusive mode on GPU
-    nvsmi -c EXCLUSIVE_PROCESS
-  fi
+  if [[ "${gpu_count}" == "0" ]] ; then return ; fi
+  # only run this function when spark < 3.0
+  if version_ge "${SPARK_VERSION}" "3.0" ; then return 0 ; fi
+  # include exclusive mode on GPU
+  nvsmi -c EXCLUSIVE_PROCESS
+  clear_nvsmi_cache
```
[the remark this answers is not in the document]
cjac: good catch.  I think I accidentally merged this line in from my template work.  I've got some code to cache results of nvidia-smi to XML to avoid excessively requesting the same information.  Too many executions of nvidia-smi tends to bog the system down.  I've upated the getGpusResources.sh script to cache the results, too, since the value never changes ; no GPUs are attached or removed during node lifetime.

gpu/install_gpu_driver.sh
```diff
@@ -1014,21 +1490,55 @@ function configure_gpu_script() {
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+# Example output: {"name": "gpu", "addresses":["0","1","2","3","4","5","6","7"]}
+
+set -e
+resources_json="/dev/shm/nvidia/gpusResources.json"
+if test -f "${resources_json}" ; then cat "${resources_json}" ; exit 0 ; fi
```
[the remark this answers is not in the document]
cjac: this is to cache the result of nvidia-smi.  The result does not change, and each request is a heavy operation on the GPU.  Rather than running nvidia-smi every time that yarn wants to know about GPU resources, we instead cat the file with the results of the first request.

gpu/install_gpu_driver.sh
cjac: Yes.  When running the init action scripts as custom-images input

cjac (commented):
Thank you for the review, Chris and Prince.  I will make these changes and resolve each.

gpu/test_gpu.py
```diff
@@ -6,25 +6,85 @@
...
+          continue
+        else:
+          raise
 
   def verify_instance(self, name):
     # Verify that nvidia-smi works
-    time.sleep(3) # Many failed nvidia-smi attempts have been caused by impatience
+    import random
+    # Many failed nvidia-smi attempts have been caused by impatience and temporal collisions
+    time.sleep( 3 + random.randint(1, 30) )
```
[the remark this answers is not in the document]
cnauroth: Oh, now I get it! Thanks for clarifying.

gpu/test_gpu.py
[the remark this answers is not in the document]
cnauroth: If it's exercised elsewhere, then should we remove the test entirely instead of skipping it?

gpu/install_gpu_driver.sh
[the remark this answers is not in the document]
cnauroth: Got it, I missed the broader context that this is ultimately executed inside the GPU discovery script, which we are generating here.

gpu/install_gpu_driver.sh
cnauroth: I hadn't thought of that use case. Can you please add a comment that the final else path is meant to support custom image builds?

gpu/test_gpu.py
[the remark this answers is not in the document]
cjac: that seems reasonable

gpu/install_gpu_driver.sh
cjac: sure thing

gpu/install_gpu_driver.sh
cjac: https://issuetracker.google.com/issues/393219117

gpu/install_gpu_driver.sh
```diff
@@ -668,47 +922,36 @@ function add_nonfree_components() {
...
+
+  os_add_repo nvidia-container-toolkit \
+              "${signing_key_url}" \
+              "${repo_data}" \
+              "no"
 }
 
 function add_repo_cuda() {
   if is_debuntu ; then
-    local kr_path=/usr/share/keyrings/cuda-archive-keyring.gpg
```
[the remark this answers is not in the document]
cjac: I did patch it back in.

cjac:
@cnauroth - I believe that I addressed all of your concerns in the previous change, https://github.com/GoogleCloudDataproc/initialization-actions/pull/1275/commits/eacb99f5d57ea4a3a993a181747d87071ca73108

Please let me know if you see anything that needs further attention.

gpu/install_gpu_driver.sh
```diff
@@ -134,13 +134,13 @@ readonly ROLE
 
 # Minimum supported version for open kernel driver is 515.43.04
 # https://github.com/NVIDIA/open-gpu-kernel-modules/tags
-latest="$(curl -s https://us.download.nvidia.com/XFree86/Linux-x86_64/latest.txt | awk '{print $1}')"
+#latest="$(curl -s https://us.download.nvidia.com/XFree86/Linux-x86_64/latest.txt | awk '{print $1}')"
```
cnauroth, line 137: Remove this line if `latest` is no longer being used?

gpu/install_gpu_driver.sh
```diff
@@ -625,11 +627,22 @@ function install_nvidia_nccl() {
     local gcs_tarball="${pkg_bucket}/nvidia/nccl/${_shortname}/${build_tarball}"
 
     if [[ "$(hostname -s)" =~ ^test && "$(nproc)" < 32 ]] ; then
-      # do not build in tests with < 32 cores
+      # when running with fewer than 32 cores, yield to in-progress build
       sleep $(( ( RANDOM % 11 ) + 10 ))
-      while gsutil ls "${gcs_tarball}.building" 2>&1 | grep -q "${gcs_tarball}.building" ; do
-        sleep 5m
-      done
+      if gsutil ls -j "${gcs_tarball}.building" > "${local_tarball}.building.json" ; then
```
cnauroth, line 632: Is `-j` a recent new feature? I couldn't find this option in my `gsutil`.

cjac: it dumps the object metadata in JSON format ; I thought it would be `--format json` to match other gcloud commands, but I guess that it's slightly different because of gsutil.  The argument is documented for `gcloud storage ls --help` which I assumed used the same ABI as gsutil, but I don't see the argument in the `gcloud ls --help` documentation.

cjac: as for whether it's new, I don't think it is.  I looked through the release history[1], and it seems that JSON support has been included from the beginning.  But I couldn't find the argument parsing code where the argument is detected on a quick review of [2].

[1] https://github.com/GoogleCloudPlatform/gsutil/blob/master/CHANGES.md
[2] https://github.com/GoogleCloudPlatform/gsutil

cjac:
2.2.44-rocky9 is able to build the kernel driver, so I've removed the test skip for rocky9.

cnauroth (approved):
@cjac , that addresses all of my feedback. Thank you very much!

LGTM, pending a successful CI run on the latest changes.

@prince-cs and @kuldeepkk-dev , please let us know if you had any other feedback.

PR #1293, opened 2025-01-15
Add script and readme for hive lineage

codelixir:
This initialization action configures Google Cloud Dataproc to enable lineage for Hive jobs.

Lineage for Hive is currently not supported, and is not configurable through cluster properties as of now. We are going preview with this feature through initialization actions.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -138,14 +138,16 @@ py_library(
 
 py_test(
     name = "test_hive_lineage",
-    size = "small",
+    size = "enormous",
     srcs = ["hive-lineage/test_hive_lineage.py"],
     data = [
         "hive-lineage/hive-lineage.sh",
         "hive-lineage/hivetest.hive",
     ],
     local = True,
+    shard_count = 3,
     deps = [
         "//integration_tests:dataproc_test_case",
+        "@io_abseil_py//absl/testing:parameterized",
     ],
 )
--- a/BUILD
+++ b/BUILD
@@ -6,8 +6,8 @@ test_suite(
         ":test_cloud_sql_proxy",
         ":test_dr_elephant",
         ":test_hive_hcatalog",
-        ":test_starburst_presto",
         ":test_spark_rapids",
+        ":test_starburst_presto",
         "//alluxio:test_alluxio",
         "//atlas:test_atlas",
         "//bigtable:test_bigtable",
@@ -60,7 +60,10 @@ py_test(
     name = "test_cloud_sql_proxy",
     size = "enormous",
     srcs = ["cloud-sql-proxy/test_cloud_sql_proxy.py"],
-    data = ["cloud-sql-proxy/cloud-sql-proxy.sh", "cloud-sql-proxy/hivetest.hive"],
+    data = [
+        "cloud-sql-proxy/cloud-sql-proxy.sh",
+        "cloud-sql-proxy/hivetest.hive",
+    ],
     local = True,
     shard_count = 4,
     deps = [
@@ -114,10 +117,10 @@ py_test(
     size = "enormous",
     srcs = ["spark-rapids/test_spark_rapids.py"],
     data = [
+        "spark-rapids/mig.sh",
         "spark-rapids/spark-rapids.sh",
         "spark-rapids/verify_xgboost_spark_rapids.scala",
         "spark-rapids/verify_xgboost_spark_rapids_sql.scala",
-        "spark-rapids/mig.sh",
     ],
     local = True,
     shard_count = 3,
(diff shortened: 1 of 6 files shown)


hive-lineage/README.md
```diff
@@ -0,0 +1,36 @@
+# Hive Lineage Initialization Action
```
cjac, line 1: our build system doesn't much care for hyphens in directory names.  You'll have to jump through some hoops.  Take a look at /BUILD for the spark-rapids example

hive-lineage/hive-lineage.sh
```diff
@@ -0,0 +1,52 @@
...
+#   This initialization script installs the required
+#   jars and sets the hive conf to enable lineage.
+#
+#   To use this script, add the following using the
+#   --initialization-actions flag during cluster creation.
+#   gs://dataproc-hive-lineage-prototype/v3/initialization-actions/enable_lineage.sh
+
+set -euxo pipefail
+
+export HIVE_HOME="/usr/lib/hive"
```
cjac, line 12: can we move environment variable definition to a prepare function called after all function definitions?  It will make the refactor to templates smoother.

cjac (commented):
It's much shorter than I expected.  Has this been exercised on all variants?  ubuntu, debian, rocky by 2.0, 2.1, 2.2 ?  Do we know that some variants are unsupported?

hive-lineage/README.md
codelixir: Thanks, I referred to it to make changes to the BUILD file

hive-lineage/hive-lineage.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+#   This initialization script installs the required
```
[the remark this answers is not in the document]
codelixir: Done, thanks

hive-lineage/hive-lineage.sh
codelixir: Done

codelixir:
> It's much shorter than I expected. Has this been exercised on all variants? ubuntu, debian, rocky by 2.0, 2.1, 2.2 ? Do we know that some variants are unsupported?

I think it should work on all variants. Will test it out with different combinations and post here

hive-lineage/hive-lineage.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+#
+# Copyright 2015 Google LLC and contributors
```
Deependra-Patel, line 3: Nit: 2025

hive-lineage/hive-lineage.sh
```diff
@@ -0,0 +1,64 @@
...
+}
+
+function install_jars() {
+  echo "Installing openlineage-hive hook"
+  gsutil cp -P "$INSTALLATION_SOURCE/hive-openlineage-hook-$HIVE_HIVE_OL_HOOK_VERSION.jar" "$HIVE_LIB_DIR/hive-openlineage-hook.jar"
+}
+
+function restart_hive_server2_master() {
+  ROLE=$(curl -f -s -H Metadata-Flavor:Google http://metadata/computeMetadata/v1/instance/attributes/dataproc-role)
+  if [[ "${ROLE}" == 'Master' ]]; then
```
Deependra-Patel, line 55: Do also test with ha clusters (with multiple masters). It should still work though but just to be sure.

hive-lineage/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+You can use this initialization action to create a new Dataproc cluster with Lineage enabled for Hive jobs.
+Note that this feature is in preview for now.
+
+```shell 
+REGION=<region>
+CLUSTER_NAME=<cluster_name>
+gcloud dataproc clusters create ${CLUSTER_NAME} \
+  --region ${REGION} \
+  --scopes cloud-platform \
```
Deependra-Patel, line 17: Is scopes needed?

Deependra-Patel: Discussed offline. It is indeed needed for lineage.

hive-lineage/hive-lineage.sh
codelixir: oops, thanks for pointing out. fixed.

hive-lineage/hive-lineage.sh
codelixir: added 'HA' as a test parameter

BUILD
```diff
@@ -138,7 +138,7 @@ py_library(
 
 py_test(
```
Deependra-Patel, line 139: We should set shard_count to maybe 3? It runs tests in parallel

hive-lineage/test_hive_lineage.py
```diff
@@ -13,8 +15,15 @@ def __submit_hive_job(self, cluster_name):
...
 
+  @parameterized.parameters(
+      'STANDARD',
+      'HA',
+  )
   def test_hive_job_success(self, configuration):
-    self.createCluster(configuration, self.INIT_ACTIONS, scopes='cloud-platform')
+    self.createCluster(configuration,
+                       self.INIT_ACTIONS,
+                       timeout_in_minutes=30,
```
Deependra-Patel, line 25: Does it really need 30 mins? I think we should set with how long the test takes + 5 mins as buffer. Setting too much value here, doesn't allow us to fail fast.

codelixir: Earlier the test was failing because of timeout, hence I had increased it to 30.

codelixir: I removed it and tests still pass
(I think changing test size to "enormous" does the trick; otherwise it fails)

Deependra-Patel (approved):

Issue #1294, opened 2025-02-04
GPU utilization in spark error


m15pradeep:
I'm using attached install_gpu_driver.sh in dataproc 2.2. GPU is not getting recognized in spark. Attached installation logs for reference

[dataproc-gpu-main.txt](https://github.com/user-attachments/files/18652635/dataproc-gpu-main.txt)
[dataproc-initialization-script-0.log](https://github.com/user-attachments/files/18652636/dataproc-initialization-script-0.log)
[install_gpu_driver.txt](https://github.com/user-attachments/files/18652634/install_gpu_driver.txt)

Command:
**Library: tensorflow[and-cuda]
import tensorflow as tf
print(tf.config.list_physical_devices('CPU'))
print(tf.config.list_physical_devices('GPU'))**

Log:
2025-02-04 06:04:29.413125: I tensorflow/core/util/port.cc:153] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2025-02-04 06:04:31.327492: E external/local_xla/xla/stream_executor/cuda/cuda_fft.cc:477] Unable to register cuFFT factory: Attempting to register factory for plugin cuFFT when one has already been registered
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
E0000 00:00:1738649071.978220 80542 cuda_dnn.cc:8310] Unable to register cuDNN factory: Attempting to register factory for plugin cuDNN when one has already been registered
E0000 00:00:1738649072.400870 80542 cuda_blas.cc:1418] Unable to register cuBLAS factory: Attempting to register factory for plugin cuBLAS when one has already been registered
2025-02-04 06:04:36.546724: I tensorflow/core/platform/cpu_feature_guard.cc:210] This TensorFlow binary is optimized to use available CPU instructions in performance-critical operations.
To enable the following instructions: AVX2 AVX512F AVX512_VNNI FMA, in other operations, rebuild TensorFlow with the appropriate compiler flags.
2025-02-04 06:04:50.776297: E external/local_xla/xla/stream_executor/cuda/cuda_driver.cc:152] failed call to cuInit: INTERNAL: CUDA error: Failed call to cuInit: CUDA_ERROR_UNKNOWN: unknown error
[...]

PR #1302, opened 2025-02-15
[gpu] Exercise installer with all supported GPU types

### LINKED ISSUE #1299: [gpu] driver installer has not been exercised with P4, P100, or V100 GPUs in some time
The installer should be exercised with these GPUs

The installer should report that the dataproc image on which it is running is not capable of building the driver if it is not

The installer should be updated to correct any problems with the installation process

cjac:
* enabling GPU agent by default
* for cuda repo, add known signing keys to trustdb
* move variable definition to prepare function
* add some TODO notes for installing this script as a custom image
* spark supports maximum 1 GPU per executor
* printing warning when old cuda is used with ubuntu22
* built a list of PCI devices for each supported GPU type
* improvements to disk usage metrics display
* Supported on 2.0.27-debian10
* * added fallback to using gsutil when gcloud sdk < 402
* * adoptium, cran-r and gcloud pgp databases updated
* * writing multi-valued gpg databases using `gpg --import --no-default-keyring --keyring ...` vice `gpg --dearmor -o`
* * evaluating variable name for keyring path in sources.list file
* driver builds for P4, P100, and V100 now tested
* * build driver from proprietary source when PCI Device ID < 0x1E00 (Turing)
* * separating open vs non-free kmod builds into their own gcs path
* renamed completion variable for nvidia container toolkit from nvtk to nvctk
* removed export of METADATA_HTTP_PROXY variable, since it's only used in this one function

Fixes #1299 , #1294

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -449,7 +449,7 @@ GPU_DRIVER_PROVIDER=$(get_metadata_attribute 'gpu-driver-provider' 'NVIDIA')
 readonly GPU_DRIVER_PROVIDER
 
 # Whether to install GPU monitoring agent that sends GPU metrics to Stackdriver
-INSTALL_GPU_AGENT=$(get_metadata_attribute 'install-gpu-agent' 'false')
+INSTALL_GPU_AGENT=$(get_metadata_attribute 'install-gpu-agent' 'true')
 readonly INSTALL_GPU_AGENT
 
 # Dataproc configurations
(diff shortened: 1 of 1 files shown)


cjac:
image

@prince-cs - can you help me to exercise this code using the internal tests before merging to master?

Or will it be less hassle to merge it to master, push it to GCS and then run the internal tests?

prince-cs:
Would prefer merging first as it would be easier to let automation take care of the flow.

cjac:
then let's get 'er done

prince-cs (commented):
LGTM

GHWendy (approved):

PR #1304, opened 2025-02-23
Fix the script to fix two scenarios:

xby-G:
	2. Multiple source list files need to be updated


cjac:
Looks good from the surface.  Do you have successful build logs that you can submit instead of a gcbrun?  I think you are allowed to run that.

cjac (approved):
LGTM! Thank you for your contribution!

PR #1306, opened 2025-02-26
[hive-lineage] Update hive-lineage.sh

codelixir:
This change appends hive confwhitelist in hive-lineage.sh to allow users to set custom hive openlineage properties at runtime.

In the previous version of the script, workloads were failing for kerberos-enabled clusters when setting `hive.openlineage.namespace` in job properties. This change fixes the issue.


Deependra-Patel:
Can we add a test for this?

Deependra-Patel (approved):

PR #1310, opened 2025-02-28
Fix user-environment.sh

xby-G:
Fix the script to fix two scenarios  
                2. Multiple source list files need to be updated

### DIFF
--- a/user-environment/user-environment.sh
+++ b/user-environment/user-environment.sh
@@ -17,6 +17,37 @@ set -o errexit
 set -o nounset
 set -o xtrace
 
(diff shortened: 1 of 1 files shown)


xby-G:
It seems there is no validation script

cjac:
I have an issue open internally to discuss whether we want to drop this action.  I have received no customer use reports on this one.

What do you think, @xby-G and @kuldeepkk-dev ?

xby-G:
@cjac 

Acutally I felt the same way. How do you think we can proceed? I think we can rename it as "deprecated" for now.

kuldeepkk-dev:
+1 
I haven't seen any customer use case on this init action. 
@cjac We can merge the changes and mark this as deprecated?

PR #1314, opened 2025-04-03
Removing passwords from environment variable assignment

cjac:
Adding secret manager as a password specification source


cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -209,13 +209,19 @@ fi
 readonly CLOUDSQL_INSTANCE_TYPE
 
 METASTORE_PROXY_PORT="$(/usr/share/google/get_metadata_value attributes/metastore-proxy-port || echo '')"
-if [[ "${METASTORE_INSTANCE}" =~ =tcp:[0-9]+$ ]]; then
-  METASTORE_PROXY_PORT="${METASTORE_INSTANCE##*:}"
-else
-  METASTORE_PROXY_PORT=${DEFAULT_DB_PORT["${CLOUDSQL_INSTANCE_TYPE}"]}
+if [[ -z "${METASTORE_PROXY_PORT}" ]] ; then
```
cjac, line 212: don't overwrite the value that was collected from the metadata attribute ; only attempt to set this value if the attrbiute was not set.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -233,39 +239,46 @@ readonly KMS_KEY_URI
 DB_ADMIN_PASSWORD_URI="$(/usr/share/google/get_metadata_value attributes/db-admin-password-uri || echo '')"
 readonly DB_ADMIN_PASSWORD_URI
 
-DB_ADMIN_PASSWORD=''
```
cjac, line 236: remove DB_ADMIN_PASSWORD environment variable entirely.  Do not store sensitive data in environment variables.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -233,39 +239,46 @@ readonly KMS_KEY_URI
...
-if [[ -n "${DB_ADMIN_PASSWORD_URI}" ]]; then
+if [[ -n "${DB_ADMIN_SECRET}" ]] ; then
+  gcloud secrets versions access "${DB_ADMIN_SECRET#*:}" \
+      --project="${METASTORE_INSTANCE%%:*}" \
+      --secret="${DB_ADMIN_SECRET%:*}" > /dev/shm/db-pw
```
cjac, line 245: Instead of storing classified material in environment variables, store to ephemeral storage, and ensure it is purged on exit of script.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -233,39 +239,46 @@ readonly KMS_KEY_URI
...
     gcloud kms decrypt \
       --ciphertext-file - \
       --plaintext-file - \
-      --key "${KMS_KEY_URI}")"
+      --key "${KMS_KEY_URI}" > /dev/shm/db-pw
+else
+  touch /dev/shm/db-pw
 fi
-if [[ "${CLOUDSQL_INSTANCE_TYPE}" == "POSTGRES" && -z "${DB_ADMIN_PASSWORD}" ]]; then
+if [[ "${CLOUDSQL_INSTANCE_TYPE}" == "POSTGRES" && -z "$(perl -pe 'chomp' < /dev/shm/db-pw)" ]]; then
```
cjac, line 256: this test could instead be done with `wc -c` to eliminate the final instance of the variable being stored in bash strings

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -233,39 +239,46 @@ readonly KMS_KEY_URI
...
+  gcloud secrets versions access "${DB_HIVE_SECRET#*:}" \
+      --project="${METASTORE_INSTANCE%%:*}" \
+      --secret="${DB_HIVE_SECRET%:*}" > /dev/shm/hive-pw
+elif [[ -n "${DB_HIVE_PASSWORD_URI}" ]]; then
   # Decrypt password
-  DB_HIVE_PASSWORD="$(gsutil cat "${DB_HIVE_PASSWORD_URI}" |
+  gsutil cat "${DB_HIVE_PASSWORD_URI}" |
     gcloud kms decrypt \
       --ciphertext-file - \
       --plaintext-file - \
```
cjac, line 272: `--plaintext-file /dev/shm/hive-pw`

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -394,13 +407,6 @@ function install_cloud_sql_proxy() {
   local proxy_flags
   proxy_flags="$(get_proxy_flags)"
 
-  # Validate db_hive_password and escape invalid xml characters if found.
```
cjac, line 397: these lines are duplicated in the single perl line below

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -443,12 +449,12 @@ EOF
...
   <property>
     <name>javax.jdo.option.ConnectionPassword</name>
-    <value>${db_hive_password_xml_escaped}</value>
+    <value>$(perl -pe 'chomp ; s:<:&lt;:g; s:>:&gt;:g ; s:":&quot;:g' < /dev/shm/hive-pw)</value>
   </property>
 </configuration>
 EOF
 
-    bdconfig merge_configurations \
+    /usr/local/bin/bdconfig merge_configurations \
```
cjac, line 457: rocky doesn't have `/usr/local/bin` in its PATH

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -458,36 +464,57 @@ EOF
...
+  rm -f /dev/shm/*-pw /dev/shm/*-db.cnf /dev/shm/*.sql
+}
+
+trap exit_handler EXIT
+
 function initialize_postgres_metastore_db() {
   log 'Initialzing POSTGRES DB for Hive metastore ...'
-  local admin_connection=postgresql://"${DB_ADMIN_USER}":"${DB_ADMIN_PASSWORD}"@127.0.0.1:"${METASTORE_PROXY_PORT}"/
-  local hive_connection=postgresql://"${DB_HIVE_USER}":"${DB_HIVE_PASSWORD}"@127.0.0.1:"${METASTORE_PROXY_PORT}"/postgres
+  local admin_connection=postgresql://"${DB_ADMIN_USER}":"$(perl -pe 'chomp' < /dev/shm/db-pw)"@127.0.0.1:"${METASTORE_PROXY_PORT}"/
```
cjac, line 516: I was mistaken, after the `wc` change above, these two lines will be the final instance of assignment of classified material to bash environment variable.  I'm not certain how best to address this problem, but I will stand up a PG instance to see if there's a way to specify DSN components from a config file as is done for mysql with `--defaults-file=${admin_defaults_file}`

cjac (commented):
Some notes about why the changes are being made, and thoughts about how to improve.

cloud-sql-proxy/cloud-sql-proxy.sh
```diff
@@ -233,39 +237,46 @@ readonly KMS_KEY_URI
...
       --ciphertext-file - \
-      --plaintext-file - \
-      --key "${KMS_KEY_URI}")"
+      --plaintext-file /dev/shm/db-pw \
+      --key "${KMS_KEY_URI}"
+else
+  touch /dev/shm/db-pw
 fi
-if [[ "${CLOUDSQL_INSTANCE_TYPE}" == "POSTGRES" && -z "${DB_ADMIN_PASSWORD}" ]]; then
+if [[ "${CLOUDSQL_INSTANCE_TYPE}" == "POSTGRES" ]] && [[ "$(perl -pe 'chomp' < /dev/shm/db-pw | wc -c)" != "0" ]]; then
```
TheAkertu: What does the perl validation do in this context?

cjac: Thanks for asking, Axel!

The < directs to perl's STDIN file handle the contents of the named file 

`perl -e` instructs Perl to evaluate the value of the -e argument instead of loading and evaluating a named file.  For instance, the following instructs Perl to print hello world (followed by a newline character [$/]) to STDOUT: `perl -e 'print "hello world$/"'`

`perl -pe` instructs Perl to act as a filter, printing to STDOUT, one line at a time, anything that it receives on STDIN, after evaluating the code in the argument to `-e`.  A string containing the line being processed is assigned to the default variable, `$_` before the -e code is evaluated, and `$_` is printed to STDOUT after the code is evaluated.

The Perl (`chomp`)[https://perldoc.perl.org/functions/chomp] function strips a trailing newline (`$/`) from a string variable named as chomp's first argument.  If no argument is passed, strips trailing newline from the default variable, `$_`.  If no trailing newline exists, `chomp` is a null operation.

So in toto, this command reads the file, strips the newline, and compares the result to an empty string.  The functionality of the Perl code is the same as the bash code, but does not expose the secret to the log when bash is executed with the -x option.

PR #1315, opened 2025-04-04
[gpu] Revert experiment with debug log reduction

cjac:
Validate and correct invalid values for executor_cores and task_cpus

re-include tests for running the script with no GPU hardware


TheAkertu (approved):
LGTM

PR #1320, opened 2025-05-04
[gpu] Defer GPU config on custom images to resolve #1303

### LINKED ISSUE #1303: [custom-images] install_gpu_driver.sh does not configure spark when run as startup script with custom-images #110
##Problem

Although the gpu driver installer completes successfully when used as an image customization script, it does not create or alter the spark defaults file to include gpu parameters. This is because at the time of the run of the startup script, bdutil has not yet created the files.

See configure_yarn_resources[1] for example

##Proposed solution:

Detect whether the script is being run as a image customization. If custom image environment not detected, run as normal; if custom image environment is detected, delay execution by installing a service to run the configuration update phase on first boot.

Scripts to be run after first boot can be placed into a directory such as /etc/rc.local.d/

An entry could be added to /etc/rc.local to

* check for the existence of files in this directory
* execute them
* upon successful execution, remove them

[1] https://github.com/GoogleCloudDataproc/initialization-actions/blob/7e875225563b21680623ffc8bd04d8888aaef626/gpu/install_gpu_driver.sh#L1436

cjac:
This pull request refactors the `install_gpu_driver.sh` script, primarily to **resolve the issue of Spark and Hadoop configurations failing during custom image creation, as detailed in [GitHub Issue #1303](https://github.com/GoogleCloudDataproc/initialization-actions/issues/1303).**

The core problem was that the script attempted to modify configuration files (like `spark-defaults.conf`) before they were created by `bdutil` during the image customization process. This PR implements the proposed solution by deferring these configuration steps until the first boot of the instance.

* **Deferred Configuration for Custom Images:**
    * The script now detects if it's running in a custom image build context by checking the `invocation-type` metadata attribute. This is stored in the `IS_CUSTOM_IMAGE_BUILD` variable.
    * When `IS_CUSTOM_IMAGE_BUILD` is true, critical Hadoop and Spark configuration steps are no longer executed immediately. Instead, a new systemd service (`dataproc-gpu-config.service`) is generated and enabled.
    * This service is responsible for running a newly created script (`/usr/local/sbin/apply-dataproc-gpu-config.sh`) on the instance's first boot. This generated script now contains all the necessary logic for Hadoop/Spark/GPU configuration (moved into a `run_hadoop_spark_config` function).
    * This deferral mechanism **explicitly solves issue #1303** by ensuring that configurations are applied only after the Dataproc environment, including necessary configuration files, has been fully initialized.
* **Script Structure for Deferred Execution:**
    * The `main` function has been refactored. It now orchestrates the installation of drivers and core components as before. However, for Hadoop/Spark configurations, it either executes the `apply-dataproc-gpu-config.sh` script directly (if not a custom image build) or enables the systemd service to run it on first boot.
    * The `create_deferred_config_files` function is responsible for generating the systemd service unit and the `apply-dataproc-gpu-config.sh` script. This script is carefully constructed to include all necessary helper functions and variables from the main `install_gpu_driver.sh` script to run independently.
* **Re-evaluation of Environment in Deferred Script:** The deferred script (`apply-dataproc-gpu-config.sh`) re-evaluates critical environment variables like `ROLE`, `SPARK_VERSION`, `gpu_count`, and `IS_MIG_ENABLED` at the time of its execution (first boot) to ensure accuracy.

* **CUDA and Driver Updates:**
    * Added support for Dataproc image version "2.3", defaulting to CUDA version "12.6.3".
    * Improved robustness in `install_build_dependencies` for Rocky Linux with fallbacks for kernel package downloads.
* **Error Handling and Robustness:**
    * Several commands, like `gsutil rm`, `pip cache purge`, and `wget` in `Workspace_mig_scripts`, have improved error handling or are wrapped in `execute_with_retries`.
    * Suppressed benign errors from `du` commands during cleanup.
    * Zeroing of free disk space is now more robust and conditional on custom image builds.
* **Configuration and Installation Improvements:**
    * Dynamically sets `conda_root_path` based on `DATAPROC_IMAGE_VERSION`.
    * Corrected GPG key handling for the NVIDIA Container Toolkit repository on Debian systems.
    * Ensures `python3-venv` is installed for the GPU agent on newer Debian-based images.
    * Streamlined several configuration functions by removing redundant GPU count checks.
    * Ensures RAPIDS properties are added to `spark-defaults.conf` idempotently.
    * The `check_secure_boot` function now handles cases where `mokutil` might not be present and provides a clearer error for missing signing material.
    * The script entry point and preparation steps (`prepare_to_install`) are more clearly defined.

By implementing a deferred configuration mechanism for custom image builds, this pull request directly addresses and **resolves the core problem outlined in GitHub issue #1303**, ensuring that GPU-related Hadoop and Spark configurations are applied reliably.

### DIFF
--- a/gpu/install_gpu_driver.sh
+++ b/gpu/install_gpu_driver.sh
@@ -108,14 +108,15 @@ function get_metadata_value() {
     print_metadata_value_if_exists ${MDS_PREFIX}/project/${varname}
     return_code=$?
   fi
-
   return ${return_code}
 }
 
 function get_metadata_attribute() {
   local -r attribute_name="$1"
   local -r default_value="${2:-}"
+  set +e
   get_metadata_value "attributes/${attribute_name}" || echo -n "${default_value}"
+  set -e
 }
 
 OS_NAME="$(lsb_release -is | tr '[:upper:]' '[:lower:]')"
@@ -186,6 +187,7 @@ function set_cuda_version() {
     "2.0" ) DEFAULT_CUDA_VERSION="12.1.1" ;; # Cuda 12.1.1 - Driver v530.30.02 is the latest version supported by Ubuntu 18)
     "2.1" ) DEFAULT_CUDA_VERSION="12.4.1" ;;
     "2.2" ) DEFAULT_CUDA_VERSION="12.6.3" ;;
+    "2.3" ) DEFAULT_CUDA_VERSION="12.6.3" ;;
     *   )
       echo "unrecognized Dataproc image version: ${DATAPROC_IMAGE_VERSION}"
       exit 1
@@ -462,6 +464,8 @@ readonly SPARK_CONF_DIR='/etc/spark/conf'
 NVIDIA_SMI_PATH='/usr/bin'
 MIG_MAJOR_CAPS=0
 IS_MIG_ENABLED=0
+
+IS_CUSTOM_IMAGE_BUILD="false" # Default
 
 function execute_with_retries() (
   local -r cmd="$*"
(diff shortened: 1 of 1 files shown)


cjac:
merged into #1363

PR #1322, opened 2025-05-21
Set default task.resource.gpu.amount to a number much smaller than possible cores

abellina:
We would like to set the default `spark.task.resource.gpu.amount` to a very small number, effectively removing this as part of the equation when Spark is deciding how many tasks can run in parallel in this executor. In the past, we have been setting this number to 1/# of cores, and we could do that here if we knew the number of cores that workers are assigned. That said, we think choosing a small number is simpler.

Note that this can still be overwritten by the user. Note also that the current default of 0.25 is not great, as we tend to run with 16 core executors.

@viadea @tgravescs @SurajAralihalli fyi

### DIFF
--- a/spark-rapids/spark-rapids.sh
+++ b/spark-rapids/spark-rapids.sh
@@ -299,8 +299,12 @@ spark.executor.resource.gpu.discoveryScript=/usr/lib/spark/scripts/gpu/getGpusRe
 spark.dynamicAllocation.enabled=false
 spark.sql.autoBroadcastJoinThreshold=10m
 spark.sql.files.maxPartitionBytes=512m
-# please update this config according to your application
-spark.task.resource.gpu.amount=0.25
+# For Spark SQL, we want the scheduler to use the number of CPU cores as the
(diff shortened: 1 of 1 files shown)


viadea:
@jayadeep-jayaraman FYI

tgravescs (approved):

jayadeep-jayaraman (commented):
@abellina - can setting this to a small number have any adverse impact like OOM if too many tasks start running concurrently on the GPU utilizing all of the GPU memory ?

jayadeep-jayaraman (approved):

abellina:
Sorry @jayadeep-jayaraman for the delay. The small number only ensures that the GPU is not the limiting factor when scheduling, that instead we use all spark.executor.cores the user intended. If users want to have an artificially smaller number of tasks run only when GPU are handed to a stage (via resource profiles for example) they should worry about this, but the fact is that the vast majority of users really just want to match the # of cores given in spark.executor.cores, so this other setting is just confusing and in the way.

PR #1331, opened 2025-07-01
Update presubmit.sh to change the default branch from master to main

jayadeep-jayaraman:


jayadeep-jayaraman:
This is a duplicate of PR #1330

PR #1333, opened 2025-07-29
Update the incompatibility notice to alert users not to use CloudSQL …

dilipgodhia:
…Proxy (V1) with CloudSQL MySQL 8.4.


cjac (approved):
LGTM

PR #1359, opened 2025-09-24
[spark-rapids.sh] Refactor NVIDIA driver installation for Rocky Linux to use run file

SurajAralihalli:
Updated the installation process for the NVIDIA GPU driver on Rocky Linux. The script now installs kernel development packages directly and downloads the CUDA installer run file, executing it in silent mode. The installer file is removed post-installation to clean up.

### DIFF
--- a/spark-rapids/spark-rapids.sh
+++ b/spark-rapids/spark-rapids.sh
@@ -501,17 +501,20 @@ function install_nvidia_gpu_driver() {
 
   elif is_rocky ; then
 
-    # Ensure the Correct Kernel Development Packages are Installed
-    execute_with_retries "dnf -y -q update --exclude=systemd*,kernel*"
-    execute_with_retries "dnf -y -q install pciutils kernel-devel gcc"
+    # Install kernel development packages
+    execute_with_retries "dnf install -y kernel-devel-$(uname -r) kernel-headers-$(uname -r)"
 
-    readonly NVIDIA_ROCKY_REPO_URL="${NVIDIA_REPO_URL}/cuda-${shortname}.repo"
-    execute_with_retries "dnf config-manager --add-repo ${NVIDIA_ROCKY_REPO_URL}"
-    execute_with_retries "dnf clean all"
-    configure_dkms_certs
-    execute_with_retries "dnf -y -q module install nvidia-driver:latest-dkms"
-    clear_dkms_key
-    execute_with_retries "dnf -y -q install cuda-toolkit"
+    # Download the CUDA installer run file
+    curl -o driver.run \
+        "https://developer.download.nvidia.com/compute/cuda/${CUDA_VERSION}/local_installers/cuda_${CUDA_VERSION}_${NVIDIA_DRIVER_VERSION}_linux.run"
+    
+    # Run the installer in silent mode
+    bash driver.run --silent
+
+    # Remove the installer file after installation to clean up
+    rm driver.run
+
+    # Load the NVIDIA kernel module
     modprobe nvidia
 
   else


cjac:
gcbrun

SurajAralihalli:
Thanks @cjac, is this PR ready to be merged?

cjac:
It appears that the rocky 8 tests were disabled.  I'm going to re-enable them and test your change in my dev environment.

cjac:
I found the problem with rocky9.  It was the one I thought might be to blame.  I'm exercising the change on rocky 9 and if it resolves the issue in my repro environment, I'll commit and push.

cjac:
in addition to the move of the kernel packages to the vault, 2.4.1 will not compile against the kernel in Rocky 9.  I'm not sure whether you want to:

* disable support on the 2.0 images
* disable support on the 2.1-Rocky8 images
* select the CUDA and driver versions based on the dataproc version

My recommendation would be to use gpu/install_gpu_driver.sh instead, since it already handles all of these issues, and it would reduce the maintenance load.

SurajAralihalli:
> * select the CUDA and driver versions based on the dataproc version

Can you please explain how to choose the appropriate CUDA and driver versions depending on the Dataproc version being used?

cjac:
> > * select the CUDA and driver versions based on the dataproc version
> 
> Can you please explain how to choose the appropriate CUDA and driver versions depending on the Dataproc version being used?
> 
> > My recommendation would be to use gpu/install_gpu_driver.sh instead, since it already handles all of these issues, and it would reduce the maintenance load.
> 
> I agree, but the MIG issue is blocking us from doing that [#1269 (comment)](https://github.com/GoogleCloudDataproc/initialization-actions/pull/1269#issuecomment-2611549158)
> 
> Update: If MIG support is disabled, can we invoke `gpu/install_gpu_driver.sh` directly from `spark-rapids.sh` so users only need a single init script, instead of two separate ones? I also want to avoid duplicating the logic from `gpu/install_gpu_driver.sh` into `spark-rapids.sh`, which could otherwise lead to inconsistencies.

You're right.  I need to make it reboot safe.  What do you think about changing the test suite such that instead of a fire-and-forget approach to the initialization action, we start a screen session which executes the main script and installs a "nanny" systemd service to poll every N seconds.  If it finds that the success criteria have not been met, but there is not presently a screen session running the installer, it runs the script in a screen session.  That way, the process becomes immune from some environmental factors which might terminate its connection to the controlling terminal of the parent process.

It would also mean that the script log would not always be printing to /var/log/dataproc-initialization-script-0.log ; this might break the ABI of some use cases.  Perhaps when clusters which have not completed initialization re-run their startup script on boot.  I will investigate and let you know as I know more.

cjac (approved):
this is sufficient to get 2.1-rocky8 up and running.

For 2.2-rocky9, we should build from github.

SurajAralihalli:
> This change makes 2.1-rocky8 functional. 2.2-rocky9 will take quite a bit more effort, and I think it would be better to resolve the issues with install_gpu_driver.sh instead of putting that effort into a script we intend to replace moving forward.

Thanks @cjac for working on this! I’m open to switching `spark-rapids.sh` to the `install_gpu_driver.sh` approach. Is it possible to read and invoke the existing `install_gpu_driver.sh`  from within `spark-rapids.sh`. I’d like to avoid duplicating the CUDA/driver installation logic in `spark-rapids.sh`, since that create maintenance challenges? It would be easier for the user to work with a single init script rather than having separate ones for Spark and for CUDA/driver installation.

I'm happy to create an issue/ticket for us to track.

PR #1371, opened 2025-12-26
Handle missing dataproc-role metadata in hive-lineage.sh

codelixir:
Replaces the `curl -f` call to the metadata server with the safer `/usr/share/google/get_metadata_value` utility to retrieve the `dataproc-role` attribute.

This change ensures the initialization action completes successfully even if the metadata key is not yet present, by using the standard Dataproc metadata helper and providing a fallback.

### DIFF
--- a/hive-lineage/hive-lineage.sh
+++ b/hive-lineage/hive-lineage.sh
@@ -52,7 +52,8 @@ function install_jars() {
 }
 
 function restart_hive_server2_master() {
-  ROLE=$(curl -f -s -H Metadata-Flavor:Google http://metadata/computeMetadata/v1/instance/attributes/dataproc-role)
+  # Safely get metadata without failing the script if the key is missing
+  ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role || echo "Unknown")
   if [[ "${ROLE}" == 'Master' ]]; then
     echo "Restarting hive-server2"
     sudo systemctl restart hive-server2.service


codelixir:
Presubmit failure:

```
ERROR: Error computing the main repository mapping: Error accessing registry https://bcr.bazel.build/:
Failed to fetch registry file https://bcr.bazel.build/modules/apple_support/1.5.0/MODULE.bazel:
PKIX path validation failed: java.security.cert.CertPathValidatorException: validity check failed
```

Update: on retrying, the same error message for `https://bcr.bazel.build/modules/platforms/0.0.7/MODULE.bazel`

Update 2: there was an issue with expired certificates for bazel.build, which has now been fixed.

AkshayUpadhye (approved):

hive-lineage/hive-lineage.sh
```diff
@@ -52,7 +52,8 @@ function install_jars() {
 }
 
 function restart_hive_server2_master() {
-  ROLE=$(curl -f -s -H Metadata-Flavor:Google http://metadata/computeMetadata/v1/instance/attributes/dataproc-role)
+  # Safely get metadata without failing the script if the key is missing
+  ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role || echo "Unknown")
```
vamshikrishna-g, line 56: Just `ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)` will do

codelixir: shouldn't harm to have a fallback -- my primary concern is I that do not want the cluster creation to fail due to this init action, wdyt?

PR #1373, opened 2026-01-23
Proxyv2

dilipgodhia:
Support Hive metastore reuse and Cloud SQL Proxy v2 instance formatting. Added a new metadata property "db-hive-metastore-reuse" which must be set to true to allow the reuse of hive-metastore in CloudSQL Mysql


dilipgodhia (commented):
Initial review requested by peer

cjac (approved):
live review with Dilip.  LGTM

PR #1376, opened 2026-02-09
Install and use gnupg instead of deprecated apt-key to fix RStudio init on newer Debian images.

NavdeepCS:


cjac:
> _ No description provided. _

Hi there!  Thanks for the contribution!

Can you compare and contrast with the examples in this function and let me know what you think?  I think that we still need to create a canonical apt/dnf repo setup function.

https://github.com/LLC-Technologies-Collier/initialization-actions/blob/aecceacd715944c34533a01c13af13c014c64a77/gpu/install_gpu_driver.sh#L2860

My dream is to have all of the init actions generated from templates.  We would have a template containing functions related to system package management.  We could probably also generalize the patterns into management of other package sets such as npm, cpanm, gem, mvn, zef, cargo, php composer, go get, nuget, pip, etc.

PR #1389, opened 2026-04-24
Integrate Google Secret Manager for Hive password with logging

dilipgodhia:
Added a feature to allow the use of Google Secret Manager (GSM) for hive connectivity. In hive-site.xml, the property "javax.jdo.option.ConnectionPassword" will be removed since it will not be needed when GSM is used.

### DIFF
--- a/cloud-sql-proxy/README.md
+++ b/cloud-sql-proxy/README.md
@@ -10,14 +10,15 @@ metadata on a given Cloud SQL instance.
 
 **⚠️ IMPORTANT COMPATIBILITY NOTICE ⚠️**
 
-**Cloud SQL Proxy V2 script has been upgraded to use the latest binary client that is compatible with MySQL 8.4 and some newer Cloud SQL features.**
+**The Cloud SQL Proxy script has been upgraded to use the V2 binary client by default, which is compatible with MySQL 8.4 and some newer Cloud SQL features.**
 
-Specifically:
-* **MySQL 8.4 has updated its security model that were incompatible with the earlier version of this CloudSQL Proxy script.**
-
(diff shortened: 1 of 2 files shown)


cloud-sql-proxy/README.md
```diff
@@ -394,3 +397,31 @@ the `hive` user does not already exist in MySQL. Proceed as follows:
...
+        --scopes cloud-platform \
+        --initialization-actions gs://goog-dataproc-initialization-actions-${REGION}/cloud-sql-proxy/cloud-sql-proxy.sh \
+        --metadata "hive-metastore-instance=${PROJECT_ID}:${REGION}:${INSTANCE_NAME}" \
+        --properties "hive:hadoop.security.credential.provider.path=gsm://projects/${PROJECT_ID}" \
+        --properties "hive:hadoop.security.credstore.google-secret-manager.secret-version=latest"
+    ```
+
+    *Note: The `hadoop.security.credential.provider.path` must be set to a value other than `None` (e.g., `gsm://projects/${PROJECT_ID}`) to enable the Secret Manager integration. You can specify a specific version instead of `latest` using the `hadoop.security.credstore.google-secret-manager.secret-version` property.*
+
+    When Secret Manager is used, the initialization action will automatically remove the plaintext `javax.jdo.option.ConnectionPassword` from the `/etc/hive/conf/hive-site.xml` file after the proxy is configured, providing an extra layer of security.
```
dilipgodhia: The new metadata attribute db-hive-metastore-reuse is used in the script but not documented in the README. Please add a section or note explaining its purpose (e.g., to skip user/database creation when reusing an existing metastore).

dilipgodhia:
Need to update the Readme file also

PR #1397, opened 2026-07-01
Fix RStudio installation on Debian 13 (trixie)

DiwakarRana:
This PR updates the script to be compatible with Debian 13 (Trixie) and Ubuntu 24.04 (Noble) while maintaining full backward compatibility with older Dataproc versions:

1. Robust GPG Key Import:
-----Added a fallback to gpg if apt-key is missing (deprecated in newer OS versions).
-----Explicitly installs dirmngr alongside gnupg, which is required for GPG keyserver operations on minimal images.
-----Uses the update_apt_get helper with retries to mitigate transient network failures during key import.
2. Bypassed add-apt-repository & Dynamic CRAN Suffix:
-----Writes the CRAN repository definition directly to /etc/apt/sources.list.d/cran.list (faster and works out-of-the-box).
-----Dynamically selects the CRAN repository suffix: uses cran46 for Debian 12/13 (which transitioned to R 4.6+) and falls back to cran40 for Ubuntu 22/24.
3. Upgraded RStudio Server Version (libssl3 Compatibility):
-----For Debian 12 (bookworm), Debian 13 (trixie), Ubuntu 22.04 (jammy), and Ubuntu 24.04 (noble), the script now installs RStudio Server 2023.09.1-494 (which supports libssl3) instead of 1.2.5019 (which required the deprecated libssl1.1).
-----Ensures the systemd service file path is correctly mapped for these newer OS versions when running in passwordless mode.

Detailed Changes:
Modified rstudio/rstudio.sh:

1. Updated GPG key import logic to support gpg fallback, install dirmngr, and use update_apt_get.
2. Implemented dynamic CRAN suffix selection (cran46 vs cran40) and direct writing to cran.list.
3. Updated RSTUDIO_SERVER_VERSION and service_file path logic to include bookworm, trixie, jammy, and noble.
4. Updated the supported_os list to include Debian 13 and Ubuntu 24.04.

### DIFF
--- a/rstudio/rstudio.sh
+++ b/rstudio/rstudio.sh
@@ -27,9 +27,9 @@ function version_le() ( set +x ;  [ "$1" = "$(echo -e "$1\n$2" | sort -V | head 
 function version_lt() ( set +x ;  [ "$1" = "$2" ] && return 1 || version_le $1 $2 ; )
 
 readonly -A supported_os=(
-  ['debian']="10 11 12"
+  ['debian']="10 11 12 13"
   ['rocky']="8 9"
-  ['ubuntu']="18.04 20.04 22.04"
+  ['ubuntu']="18.04 20.04 22.04 24.04"
 )
 
 # dynamically define OS version test utility functions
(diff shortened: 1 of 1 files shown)


vinayakumarb (approved):

rstudio/rstudio.sh
```diff
@@ -166,13 +166,23 @@ if [[ "${ROLE}" == 'Master' ]]; then
...
+  fi
   # https://cran.r-project.org/bin/linux/ubuntu/
   if [[ "${OS_ID}" == "ubuntu" ]]; then
     curl -fsSL --retry-connrefused --retry 10 --retry-max-time 30 https://cloud.r-project.org/bin/linux/ubuntu/marutter_pubkey.asc | tee -a /etc/apt/trusted.gpg.d/cran_ubuntu_key.asc
   fi
-  apt-get install -y software-properties-common
-  add-apt-repository "deb http://cran.r-project.org/bin/linux/${OS_ID} ${OS_CODE}-cran40/"
+  cran_suffix="cran40"
+  if [[ "${OS_ID}" == "debian" ]] && [[ "${OS_CODE}" == "bookworm" || "${OS_CODE}" == "trixie" ]]; then
+    cran_suffix="cran46"
```
vinayakumarb, line 183: Can you limit this to only trixie?
cran46 for bookworm, may be a backward incompatible change.

vinayakumarb (changes requested):

vinayakumarb (approved):

PR #1398, opened 2026-07-01
Add Dataproc 3.0 and Hive 4 compatibility to Hive lineage script

pathakriya-cyber:
On Dataproc 3.0 (Hive 4), OpenLineage requires the GCP lineage transport library (transports-gcplineage.jar) to be explicitly copied into Hive's library folder (/usr/lib/hive/lib/) so Hive can send lineage events to GCP Dataplex without throwing ClassNotFoundException.
  
  This change:
  1. Automatically copies Spark's transports-gcplineage.jar into Hive's library folder.
  2. Modernizes legacy "gsutil cp" calls to "gcloud storage cp".
  
  Verified end-to-end against Dataproc 3.0 Debian 13:
  http://sponge2/11f79ec0-b6d7-403e-8bc9-f7e923d70635

### DIFF
--- a/hive-lineage/hive-lineage.sh
+++ b/hive-lineage/hive-lineage.sh
@@ -48,8 +48,15 @@ function set_hive_lineage_conf() {
 
 function install_jars() {
   echo "Installing openlineage-hive hook"
-  gsutil cp -P "$INSTALLATION_SOURCE/hive-openlineage-hook-$HIVE_OL_HOOK_VERSION.jar" "$HIVE_LIB_DIR/hive-openlineage-hook.jar"
-}
+  gcloud storage cp "$INSTALLATION_SOURCE/hive-openlineage-hook-$HIVE_OL_HOOK_VERSION.jar" "$HIVE_LIB_DIR/hive-openlineage-hook.jar"
+  
+    echo "Copying GCP lineage transport jar into Hive lib folder for Dataproc 3.0 compatibility..."
+    if [[ -f "/usr/lib/spark/connector/transports-gcplineage.jar" ]]; then
+      cp "/usr/lib/spark/connector/transports-gcplineage.jar" "$HIVE_LIB_DIR/openlineage-gcp-transport.jar"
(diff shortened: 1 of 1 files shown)


hive-lineage/hive-lineage.sh
```diff
@@ -48,8 +48,15 @@ function set_hive_lineage_conf() {
 
 function install_jars() {
   echo "Installing openlineage-hive hook"
-  gsutil cp -P "$INSTALLATION_SOURCE/hive-openlineage-hook-$HIVE_OL_HOOK_VERSION.jar" "$HIVE_LIB_DIR/hive-openlineage-hook.jar"
-}
+  gcloud storage cp "$INSTALLATION_SOURCE/hive-openlineage-hook-$HIVE_OL_HOOK_VERSION.jar" "$HIVE_LIB_DIR/hive-openlineage-hook.jar"
```
codelixir, line 51:
```
gcloud storage cp -P ...
```

hive-lineage/hive-lineage.sh
```diff
@@ -48,8 +48,15 @@ function set_hive_lineage_conf() {
...
+  
+    echo "Copying GCP lineage transport jar into Hive lib folder for Dataproc 3.0 compatibility..."
+    if [[ -f "/usr/lib/spark/connector/transports-gcplineage.jar" ]]; then
```
codelixir, line 54: `transports-gcplineage:1.27` is bundled in `hive-openlineage-hook.jar`, and the jar you're copying from spark classpath is based on `transports-gcplineage:1.34`

This might cause some issues with class conflicts. I'm assuming that the reason v1.27 is not working well here is because some of the classes do not support Hive 4? It is probably not a good idea to have multiple versions of the same class in the Hive class path. (Even if our local tests are passing, customer workloads might break as customers might bring in their own classes to the classpath)

codelixir (changes requested):
If you could share the fully qualified names of the class(es) that result in `ClassNotFoundException`, it would help determine if this is the right fix

